PR #36, opened 2024-07-22
Rust CI

olynch:


epatters:
Thanks Owen! How stable is Rust nightly? If running on nightly will result in frequent random CI failures I'd rather not do it.

epatters (approved):

PR #38, opened 2024-07-22
Build dev docs in CI

olynch:
Same as #37, but without including a 3.4MiB binary in the source.


olynch:
Building is successful, now we just have to deploy!

PR #41, opened 2024-07-23
CI: Linting and Formatting for Rust and Typescript

olynch:


epatters:
Thanks Owen! I tweaked the `rustfmt` settings so that it will wrap less aggressively at less than 80 characters. I'm happy with how the Rust code looks. My only complaint is the convention for wrapping `where` clauses, which gives lots of vertical real estate to what (IMO) you usually want to gloss over. However, the [setting](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#where_single_line) to control that is not stable, so I'll just live with it.

I'll look at the TypeScript next.

.github/workflows/ci.yml
```diff
@@ -20,5 +20,61 @@ jobs:
...
+          cargo test --verbose
+
+  catlog_formatting:
+    name: Catlog Formatting
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Check formatting
```
epatters, line 41: We should also run `cargo fmt --check` on the auxiliary crate in `frontend/catlog-wasm`.

epatters (changes requested):

frontend/package.json
```diff
@@ -1,43 +1,44 @@
...
-    "vite-plugin-top-level-await": "^1.4.1",
-    "vite-plugin-wasm": "^3.3.0",
-    "wasm-pack": "^0.13.0"
-  }
+    "name": "frontend",
+    "private": true,
+    "license": "MIT",
+    "version": "0.0.0",
+    "type": "module",
+    "scripts": {
```
epatters, line 7: We should have a `format` script so that we don't have to remember how to run the formatter.

.github/workflows/ci.yml
olynch: This is now in #42!

frontend/package.json
olynch: Sounds good, I'll also add a `lint` script

epatters (approved):

Issue #22, opened 2024-07-10
Tab completion for object/morphism references


epatters:
When editing the domain or codomain of a morphism, pressing Tab should pop up a list of objects that could fill the slot. Similarly, morphisms should be complete-able when editing an arrow in a path equation.

The [`solid-select`](https://github.com/thisbeyond/solid-select) component is more complex/full-featured than we need, but could be a useful source of inspiration, including for [CSS](https://github.com/thisbeyond/solid-select/blob/main/public/style.css).

epatters:
Completed last month but I forgot to link this issue.

PR #57, opened 2024-07-25
Stem cells and choosebox

### LINKED ISSUE #50: New cell logic
Right now, the logic for the new cell menu is the following.

1. There is a notebook-level boolean signal `isCellMenuOpen`.
2. Whatever the currently-focused cell is looks at that boolean, and if it is true, it opens up the new cell menu.
3. When something is clicked in the new cell menu, a new cell is created.

This leads to the slightly glitchy behavior that if you open up the cell menu, and then shift focus to another cell, the cell menu jumps to the cell that you focused to, before disappearing.

I think it makes more sense to do the following.

1. We can have a "stem cell" cell type, which simply displays a menu of options for cells to transform into. When a stem cell is focused, these options can be selected via keyboard.
2. Hitting shift-enter produces a new stem cell after the current cell, and focuses that stem cell (replicating the current "cellMenu" behavior).
3. We can have "+" buttons between every two cells, and at the beginning and end of the notebook which insert stem cells.

olynch:
Closes #50.


epatters:
This is awesome, thanks Owen!

You're planning to add back the description and keybinding to popover menu, right?

olynch:
Yes, I'm working on that now

olynch:
Question: should we attempt to unify

- `choosebox.tsx`
- `inline_input.tsx`
- `command.tsx`

into "one textbox to rule them all"? I think that the `CommandMenu` is now not used anywhere. And unifying `inline_input` with `choosebox` would enable #22.

epatters:
Agree that choosebox should be able to subsume and replace command menu.

Possibly inline input as well, although that already has a large proliferation of options. Rather, we might want to somehow attach a choosebox as a tab completion option (#22) for inline inputs.

PR #61, opened 2024-07-26
Folder reorganization to top-level package directory

olynch:
- [x] Make a top-level cargo workspace
- [x] Make a top-level npm workspace
- [x] Delete `sandbox`  directory


epatters (commented):
Thanks for doing this re-org!

One question: what should be our policy about whether to commit `Cargo.lock` files? With this PR as is, we would have a lock file committed at the top-level but .gitignored in `catlog` and `catlog-wasm`.

olynch:
Reading https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html:

> When in doubt, check Cargo.lock into the version control system (e.g. Git). For a better understanding of why and what the alternatives might be, see [“Why have Cargo.lock in version control?” in the FAQ](https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control). We recommend pairing this with [Verifying Latest Dependencies](https://doc.rust-lang.org/cargo/guide/continuous-integration.html#verifying-latest-dependencies).

I think the idea is that more application-y things should have Cargo.lock in version control, and more library-y things should not. So I don't know where that puts us exactly. I'm kind of inclined for the sake of predictability to put the lockfile in version control

epatters (approved):

PR #65, opened 2024-07-27
Show cell kind in right hand column

### LINKED ISSUE #64: Show cell type to user
Thinking about the "stem cell" design, I think that it is a bit confusing to "lose" the text that was typed into the stem cell when the stem cell becomes a normal cell. Rather, the UI could simply "fold" the tag input field so that it is hidden, but can be unfolded. I think that this would help with learnability: when you look at a notebook, you have access to the formal tag names of each cell.

olynch:
Closes #64.


epatters (approved):
Thanks, I like it!

PR #70, opened 2024-07-27
Upgrade forester theme for v4.1

epatters:
Attempting to fix that the forester theme is broken locally on Forester v4.1 but working in the CI.


epatters:
Hmm, now the theme seems broken in the preview despite working locally.

epatters:
Now I see the problem. The tarball downloaded by the CI *claims* to be v4.1 but is secretly v4.2:

```shell
$ ./forester --version
4.2.0
```

I notice that v4.2 isn't even released on [opam](https://opam-5.ocaml.org/packages/forester/) yet. Is this a nightly build?

Issue #53, opened 2024-07-24
Typescript docs


olynch:
We have great rust API docs, but the typescript docs are a little lacking: https://next.catcolab.org/dev/frontend/. Now that we actually build them, maybe we should have a policy like for the rust where docstrings are mandatory? And we should at least update the README!

epatters:
That's fair. I've been pretty inconsistent about writing API docs for the frontend so far.

For Rust, we have the policy that every public export should have a doc-comment, which makes sense because `catlog` is a library, not an application. The frontend is an application, so what counts as "public" is less obvious to me. That said, requiring that everything with `export` have a doc-comment is stringent but should perhaps still be the standard.

PR #92, opened 2024-07-31
Improve build instructions in README

KevinDCarlson:
I've tried edited the main and the frontend readmes to accurately reflect how to make a local build.

Minimally responds to #53 .

# Default TODO list

- [x] Are any relevant issues to the PR mentioned so that they will be automatically closed when the PR merges?

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,17 +7,12 @@ CatColab (working name) is a collaborative environment for
 
 > formal, interoperable, conceptual modeling.
 
-To build locally, first, be sure you have wasm-pack installed. Clone the repository and run
+To build locally, clone the repository and run
 
 ```bash
 > npm install
 > npm run build
-```
-
-from packages/frontend. This will build necessary WebAssembly, Typescript, and Javascript packages (via Vite.) Then, run 
-
-```bash
-> vite
+> npm run dev
 ```
 
 and navigate your browser to the URL provided by Vite.
--- a/README.md
+++ b/README.md
@@ -6,3 +6,18 @@
 CatColab (working name) is a collaborative environment for
 
 > formal, interoperable, conceptual modeling.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -6,3 +6,18 @@
 CatColab (working name) is a collaborative environment for
 
 > formal, interoperable, conceptual modeling.
+
+To build locally, first, be sure you have wasm-pack installed. Clone the repository and run
+
+```bash
```
epatters, line 12: The instructions would be easier to follow if you put `cd packages/frontend` as the first command rather than saying to run the commands from that directory.

README.md
```diff
@@ -6,3 +6,18 @@
 CatColab (working name) is a collaborative environment for
 
 > formal, interoperable, conceptual modeling.
+
+To build locally, first, be sure you have wasm-pack installed. Clone the repository and run
```
epatters: I'm hoping that a manual/global install of `wasm-pack` isn't necessary? It's in the dev dependencies for the frontend, which ought to suffice.

epatters (commented):
Thanks Kevin for paying attention to this.

README.md
KevinDCarlson: Hm, yeah, I got it built first by installing it but let me try redoing this without it installed.

README.md
KevinDCarlson: Sure, also, is there any reason why we build from `/packages/frontend` instead of from `/`?

KevinDCarlson: I added npm scripts to build and run from the root directory, lemme know if there's any issue with that.

README.md
KevinDCarlson: Yeah, you're right.

README.md
epatters: Thanks, that's great!

epatters (approved):

PR #114, opened 2024-08-20
ODE simulation in Rust core

### LINKED ISSUE #110: ODE solving in Rust
The lowest effort approach to implementing ODE semantics for models is to use an ODE solver native to Rust, compile it to wasm, and run it locally in the browser. The scientific computing ecosystem in Rust is quite weak compared to, say, Julia, but there are a few crates for solving ODEs:

- [`ode-solvers`](https://github.com/srenevey/ode-solvers): has the standard Runge-Kutta plus a few other solvers
- [`diffsol`](https://github.com/martinjrobins/diffsol): brand new, with a focus on stiff equations

olynch:
*PR description written by @epatters*.

- Lexer, parser, and interpreter for simple mathematical expressions
- Solve ODEs specified dynamically by such expressions
- Solve ODEs defined by stock-flow diagrams with arbitrary flow expressions

Closes #110.


epatters (approved):
Thanks Owen!

Models produced by the frontend have UUIDs rather Ustrs as IDs. I started generalizing `StockFlowAnalysis` et al to have a generic type parameter `Id` rather hardcoding `Ustr` but then realized that `Ustr` is baked into the expression compiler. Even if it weren't, we would still need to use a mapping between UUIDs and names for the user to be able to input the expressions in the first place. I'm going to merge this and defer that problem to later. Instead, I'll start the frontend integration with simple ODE semantics that don't require arbitrary expressions, such as Lokta-Volterra or mass-action.

PR #118, opened 2024-08-22
Refactor the trait interface for finitely generated things

olynch:
This compiles, but doesn't pass lints. The idea behind this refactor is that we should be more explicit in our traits about finitely-generated categories and double theory models. Specifically, "finitely generated" traits should expose actual implementations of FinGraph. This is in accordance with what "finitely generated" means mathematically for categories.

# Default TODO list

- [x] Are any relevant issues to the PR mentioned so that they will be automatically closed when the PR merges?

### DIFF
--- a/packages/catlog-wasm/src/model.rs
+++ b/packages/catlog-wasm/src/model.rs
@@ -8,8 +8,9 @@ use tsify_next::Tsify;
 use wasm_bindgen::prelude::*;
 
 use super::theory::*;
-use catlog::dbl::model::{self as dbl_model, DblModel as BaseDblModel, InvalidDiscreteDblModel};
+use catlog::dbl::model::{self as dbl_model, FgDblModel, InvalidDiscreteDblModel};
 use catlog::one::fin_category::UstrFinCategory;
+use catlog::one::Category as _;
 use catlog::one::Path;
 use catlog::validate::{self, Validate};
 
@@ -220,7 +221,7 @@ impl DblModel {
     #[wasm_bindgen]
     pub fn morphisms(&self) -> Vec<Mor> {
         all_the_same!(match &self.0 {
-            DblModelBox::[Discrete](model) => model.morphisms().map(|f| f.into()).collect()
+            DblModelBox::[Discrete](model) => model.morphisms().map( Mor::Basic).collect()
         })
     }
 
--- a/packages/catlog-wasm/src/model.rs
+++ b/packages/catlog-wasm/src/model.rs
@@ -8,10 +8,10 @@ use tsify_next::Tsify;
 use wasm_bindgen::prelude::*;
 
 use super::theory::*;
-use catlog::dbl::model::{self as dbl_model, FgDblModel, InvalidDiscreteDblModel};
+use catlog::dbl::model::{self as dbl_model, InvalidDiscreteDblModel};
 use catlog::one::fin_category::UstrFinCategory;
-use catlog::one::Category as _;
 use catlog::one::Path;
+use catlog::one::{Category as _, FgCategory};
 use catlog::validate::{self, Validate};
 
 /// An object in a model of a double theory.
@@ -213,7 +213,7 @@ impl DblModel {
     #[wasm_bindgen]
     pub fn objects(&self) -> Vec<Ob> {
         all_the_same!(match &self.0 {
-            DblModelBox::[Discrete](model) => model.objects().map(|x| x.into()).collect()
+            DblModelBox::[Discrete](model) => model.object_generators().map(|x| x.into()).collect()
         })
     }
 
@@ -221,7 +221,7 @@ impl DblModel {
     #[wasm_bindgen]
     pub fn morphisms(&self) -> Vec<Mor> {
         all_the_same!(match &self.0 {
-            DblModelBox::[Discrete](model) => model.morphisms().map( Mor::Basic).collect()
+            DblModelBox::[Discrete](model) => model.morphism_generators().map(Mor::Basic).collect()
         })
     }
 
--- a/packages/catlog-wasm/src/theory.rs
+++ b/packages/catlog-wasm/src/theory.rs
@@ -34,9 +34,8 @@ pub enum MorType {
     /// Basic or generating morphism type.
     Basic(Ustr),
 
-    /// Mor type on an object type.
-    #[serde(rename = "Hom")]
-    Identity(Box<ObType>),
+    /// Hom type on an object type.
+    Hom(Box<ObType>),
 }
 
 /// Convert from object type in a discrete double theory.
@@ -51,7 +50,7 @@ impl From<FinMor<Ustr, Ustr>> for MorType {
     fn from(mor: FinMor<Ustr, Ustr>) -> Self {
         match mor {
             FinMor::Generator(e) => MorType::Basic(e),
-            FinMor::Id(v) => MorType::Identity(Box::new(ObType::Basic(v))),
+            FinMor::Id(v) => MorType::Hom(Box::new(ObType::Basic(v))),
         }
     }
 }
@@ -75,7 +74,7 @@ impl TryFrom<MorType> for FinMor<Ustr, Ustr> {
     fn try_from(mor_type: MorType) -> Result<Self, Self::Error> {
         match mor_type {
             MorType::Basic(name) => Ok(FinMor::Generator(name)),
-            MorType::Identity(x) => (*x).try_into().map(FinMor::Id),
+            MorType::Hom(x) => (*x).try_into().map(FinMor::Id),
         }
     }
 }
@@ -95,7 +94,7 @@ impl From<TabMorType<Ustr, Ustr>> for MorType {
     fn from(mor_type: TabMorType<Ustr, Ustr>) -> Self {
         match mor_type {
             TabMorType::Basic(name) => MorType::Basic(name),
-            TabMorType::Mor(x) => MorType::Identity(Box::new((*x).into())),
+            TabMorType::Hom(x) => MorType::Hom(Box::new((*x).into())),
         }
     }
 }
@@ -119,7 +118,7 @@ impl TryFrom<MorType> for TabMorType<Ustr, Ustr> {
     fn try_from(mor_type: MorType) -> Result<Self, Self::Error> {
         match mor_type {
             MorType::Basic(name) => Ok(TabMorType::Basic(name)),
-            MorType::Identity(x) => (*x).try_into().map(|x| TabMorType::Mor(Box::new(x))),
+            MorType::Hom(x) => (*x).try_into().map(|x| TabMorType::Hom(Box::new(x))),
         }
     }
 }
--- a/packages/catlog-wasm/src/theory.rs
+++ b/packages/catlog-wasm/src/theory.rs
@@ -34,8 +34,8 @@ pub enum MorType {
     /// Basic or generating morphism type.
     Basic(Ustr),
 
-    /// Hom type on an object type.
-    Hom(Box<ObType>),
+    /// Mor type on an object type.
+    Mor(Box<ObType>),
 }
 
 /// Convert from object type in a discrete double theory.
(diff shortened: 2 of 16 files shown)


packages/catlog-wasm/src/model.rs
```diff
@@ -177,19 +177,19 @@ impl DblModel {
         })
     }
 
-    #[cfg(test)]
```
epatters: Thanks Owen, this refactor is great!

I've actually replaced this deleted/commented code in #121. What do you think about merging that first and then rebasing this PR? Hopefully will make for a cleaner diff.

olynch: Sounds good!

epatters: Great, I've gone ahead and merged that one.

packages/catlog-wasm/src/theory.rs
```diff
@@ -34,8 +34,8 @@ pub enum MorType {
     /// Basic or generating morphism type.
     Basic(Ustr),
 
-    /// Hom type on an object type.
-    Hom(Box<ObType>),
+    /// Mor type on an object type.
+    Mor(Box<ObType>),
```
epatters: I find this to be less clear than before: having a ` Mor` variant in `MorType` doesn't evoke anything for me. What do you think about either (1) keeping `Hom` here or (2) using `Id` instead?

packages/catlog/src/dbl/mod.rs
```diff
@@ -7,3 +7,5 @@ pub mod pasting;
 pub mod model;
 pub mod model_morphism;
 pub mod theory;
+
+// pub mod semantics;
```
epatters: Left over from a previous design?

packages/catlog/src/dbl/model.rs
```diff
@@ -50,7 +50,8 @@ use super::theory::{DblTheory, DiscreteDblTheory};
 use crate::one::fin_category::{FpCategory, InvalidFpCategory, UstrFinCategory};
 use crate::one::*;
 use crate::validate::{self, Validate};
-use crate::zero::{Column, IndexedHashColumn, Mapping};
+use crate::zero::{IndexedHashColumn, Mapping};
+use std::iter::Iterator;
```
epatters: I try to keep the `std` imports at the top. Unfortunately, the formatter isn't smart enough to do this on its own.

packages/catlog/src/dbl/model.rs
```diff
@@ -142,30 +115,32 @@ pub trait DblModel {
...
-    */
-    fn morphisms(&self) -> impl Iterator<Item = Self::Mor>;
+/// A finitely-generated double model
+pub trait FgDblModel: DblModel + FgCategory {
+    /// get the type of a morphism generator
+    fn mor_gen_type(&self, mor: &Self::MorGen) -> Self::MorType;
 
-    /// Iterates over basic objects of the given type.
-    fn objects_with_type(&self, typ: &Self::ObType) -> impl Iterator<Item = Self::Ob>;
+    /// an iterator over the objects in the model
```
epatters: Our doc comments should follow the conventions commonly used in the Rust community, including the standard library ([typical examples](https://doc.rust-lang.org/std/vec/struct.Vec.html#implementations)). This seems to include making the first line a short sentence, capitalized and punctuated appropriately, with an active verb.

So this one could be

> Iterates over the objects in the model.

or more verbosely

> Returns an iterator over the objects in the model.

Similarly for the doc comments below.

packages/catlog/src/dbl/model.rs
```diff
@@ -210,14 +185,9 @@ where
...
 
-    /// Discrete double theory that the model is of.
-    pub fn theory(&self) -> &Arc<DiscreteDblTheory<Cat>> {
-        &self.theory
-    }
-
-    /// Graph of object and morphism generators for the model.
-    pub fn generating_graph(&self) -> &impl FinGraph<V = Id, E = Id> {
-        self.category.generators()
+    /// get the Arc for the theory of this model
```
epatters: Not important right now, but I just used `Arc` instead of say `archery` to keep things simple. I gather the overhead is minimal compared to the `Rc`. But it would be nice to have some common guidelines. I noticed you used `archery` in the directories implementation.

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -333,14 +334,12 @@ mod tests {
 
         let maps = DiscreteDblModelMapping::morphisms(&positive_loop, &positive_loop).find_all();
         assert_eq!(maps.len(), 2);
-        assert!(matches!(maps[0].apply_mor(&pos), Some(Path::Id(_))));
-        assert!(matches!(maps[1].apply_mor(&pos), Some(Path::Seq(_))));
```
epatters, line 337: OK, I know this test was bad, but this feels like throwing out the baby with the bathwater! I'm tempted to add an `Ord` implementation to `Path` (whenever `V` and `E` are `Ord`) consistent with the [lexicographic order](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Ord-for-Vec%3CT,+A%3E) for `Vec`.

For now, can you just comment these out but not delete them? I'll fix them later.

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -353,8 +352,9 @@ mod tests {
         let maps =
             DiscreteDblModelMapping::morphisms(&negative_loop, &negative_feedback).find_all();
         assert_eq!(maps.len(), 2);
-        assert_eq!(maps[0].apply_ob(&base_pt), Some(ustr("x")));
-        assert_eq!(maps[1].apply_ob(&base_pt), Some(ustr("y")));
+        let map_obs = maps.iter().map(|ob| ob.apply_ob(&base_pt)).collect::<Vec<_>>();
+        assert!(map_obs.contains(&Some(ustr("x"))));
+        assert!(map_obs.contains(&Some(ustr("y"))));
```
epatters: Thanks for fixing this one :)

packages/catlog/src/dbl/theory.rs
```diff
@@ -38,16 +38,16 @@ summarized by the table:
 Models of a double theory are automatically *categorical* structures, rather
 than merely *set-theoretical* ones, because each object type is assigned not
 just a set of objects but also a span of morphisms between those objects,
-constituting a category. The morphism data comes from a distinguished "Hom" type
+constituting a category. The morphism data comes from a distinguished "Mor" type
```
epatters: Ditto my first comment about `Mor` in this context.

packages/catlog/src/one/path.rs
```diff
@@ -48,6 +48,12 @@ pub enum Path<V, E> {
     Seq(NonEmpty<E>),
 }
 
+impl<V, E> From<E> for Path<V, E> {
+    fn from(e: E) -> Self {
+        Path::single(e)
+    }
+}
```
epatters, line 55: Possibly confusing in the common case where `V = E`, say both are UUIDs. When you convert into `Path`, do you get `Path::Id` or `Path::single`?

epatters (changes requested):
Good refactor, thanks Owen! Always nice to see more red lines than green in a refactor.

packages/catlog/src/dbl/model.rs
olynch: I think archery would be overkill here. This is a impl for the struct, not the trait, so we can be specific here.

packages/catlog/src/dbl/model_morphism.rs
olynch: I think that this actually makes sense: if `maps.len() == 2`, and `x` and `y` are both in `map_obs`, then that's all we care about...

olynch: Oh, I thought this was for another one. I'll do the same strategy here.

packages/catlog/src/one/path.rs
olynch: Oh, I see, that's a subtle bug. I've addressed everything else, let me think about this one a bit.

olynch: I think I was assuming from would be called on an morphism generator. Maybe there's a better way to do this though.

olynch: E.g., have this be a method on `FgCategory` with some descriptive name like "mor_from_mor_gen". What would be a good name for this?

epatters:
@olynch, I realized that there is a conceptual problem with the new design that I didn't detect in my first review. The problem is that:

- The trait `FgCategory` is being thought of as describing bare categories (categories without extra structure) since object generators are identified with objects. Thus, it not make sense to write, say, `FgMonoidalCategory: FgCategory` because in a finitely generated monoidal category, the objects and the object generators will not coincide.
- The trait `FgDblModel` should apply to models of double theories beyond discrete ones (or else what is the point of having a trait beyond the struct `DiscreteDblModel`), yet it extends `FgCategory`. But this means that it identifies objects and object generators, which won't be the case for double theories with nontrivial object operations.

I see two possible solutions:

1. change the scope of `FgCategory` by adding an `ObGen` associated type, or
2. have `FgDblModel` directly add the types it needs rather than extending `FgCategory`.

I'm inclined toward (1).

packages/catlog/src/one/path.rs
epatters: On further discussion, we agreed that this is fine because paths are morphisms, so it makes sense to have the standard conversion be from edges to paths.

packages/catlog/src/dbl/theory.rs
```diff
@@ -307,8 +293,8 @@ pub enum TabMorType<V, E> {
     /// Basic or generating morphism type.
     Basic(E),
 
-    /// Hom type on an object type.
-    Hom(Box<TabObType<V, E>>),
+    /// Mor type on an object type.
+    Mor(Box<TabObType<V, E>>),
```
epatters: We decided to use `Id` or `Identity` instead of `Mor` here.

packages/catlog/src/dbl/theory.rs
```diff
@@ -330,8 +316,8 @@ pub enum TabMorOp<V, E> {
     /// Identity operation on a morphism type.
     Id(TabMorType<V, E>),
 
-    /// Hom operation on an object operation.
-    Hom(TabObOp<V, E>),
+    /// Mor operation on an object operation.
+    Mor(TabObOp<V, E>),
```
epatters: Ditto, but also how are going to avoid confusion here since we already have an `Id` in this enum?

packages/catlog/src/dbl/theory.rs
```diff
@@ -178,12 +172,12 @@ pub trait DblTheory {
...
     ) -> Self::MorOp;
 
-    /** Hom operation for an object operation.
+    /** Identity morphism operation for an object operation.
 
     Viewing the theory as a double category, this is the identity cell on an
     arrow.
     */
-    fn hom_op(&self, f: Self::ObOp) -> Self::MorOp {
+    fn op_id(&self, f: Self::ObOp) -> Self::MorOp {
```
epatters: I'm struggling a lot with the terminology here too. It's confusing.

epatters (commented):
I made some improvements tonight but there are still issues with consistent use of terminology (see below). The root problem is that the word "identity" is overloaded in double category theory because there are identities in both directions. I'm tempted to return to using `Id` for internal identities and `Hom` for external identities simply to avoid this ambiguity (while keeping the move from `Hom` to `Mor` in `Category`).

epatters:
OK, for now I'm going to revert to using "Hom" to describe the external identities in a double theory. I'm open to revision of this terminology in a future PR but it needs to be done carefully and consistently throughout the codebase and documentation, and the present state of this PR is quite far from that.

packages/catlog/src/one/category.rs
```diff
@@ -178,114 +178,94 @@ object generators (unless there are nontrivial equations between objects), but
...
-    /// Iterates over morphism generators of the category.
-    fn hom_generators(&self) -> impl Iterator<Item = Self::Hom>;
-
-    /// Iterates over morphism generators with the given domain.
-    fn generators_with_dom(&self, x: &Self::Ob) -> impl Iterator<Item = Self::Hom>;
-
-    /// Iterates over morphism generators with the given codomain.
-    fn generators_with_cod(&self, x: &Self::Ob) -> impl Iterator<Item = Self::Hom>;
+    /// The graph of generating morphisms and objects
+    fn generating_graph(&self) -> &impl FinGraph<V = Self::ObGen, E = Self::MorGen>;
```
epatters: With our revised understand of the concept, an `FgCategory` doesn't have a generating graph. It has something with the flavor of a computad: a set of object generators and an endo-span of morphism generators on *objects* (not object generators).

So I think that `FgCategory` and `DblModel` need an interface for generators that does not make direct reference to graphs. OTOH, *discrete* double models do have a generating graph and the struct `DiscreteDblModel` should have a `generating_graph` method.

epatters (changes requested):
I'm now happy with this PR subject to a lingering issue related to my previous comment about `FgCategory`.

packages/catlog/src/one/category.rs
olynch: Yes, OK, good point! Will update with regard to this.

epatters (approved):
Thanks Owen, looks good!

PR #134, opened 2024-08-29
Add missing readme step, correct server error

KevinDCarlson:
The error when somebody forgets to configure `VITE_BACKEND_HOST` was misspelling the relevant environment variable. I also added this step to the readme.

# Default TODO list

- [x ] Are all notable changes added to the changelog in `dev-docs/`?
- [ x] Are there any manual checks that you did while preparing this release that could be automated and become part of the CI?
- [x ] Are any relevant issues to the PR mentioned so that they will be automatically closed when the PR merges?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,11 @@ An early demo is available at <https://catcolab.org>.
 The staging deployment, synced to the `main` branch, is available at <https://next.catcolab.org>.
 Documentation for developers is browsable at <https://next.catcolab.org/dev/>.
 
-To build locally, clone the repository and run
+To build locally, given you have npm and Rust installed, clone the repository and run
 
 ```bash
 > npm install
+> export VITE_BACKEND_HOST=backend-next.catcolab.org
 > npm run build
 > npm run dev
 ```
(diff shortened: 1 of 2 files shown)


epatters:
Thanks for starting something here, Kevin. My preference would be reduce the friction by setting a sensible default. Here's what I wrote on our Zulip:

> @olynch, is there a reason not to just set BACKEND_HOST to the development server in the Vite config .env.development?

olynch (approved):
Looks good, yes that was a silly mistake on my part!

olynch:
> Thanks for starting something here, Kevin. My preference would be reduce the friction by setting a sensible default. Here's what I wrote on our Zulip:
> 
> > @olynch, is there a reason not to just set `VITE_BACKEND_HOST` to the development server in the Vite config `.env.development`?

My only question would be whether .env.development is usually supposed to checked in to VC or not? If it is, then this sounds reasonable to me!

epatters:
Good question. From what I read [here](https://github.com/vitejs/vite/issues/14380#issuecomment-1721208976), any sensitive info is supposed to go in `*.env.local`, which are gitignored by default. So any info that we would put in the README should be fine to go into a `*.env` file.

PR #135, opened 2024-08-29
Notebook interface for analyzing models

### LINKED ISSUE #129: Notebook interface for model views
The pane on the right-hand side of the model editor should become a notebook in its own right, whose cell types are the different views available for models of the active theory.

epatters:
Closes #129.


packages/frontend/src/model/model_analyzer.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const validatedModel = useContext(ValidatedModelContext);
+
+    return (
+        <Show when={theory?.()}>
+            {(theory) => (
+                <Show
+                    when={theory().getModelAnalysis(props.content.tag)}
+                    fallback={<span>Internal error: model view not defined</span>}
+                >
+                    {(meta) => (
```
olynch: Why is this called meta?

packages/frontend/src/model/model_analyzer.tsx
```diff
@@ -0,0 +1,97 @@
...
+                            }
+                        />
+                    )}
+                </Show>
+            )}
+        </Show>
+    );
+}
+
+function modelAnalysisCellConstructors(theory?: Theory): CellConstructor<ModelAnalysis<unknown>>[] {
```
olynch: This should just take in a theory: Theory; the case when theory is undefined should be handled elsewhere.

packages/frontend/src/model/types.ts
```diff
@@ -1,21 +1,25 @@
...
-    // Identifier of double theory that the model is of.
+    /** Identifier of double theory that the model is of. */
     theory?: TheoryId;
 
-    // Content of the model, formal and informal.
+    /** Content of the model, formal and informal. */
     notebook: Notebook<ModelJudgment>;
+
+    /** Analysis of the model, a separate notebook. */
+    analysis: Notebook<ModelAnalysis<unknown>>;
```
olynch: Why do we have this unknown parameter? This seems janky to me.

packages/frontend/src/model/model_analyzer.tsx
epatters: That object includes metadata about the analysis but I agree that it's confusing. I'll call it `analysis` instead.

PR #138, opened 2024-08-30
Move clone constraint to Category::Ob/Mor

### LINKED ISSUE #137: Ob/Mor/ObGen/MorGen should have Clone trait
Presently, these types are just required to support the `Eq` trait. However, we almost always want to clone these as well, and rather than repeatedly add these constraints to downstream methods we could declare them once up front.

kris-brown:
Closes #137

### DIFF
--- a/packages/catlog-wasm/src/theory.rs
+++ b/packages/catlog-wasm/src/theory.rs
@@ -243,7 +243,7 @@ impl MorTypeIndex {
         self.0.get(m).copied()
     }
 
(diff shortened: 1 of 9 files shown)


epatters (commented):
Thanks Kris! I think we should go one step further and add `Clone` to the element types in `catlog::zero`.

epatters (approved):
Thanks for the update. I've now pushed this a little further. That we ended up with a net-negative diff makes me think the change was worth it!

Issue #131, opened 2024-08-27
Styled tooltips for buttons


epatters:
We now have an `IconButton` component used in several places. These should have a prop to specify a tooltip that will appear on hover.

It shouldn't be necessary to roll this ourselves. We could use Corvu's [tooltip component](https://corvu.dev/docs/primitives/tooltip/), for example.

AlexVCS:
Hi there, I'd like to work on this issue. Can you assign it to me please?

epatters:
@AlexVCS, thanks for offering to work on this issue!

The problem seems to be that you need a more recent version of Rust. I've updated the Cargo.toml to reflect this (#132). I'd suggest updating to the latest stable version of Rust. The tool `rustup` is very convenient for this if you're not already using it.

We should probably also update the developer instructions in the README with info about getting set up with Rust.

AlexVCS:
> npm run dev

No prob. I've tried running that in bash and then ran `npm run dev` but the page still looks the same. Is there anything else I need to run?

epatters:
Not that I know of. Do you see any errors in the browser console? Also, as a sanity check, does the app come up when you go to <https://next.catcolab.org>?

AlexVCS:
> Not that I know of. Do you see any errors in the browser console? Also, as a sanity check, does the app come up when you go to https://next.catcolab.org?

I'm getting this error despite running the script in bash:

```
chunk-ISFT24YX.js?v=d776f89f:1039 Uncaught Error: Must set environment variable BACKEND_HOST
    at castError (chunk-ISFT24YX.js?v=d776f89f:1039:10)
    at handleError (chunk-ISFT24YX.js?v=d776f89f:1052:17)
    at runComputation (chunk-ISFT24YX.js?v=d776f89f:773:12)
    at updateComputation (chunk-ISFT24YX.js?v=d776f89f:738:3)
    at createMemo (chunk-ISFT24YX.js?v=d776f89f:264:10)
    at [solid-refresh]App (@solid-refresh:22:20)
    at chunk-ISFT24YX.js?v=d776f89f:597:14
    at untrack (chunk-ISFT24YX.js?v=d776f89f:482:12)
    at Object.fn (chunk-ISFT24YX.js?v=d776f89f:593:11)
    at runComputation (chunk-ISFT24YX.js?v=d776f89f:759:22)Caused by: Must set environment variable BACKEND_HOST
 ```
 
 The site took a little while to load but did come up!

[...]

PR #143, opened 2024-09-02
Styled tooltips

AlexVCS:
# Default TODO list

- [X] Are any relevant issues to the PR mentioned so that they will be automatically closed when the PR merges? #131


package.json
```diff
@@ -27,5 +27,8 @@
     "workspaces": [
         "packages/backend",
         "packages/frontend"
-    ]
+    ],
+    "dependencies": {
+        "@corvu/tooltip": "^0.2.1"
```
epatters: The dependency should go in `packages/frontend/package.json` rather than the top-level `package.json`.

epatters (changes requested):
Thanks for the feature!

Quick comment: you should run `npm run format` and `npm run lint` before committing. That will re-format the TypeScript files and prevent the CI from complaining.

package.json
AlexVCS: Thanks! That should be fixed now :)

epatters:
This is looking good! To test it out, perhaps you could add a tooltip to one of the existing buttons, such as [this one](https://github.com/ToposInstitute/CatColab/blob/bf3bc2fbe3b2c3b479ddcf57d097094e56f8fafe/packages/frontend/src/model/model_editor.tsx#L74).

AlexVCS:
> This is looking good! To test it out, perhaps you could add a tooltip to one of the existing buttons, such as [this one](https://github.com/ToposInstitute/CatColab/blob/bf3bc2fbe3b2c3b479ddcf57d097094e56f8fafe/packages/frontend/src/model/model_editor.tsx#L74).

Think there's something wrong with how I'm implementing it. It has put this on top of the buttons and they aren't working either!

localhost_5173_

I probably need to refactor it, will take another look at it when I can.

olynch:
Hi @AlexVCS, if you rebase on top of main, the failing check should be disabled!

AlexVCS:
I looked at their examples more and have made progress. There's the `<Tooltip.trigger>` which is what's displaying with the original `IconButton`. I'm considering the best way to proceed (do I try to hide the trigger and show the child inside it (the IconButton)?) and also how to dynamically render what would be inside the `<Tooltip.content>`. I'll have another look at it later today.

Screenshot 2024-09-04 at 9 15 46 AM

epatters:
Nice progress! I think that instead of using `Trigger`, you can wrap the `IconButton` with an `Anchor` instead: https://corvu.dev/docs/primitives/tooltip/#Anchor

epatters:
Another option, slightly shorter, might be:

```jsx
<Tooltip.Trigger class="icon-button" {...buttonProps}>
    {props.children}
</Tooltip.Trigger>
```

To be clear, I haven't tested either of these!

AlexVCS:
> Another option, slightly shorter, might be:
> 
> ```js
> <Tooltip.Trigger class="icon-button" {...buttonProps}>
>     {props.children}
> </Tooltip.Trigger>
> ```
> 
> To be clear, I haven't tested either of these!

Both of those work! I went with the shorter of the two for conciseness. Thanks!

AlexVCS:
What would like to see inside of the Tooltip? `{props.children}` of course refers to multiple elements that are rendered inside of the IconButton. I'm trying to think of the best way to handle that, do you want it to be set wherever the IconButton is used?

epatters:
I would just add an optional prop, `tooltip?: string`, to the `IconButton` props and set that as the content of the tooltip when it is defined (otherwise, have no tooltip).

We might later want to allow an arbitrary component in the tooltip, but I think doing that now would be getting ahead of ourselves since the vast majority of tooltips will just be strings.

AlexVCS:
Here's an example of the tooltip in action:

Screenshot 2024-09-06 at 10 24 33 AM

I also put in a logical && so if there's no `props.tooltip` the tooltip and arrow won't render! Please let me know what you think!

npm
epatters, line 1: Looks like an empty file crept in here.

packages/frontend/src/components/icon_button.tsx
```diff
@@ -1,19 +1,31 @@
...
+  return (
+    <Tooltip openOnHover>
+      <Tooltip.Trigger class="icon-button" {...buttonProps}>
+        {props.children}
+      </Tooltip.Trigger>
+      <Tooltip.Portal>
+        {props.tooltip && (
+          <Tooltip.Content data-corvu-tooltip-content={props}>
+            {props.tooltip}
+            <Tooltip.Arrow />
```
epatters: We should remove the `Tooltip.Arrow` (i.e., delete this line) since the arrow is rendering on top of the button. I think the meaning of the tooltip will be clear without the arrow.

epatters (changes requested):
Thanks, this is looking good! You'll need to run the formatter/linter again, plus a few small comments below.

AlexVCS:
Thanks for the feedback. I pulled in the recent changes from main but it broke my local site. I noticed the actual site and next url behave the same way (just showing an error). Wanted to let you know about this issue:

Screenshot 2024-09-07 at 3 19 04 PM

epatters:
Thanks for the heads up. The backend seems to be down. We will investigate.

epatters:
Now I'm confused because I just tried it again and it seems to be working. Perhaps the problem has resolved itself? I will test again shortly.

AlexVCS:
> Thanks for the heads up. The backend seems to be down. I get an error message in the browser console.
> 
> _Update_: Apparently the backend crashed last night. We are looking into it.

Good to know, all of the errors I said I was having before have now been resolved!

epatters:
Hi, so I've been playing around with this (see my commit above) but I'm encountering problems---not due to your code, which looks great to me, but due to interactions between the tooltip lib and other functionality. Specifically, when a tooltip is added to the notebook cell drag handles, it is no longer possible to initiate a drag. I tried a bunch of things but couldn't figure out a way around that. I also encountered more minor usability issues about when the tooltips open or close.

So unfortunately this can't be merged as it stands. I'm going to try out another library and see if we have better luck.

epatters:
Changing out the tooltip component seems to have fixed the issue. I'm going to close this PR in lieu of #163, but I appreciate your help in getting this feature started and your patience as we worked through various build and documentation problems.

Issue #104, opened 2024-08-07
Validate morphisms between models of discrete double theories


epatters:
Follow-up task to #103. Validation should follow the pattern already established for graph morphisms.

kris-brown:
I noticed there is no graph morphism validation error for having *extra data* in the map. Is this by design?

epatters:
It's not by design; I didn't even think about it :) Now that I am thinking about it, I'm not sure whether it should be an error. Maybe it's fine? What do you think?

Also CC @olynch who has been thinking about dependent types as a relation between the base and the total type.

kris-brown:
I'm fine either way, but it reminds me of the `@acset` macro which used to silently ignore any misspelled words, e.g. `Employe=3`. I'm also imagining programmatically generating the vectors for the components of a graph mapping and, due to some conceptual error, we're producing vectors larger than they need to be. It might be nicer to get an error of the "index error" kind than a "unnatural mapping" kind.

epatters:
I'm persuaded by that, especially given that CatColab is supposed to be designed to give informative error messages!

kris-brown:
I guess this only makes sense for *finitely presented* graph morphisms / model morphisms - the general interface just exposes how it acts on objects, so it wouldn't make sense to test for 'extra data'.

PR #155, opened 2024-09-07
Validate morphisms between models of discrete double theories

kris-brown:
Addresses https://github.com/ToposInstitute/CatColab/issues/104

This PR defines a `DblModelMorphism` type which contains a double model mapping as well as info about the domain and codomain. This can be validated to see if objects/generating morphisms are not mapped at all, mapped to invalid identifiers, mapped to things of the wrong type, or mapped in a way that doesn't preserve dom/codom (for the morphisms).

### DIFF
--- a/packages/catlog/src/dbl/model_morphism.rs
+++ b/packages/catlog/src/dbl/model_morphism.rs
@@ -23,12 +23,12 @@ use std::sync::Arc;
 
 use derivative::Derivative;
 
-use crate::validate::{self, Validate};
 use nonempty::NonEmpty;
 use thiserror::Error;
 
 use crate::one::graph_algorithms::{simple_paths, spec_order};
 use crate::one::*;
+use crate::validate::{self, Validate};
 use crate::zero::{Column, HashColumn, Mapping};
 
 use super::model::{DblModel, DiscreteDblModel, FgDblModel};
@@ -217,24 +217,22 @@ impl<'a, Map, Dom, Cod> DblModelMorphism<'a, Map, Dom, Cod>
 where
     Map: DblModelMapping,
     Dom: FgDblModel<Ob = Map::DomOb, Mor = Map::DomMor>,
-    Dom::ObGen: Clone,
-    Dom::MorGen: Clone,
     Cod: DblModel<Ob = Map::CodOb, Mor = Map::CodMor, ObType = Dom::ObType, MorType = Dom::MorType>,
 {
     /// Iterates over failures of the mapping to be a double model morphism
     pub fn iter_invalid(
         &self,
-    ) -> impl Iterator<Item = InvalidDblModelMorphism<Map::DomOb, Map::DomMor>> + 'a {
+    ) -> impl Iterator<Item = InvalidDblModelMorphism<Dom::ObGen, Dom::MorGen>> + 'a {
         let DblModelMorphism(mapping, dom, cod) = *self;
         let ob_errors = dom.object_generators().filter_map(|v| {
             if !mapping.is_ob_assigned(&v.clone().into()) {
-                Some(InvalidDblModelMorphism::MissingOb(v.into()))
+                Some(InvalidDblModelMorphism::MissingOb(v))
             } else {
                 let f_v = mapping.apply_ob(&v.clone().into()).unwrap();
                 if !cod.has_ob(&f_v) {
-                    Some(InvalidDblModelMorphism::Ob(v.into()))
+                    Some(InvalidDblModelMorphism::Ob(v))
                 } else if dom.ob_type(&v.clone().into()) != cod.ob_type(&f_v) {
-                    Some(InvalidDblModelMorphism::ObType(v.into()))
+                    Some(InvalidDblModelMorphism::ObType(v))
                 } else {
                     None
                 }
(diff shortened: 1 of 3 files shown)


epatters:
Thanks Kris! One quick comment: to fix the CI failure, run the linter (`cargo clippy`).

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -23,6 +23,10 @@ use std::sync::Arc;
 
 use derivative::Derivative;
 
+use crate::validate::{self, Validate};
```
epatters: I like to separate the `crate` imports from the external ones, hence this import would go below. Unfortunately the linter is not smart enough to this automatically.

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -201,6 +205,129 @@ where
...
+models should be validated *before* validating this object.
+ */
+pub struct DblModelMorphism<'a, Map, Dom, Cod>(pub &'a Map, pub &'a Dom, pub &'a Cod);
+
+impl<'a, Map, Dom, Cod> DblModelMorphism<'a, Map, Dom, Cod>
+where
+    Map: DblModelMapping,
+    Dom: FgDblModel<Ob = Map::DomOb, Mor = Map::DomMor>,
+    Dom::ObGen: Clone,
+    Dom::MorGen: Clone,
```
epatters: Hopefully with #138 it is possible to remove these two bounds.

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -201,6 +205,129 @@ where
...
+        ob_errors.chain(mor_errors)
+    }
+}
+
+impl<Map, Dom, Cod> Validate for DblModelMorphism<'_, Map, Dom, Cod>
+where
+    Map: DblModelMapping,
+    Dom: FgDblModel<Ob = Map::DomOb, Mor = Map::DomMor>,
+    Dom::ObGen: Clone,
+    Dom::MorGen: Clone,
```
epatters: Ditto.

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -201,6 +205,129 @@ where
...
+ *
+ * Note that, by specifying a model morphism via its action on generators, we
+ * obtain for free that identities are sent to identities and composites of
+ * generators are sent to their composites in the codomain.
+*/
+#[derive(Debug, Error, PartialEq, Clone)]
+pub enum InvalidDblModelMorphism<Ob, Mor> {
+    /// Invalid data
+    #[error("Object `{0}` is mapped to an object not in the codomain")]
+    Ob(Ob),
```
epatters, line 304: With the recent factor #118, we are now properly distinguishing objects/morphisms from object/morphisms *generators* in the traits for finitely generated categories and models. Mappings between double models are defined on object/morphism generators and take values in objects/morphisms. Given that, I think this enum should have signature

```rust
pub enum InvalidDblModelMorphism<ObGen, MorGen> 
```

As a side benefit, this should clean up the validation code because you'll have to write `.into()` a lot less often.

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -306,11 +433,13 @@ where
...
 #[cfg(test)]
 mod tests {
-    use ustr::ustr;
+    use ustr::{ustr, Ustr};
 
     use super::*;
     use crate::stdlib::*;
     use crate::validate::Validate;
+    use nonempty::nonempty;
+    use std::collections::HashMap;
```
epatters, line 447: Ditto the organization of imports here.

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -363,4 +492,80 @@ mod tests {
...
+        let negloop = negative_loop(theory.clone());
+        let posfeed = positive_feedback(theory.clone());
+
+        // A good map from h to itself
+        let omap: HashMap<_, _> = [(ustr("x"), ustr("x"))].into_iter().collect();
+        let mmap: HashMap<_, _> = [(ustr(""), Path::Id(ustr("negative")))].into_iter().collect();
+        let f = DiscreteDblModelMapping {
+            ob_map: omap.into(),
+            mor_map: mmap.into(),
+        };
```
epatters, line 514: Not essential but perhaps worth adding a `new` method to `DiscreteDblModelMapping` that takes two hash maps as its arguments.

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -363,4 +492,80 @@ mod tests {
         assert!(im.has_mor(&Path::single(ustr("positive"))));
         assert!(im.has_mor(&Path::single(ustr("negative"))));
     }
+
+    #[test]
+    fn validate_model_morphism() {
```
epatters, line 503: Thanks so much for thorough tests!

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -363,4 +492,80 @@ mod tests {
...
+            ob_map: omap.into(),
+            mor_map: mmap.into(),
+        };
+        let dmm = DblModelMorphism(&f, &negloop, &negloop);
+        assert!(dmm.validate().is_err());
+
+        // A bad map from h to itself that is wrong for the ob (it is in the map
+        // but sent to something that doesn't exist) and for the hom generator
+        // (not in the map)
+        let bad_ob_assign: HashMap<ustr::Ustr, Ustr> =
```
epatters: You should be able to save a few characters here and below by writing `HashMap<_, _>`.

epatters (changes requested):
Thanks Kris, this is great! I left a few comments.

packages/catlog/src/dbl/model_morphism.rs
kris-brown: To my surprise, it wasn't possible to remove these bounds without adding `Eq+Clone` here:

```rust 
pub trait FgCategory: Category {
    /// The type of object generators.
    type ObGen: Eq + Clone + Into<Self::Ob>;

    /// The type of morphism generators. Often Mor = Path<Ob, MorGen>.
    type MorGen: Eq + Clone + Into<Self::Mor>;
```

Which was not part of #138

epatters: Yep, I think that's just a spot we missed in #138.

PR #156, opened 2024-09-08
Dev docs on discrete models of double theories

epatters:
This is my second attempt, replacing #108.


KevinDCarlson:
This looks like a plausible basis for boundaries of open models to me.

PR #158, opened 2024-09-09
Make backend more robust

olynch:
# Default checklist

- [ ] Are there any manual checks you performed that could be automated and become part of the CI?
- [ ] Are any issues relevant to the PR mentioned so that they will be automatically closed when the PR merges?


packages/backend/src/server.ts
```diff
@@ -84,6 +91,10 @@ export class Server {
             noServer: true,
         });
 
+        this.wss.on("error", (err) => {
+            console.log(err);
+        });
```
epatters: Having added sentry above, should this error handler still be here?

epatters (approved):

PR #162, opened 2024-09-10
Script to build forester plus upgrade to forester v4.3.1

### LINKED ISSUE #154: Upgrade to forester v4.3
The [release notes](https://www.jonmsterling.com/jms-005P.xml#tree-779) suggest that the theme needs to be upgraded, which will take a little work since we have made a few modifications. Perhaps we should generate a patch from v4.2 to v4.3 and then apply the patch to this repo. With luck there won't even be any conflicts.

olynch:
This PR adds a script (also available at https://github.com/olynch/ez-forester) which downloads and runs forester, so that new contributors can just run `./forester build` instead of having to muck around with ocaml or nix.

The PR also updates the theme to 4.3.1, to be compatible with the version of forester that is pulled in by `./forester`.

Closes #154.


epatters (changes requested):
Thanks! A problem is that this PR deletes all my theme extensions, which I've already had to manually recreate once before. Can we generate a patch from upstream 4.2 to 4.3.1 that we can apply?

olynch:
There seems to be a regression with how forester 4.3.1 handles TeX compilation where it no longer takes into account `.sty` files in the `assets` directory. I will report this to the forester issue tracker.

epatters (approved):
Thanks!

PR #175, opened 2024-09-21
Dev docs comparing (bi)modules over models of double theories

KevinDCarlson:
Dev doc update with a definition of left- and right- modules over models of double theories, careful explication of what a bimodule 1 -|-> M is for a multicategory and why it's weird, examples of left- and right- modules over multicategories and over proarrows and remark that this difference only exists for theories with nontrivial proarrows.


dev-docs/trees/dbl-000C.tree
```diff
@@ -0,0 +1,91 @@
...
+	{1(x)} & {1(y)} & {X(z)} \\
+	{1(x)} && {X(z)}
+	\arrow["{1(m)}", "\shortmid"{marking}, from=1-1, to=1-2]
+	\arrow["{=}"{description}, from=1-1, to=2-1]
+	\arrow["{X(n)}", "\shortmid"{marking}, from=1-2, to=1-3]
+	\arrow["{=}"{description}, from=1-3, to=2-3]
+	\arrow["{X(mn)}", from=2-1, to=2-3]
+    \end{tikzcd}}
+    }
+    \li{#{X} sends all cells of \dbl{D} to isomorphisms in \dbl{E}.}
```
epatters: Thanks Kevin, this is a useful writeup!

I'm confusing about this line though. Why should M/H being a module impose this condition on X? It would be bad if it did because it would seem to break the example of monoidal categories, where a module/instance should recover the notion of monoidal copresheaf, i.e., a lax monoidal functor into (Set, \times).

KevinDCarlson: Screenshot 2024-09-27 at 12 39 40 PM
The idea is to take beta = id_g in this diagram. Then the M^ell s are identities by def of a module as is F(alpha) so the left-hand side is an identity while the right-hand side is M(alpha). Am I missing something?

epatters: Since F=1 and the left actions are identities by assumption, I agree that this proves that M(\alpha) is the identity. But in the text you are saying that G=X maps all cells to isomorphisms, which seems like a different statement.

KevinDCarlson:
@epatters I've just added a bunch of explication of why the definitions are equivalent. This wasn't trivial, but it all eventually fell perfectly into place! Good fun.

epatters:
Thanks @KevinDCarlson, sounds excellent! I'll give this a look tomorrow.

epatters (approved):
This is great! That explication/unpacking of the short conceptual definition was the missing piece for me. It all comes together so nicely.

PR #204, opened 2024-10-02
Check monicity of free simple model morphisms, initialize hom search, monic search constraint

kris-brown:
This adds some related features which combine in a test which states (informally) `simple_paths(G, x, y)` is equal to `homomorphisms(walking_edge, G; initial=(walking_src -> x, walking_tgt -> y))` (in the case of all the object/morphism types being the same). This is done for an arbitrary model with lots of loops, but in general could be a nice thing to property-based-test in the future.

Now, a path is simple if it has no repeated generating morphisms (before it couldn't repeat objects except for beginning and end). A model morphism is simple if all generators are sent to simple paths.

This PR additionally adds a check for whether a morphism is monic (conditional on it being free and simple).

Initializing hom search merely adds some clauses to the `search` function which check, before looping through possible ob/mor assignments, if the ob/mor has already been preassigned. Future work will go into making this more efficient by playing with the order which the algorithm attempts assignments to improve performance.

### DIFF
--- a/packages/catlog/src/dbl/model.rs
+++ b/packages/catlog/src/dbl/model.rs
@@ -211,6 +211,11 @@ where
         self.category.add_mor_generator(f, dom, cod)
     }
 
+    /// Adds an equation to the model, making it not free.
+    pub fn add_equation(&mut self, key: Id, eq: PathEq<Id, Id>) {
+        self.category.add_equation(key, eq);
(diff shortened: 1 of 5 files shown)


epatters (approved):
Thanks Kris, and sorry for taking so long to review this PR!

PR #215, opened 2024-10-07
Cell menu items to move cell up or down

### LINKED ISSUE #150: Pop-up menu items to move cells up or down
Clicking the drag gripper next to a cell currently opens a menu with a single item, "delete." We should add the items "move up" and "move down."

I figure that drag and drop will be the preferred method to move cells for most users, but having an alternative for local moves might also be convenient.

hamidahoderinwale:
Closes #150

Added functions to move cells up and down by dragging arrow icons.


epatters (approved):
Thanks Hamidah! I squashed your list of commits into one. I also fixed a bug (not due to you!) that I noticed while playing around with the popup menu.

PR #216, opened 2024-10-07
issue 151 changes | duplicate cells 

### LINKED ISSUE #151: Pop-up menu item to duplicate a cell
Sometimes, instead of creating a new cell from scratch, it can be more convenient to duplicate an existing cell and then modify it. That will likely become more true over time as the data specified within a single cell becomes more complicated. Notion has this feature throughout their UI.

This issue is more subtle than #150 because it is incorrect to a duplicate a cell merely by deep copying it. In addition, the cell needs a new cell ID (a UUID) and in the case of formal cells, the formal content typically has its own UUID that must regenerated as well. One possible strategy is to extend the `CellConstructor` type with a new field providing a `duplicate` function, so that each cell type knows how to safely duplicate itself.

hamidahoderinwale:
# Default checklist

- [ ] Are there any manual checks you performed that could be automated and become part of the CI?
- [ ] Are any issues relevant to the PR mentioned so that they will be automatically closed when the PR merges?

Duplicate cells with copy icon

closes #151

### DIFF
--- a/packages/frontend/.env.development
+++ b/packages/frontend/.env.development
@@ -1,2 +1,2 @@
 VITE_APP_TITLE="CatColab [dev]"
-VITE_BACKEND_HOST=https://backend-next.catcolab.org
+VITE_BACKEND_HOST=https://backend.catcolab.org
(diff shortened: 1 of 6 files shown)


epatters:
Closing for now since we decided that some preparatory refactoring would make this easier.

PR #217, opened 2024-10-07
Temporary workaround for backend in dev environment

goretkin:
I started getting

image

and this change gets it working again. I am not sure if `backend-next.catcolab.org` is gone and whether `dev-docs/trees/dev-000A.tree` should also be updated, or if there is another fix I should do that keeps using `backend-next.catcolab.org`


epatters:
Caused by #214. Merging as a temporary workaround until `backend-next` is running as a different instance in AWS.

cc @shaoweilin

PR #225, opened 2024-10-18
Selection widget for theories

### LINKED ISSUE #52: Custom select widget for theories
As discussed in #45, we should have a custom "select" widget with type-ahead support.

hamidahoderinwale:
Closes #52 

Selection widget. Theory shows upon selection, division category for organization.

### DIFF
--- a/packages/frontend/src/document/theory_selector.tsx
+++ b/packages/frontend/src/document/theory_selector.tsx
@@ -1,0 +1,70 @@
+import { For, createMemo } from "solid-js";
+import type { TheoryLibrary } from "../stdlib/types";
+import type { ModelDocument } from "./types";
+import "./model_document_editor.css";
+import type { DocHandle } from "@automerge/automerge-repo";
+import type { TheoryMeta } from "../stdlib/types";
+
+interface TheorySelectorProps {
+    docHandle: DocHandle<ModelDocument>;
+    theories: TheoryLibrary;
+    doc: ModelDocument;
+}
+
+const TheorySelector = (props: TheorySelectorProps) => {
+    const groupedTheories = createMemo(() => {
+        const theories = Array.from(props.theories.metadata()).filter(
+            (meta) => meta.divisionCategory,
+        );
+        const grouped = new Map<string, TheoryMeta[]>();
+
+        for (const theory of theories) {
+            const category = theory.divisionCategory ?? "Other";
+            grouped.set(category, [...(grouped.get(category) || []), theory]);
+        }
+
+        return Array.from(grouped.entries()).sort((a, b) => a[0].localeCompare(b[0]));
+    });
+
+    return (
+        <div id="input-selections">
+            <For each={groupedTheories()}>
+                {([category, theories]) => (
+                    <div class="selection-items">
+                        <h4 id="division-category">{category}</h4>
+                        <For each={theories}>
+                            {(meta) => (
+                                <label>
+                                    <input
+                                        type="radio"
+                                        name="theory"
+                                        value={meta.id}
+                                        onchange={(evt) => {
+                                            const id = (evt.target as HTMLInputElement).value;
+                                            props.docHandle.change((model) => {
+                                                model.theory = id ? id : undefined;
+                                            });
+                                        }}
+                                    />
+                                    <div>
+                                        <ul>
+                                            <li id="selection-items">
+                                                {meta.name}{" "}
+                                                <div>
+                                                    <span class="description">
+                                                        {meta.description}
+                                                    </span>
+                                                </div>
+                                            </li>
+                                        </ul>
+                                    </div>
+                                </label>
+                            )}
+                        </For>
+                    </div>
+                )}
+            </For>
+        </div>
+    );
+};
+export default TheorySelector;
--- a/packages/frontend/src/document/theory_selector.tsx
+++ b/packages/frontend/src/document/theory_selector.tsx
@@ -1,9 +1,10 @@
-import { For, createMemo } from "solid-js";
-import type { TheoryLibrary } from "../stdlib/types";
-import type { ModelDocument } from "./types";
-import "./model_document_editor.css";
 import type { DocHandle } from "@automerge/automerge-repo";
-import type { TheoryMeta } from "../stdlib/types";
+import { For, createMemo } from "solid-js";
+
+import type { TheoryLibrary, TheoryMeta } from "../stdlib";
+import type { ModelDocument } from "./types";
+
+import "./theory_selector.css";
 
 interface TheorySelectorProps {
     docHandle: DocHandle<ModelDocument>;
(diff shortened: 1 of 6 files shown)


packages/frontend/src/stdlib/theories.tsx
```diff
@@ -20,6 +20,7 @@ stdTheories.add(
         id: "simple-olog",
         name: "Olog",
         description: "Ontology log, a simple conceptual model",
+        divisionCategory: "Data and Knowledge",
```
epatters: I know I said "Data and knowledge," but I think I prefer "Knowledge and data" instead. Sorry.

packages/frontend/src/document/model_document_editor.css
```diff
@@ -20,3 +20,44 @@
...
+}
+
+.description {
+    color: darkgray;
+    font-size: 11px;
+    text-align: left;
+    line-height: 1px;
+}
+
+#divisionCategoryHeader {
```
epatters: We're using "hypthen-case" in CSS files, so this should be "division-category-header."

packages/frontend/src/document/theoryselector.tsx
epatters, line 1: File names should mirror the naming of their primary exports, so if this file is providing `TheorySelector`, it should be called `theory_selector.tsx`.

packages/frontend/src/document/theoryselector.tsx
```diff
@@ -0,0 +1,83 @@
...
+interface TheorySelectorProps {
+    docHandle: any;
+    theories: TheoryLibrary;
+    doc: ModelDocument;
+}
+
+const TheorySelector = (props: TheorySelectorProps) => {
+    return (
+        <div id="input-selections" class="popup">
+            <h4 id="divisionCategoryHeader">Data and Knowledge</h4>
```
epatters: The categories shouldn't be hard-coded. We can discuss alternative designs in person.

epatters (changes requested):
Thanks Hamidah, this is starting to look quite good!

I would divide the existing theories as:

- Knowledge and data
  - Ologs
  - Schemas
- Life sciences
  - Regulatory networks
- System dynamics
  - Causal loop diagrams
  - Stock and flow

Regarding the styling:

- As you click different theories, the menu jumps around the screen, which is a surprising behavior
- IMO, menu items need slightly more spacing between them

packages/frontend/src/visualization/export_visualization.tsx
epatters, line 1: This file was accidentally committed from your other branch.

packages/frontend/package.json
```diff
@@ -22,7 +22,10 @@
         "@automerge/automerge-repo-network-websocket": "^1.2.1",
         "@automerge/automerge-repo-storage-indexeddb": "^1.2.1",
         "@automerge/prosemirror": "^0.0.13",
+        "@corvu/accordion": "^0.2.3",
+        "@corvu/dialog": "^0.2.3",
         "@corvu/disclosure": "^0.2.0",
+        "@corvu/drawer": "^0.2.2",
```
epatters: Looks like there are extra dependencies here: `accordion` and `drawer` aren't being used, right?

packages/frontend/package.json
```diff
@@ -36,6 +39,7 @@
         "@viz-js/viz": "^3.7.0",
         "backend": "workspace:*",
         "catlog-wasm": "link:../catlog-wasm/pkg",
+        "computed-style-to-inline-style": "^4.0.0",
```
epatters: Also crept in from the other PR.

packages/frontend/src/document/model_document_editor.css
```diff
@@ -8,15 +8,89 @@
...
+    border-top: transparent;
+    border-left: transparent;
+}
+
+#selection-items:hover {
+    color: rgb(0, 144, 184);
+    cursor: pointer;
+}
+
+ul {
```
epatters: Defining `ul` and `ui li` without any constraining selectors means that they will apply by default to *all* unordered lists in the app, which we don't want.

You might want something like:

```css

.theory-selector {
  & ul {
    ...
  }
  & ul li {
    ...
  }
}
```

packages/frontend/src/document/theory_selector.tsx
```diff
@@ -0,0 +1,76 @@
+import { For, createMemo } from "solid-js";
+import type { TheoryLibrary } from "../stdlib/types";
+import type { ModelDocument } from "./types";
+import "./model_document_editor.css";
+import type { TheoryMeta } from "../stdlib/types";
+
+interface TheorySelectorProps {
+    docHandle: any;
```
epatters: This should have the type `DocHandle`, which you can import from `@automerge/automerge-repo`.

packages/frontend/src/document/model_document_editor.css
```diff
@@ -8,15 +8,89 @@
     justify-content: space-between;
 }
 
-.model-theory select {
-    appearance: none;
+.selector {
```
epatters: Rename to `theory-selector` for extra specificity.

Also, I think it cleanest that the CSS specific to `theory_selector.tsx` be in a separate file called `theory_selector.css`.

packages/frontend/src/document/model_document_editor.css
```diff
@@ -8,15 +8,89 @@
...
+    right: 0;
+    bottom: 0;
+    background-color: rgba(0, 0, 0, 0.5);
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    z-index: 1000;
+}
+
+.popup {
```
epatters: A class called `popup` is already defined in `index.css`: it's confusing, and possibly even ill-defined, which one will take precedence.

1. We should have generic class `popup`, to be used throughout the app for consistent styling of popups. That should be defined in `index.css`
2. Any styling specific to this widget can be defined here, but if needed it should be qualified somehow (see above)

packages/frontend/src/document/model_document_editor.css
```diff
@@ -8,15 +8,89 @@
...
+    padding: 6px;
+}
+
+#division-category {
+    font-weight: 600;
+    padding-bottom: 1px;
+    color:  #53b6b2;
+}
+
+.overlay {
```
epatters: See my comment on `popup`: similar concerns apply to `overlay`

packages/frontend/src/document/theory_selector.tsx
```diff
@@ -0,0 +1,76 @@
...
+    theories: TheoryLibrary;
+    doc: ModelDocument;
+}
+
+const TheorySelector = (props: TheorySelectorProps) => {
+    const groupedTheories = createMemo(() => {
+        const theories = Array.from(props.theories.metadata()).filter(
+            (meta) => meta.divisionCategory,
+        );
```
epatters: Is this filtering needed? Seems like you're already checking for theories without a `divisionCategory` below.

packages/frontend/src/document/theory_selector.tsx
```diff
@@ -0,0 +1,76 @@
...
+        const grouped = new Map<string, TheoryMeta[]>();
+
+        theories.forEach((theory) => {
+            const category = theory.divisionCategory ?? "Uncategorized";
+            if (!grouped.has(category)) {
+                grouped.set(category, []);
+            }
+            grouped.get(category)?.push(theory);
```
epatters: I believe you can replaced the above three lines with a single line:

```javascript
(grouped[category] ??= []).push(theory);
```

packages/frontend/src/documentation/.gitignore
epatters, line 1: I assume everything in this directory was added by accident?

epatters (changes requested):
Thanks Hamidah, the widget looks wonderful! I've suggested some ways to improve the code.

epatters (approved):

PR #227, opened 2024-10-27
Causal loop diagrams with unknown link polarities

### LINKED ISSUE #206: Causal loop diagrams with indeterminates
Implement a new theory whose models are causal loop diagrams with links of positive, negative, or unknown/indeterminate polarity.

The double theory is already implemented in the core as [`th_nullable_signed_category`](https://github.com/ToposInstitute/CatColab/blob/30b9a1b6df06f03a06d20073729c5689ee669066/packages/catlog/src/stdlib/theories.rs#L51) and in the Wasm bindings as `ThNullableSignedCategory`, so the work here is to add the theory to the frontend, perhaps with a new glyph for the indeterminate link, such as a question mark near the arrowhead.

quffaro:
This PR resolves #206 on causal loop diagrams with an "indeterminate glyph."

### DIFF
--- a/packages/frontend/src/stdlib/theories.tsx
+++ b/packages/frontend/src/stdlib/theories.tsx
@@ -263,63 +263,64 @@ stdTheories.add(
 );
 
 stdTheories.add(
-	{
-		id: "nullable-causal-loop",
-		name: "Nullable Causal Loop",
-	},
-	(meta) => {
-		const thNullableSignedCategory = new catlog.ThNullableSignedCategory();
-	    return new Theory({
-			    ...meta,
-			    theory: thNullableSignedCategory.theory(),
-			    onlyFreeModels: true,
-			    types: [
-				    {
-				        tag: "ObType",
-					      obType: { tag: "Basic", content: "Object" },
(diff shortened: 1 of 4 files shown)


packages/frontend/src/stdlib/theories.tsx
```diff
@@ -262,6 +262,66 @@ stdTheories.add(
...
+	    return new Theory({
+			    ...meta,
+			    theory: thNullableSignedCategory.theory(),
+			    onlyFreeModels: true,
+			    types: [
+				    {
+				        tag: "ObType",
+					      obType: { tag: "Basic", content: "Object" },
+					      name: "Type",
+					      description: "Type or class of things",
```
epatters: For consistency, all of these names and descriptions should mirror those for ordinary causal loop diagrams where appropriate. E.g., this first name should be "Variable".

epatters (changes requested):
Thanks Matt, this is great!

To fix the CI errors, try running `pnpm run format` and `pnpm run lint` from `packages/frontend`.

PR #232, opened 2024-10-29
Remove broken linear path in layout for links in stock-flow diagrams

### LINKED ISSUE #231: Visualization of links to non-adjacent stocks is broken
Try building an SIR model. The link from R to infection doesn't get its source moved to R! Seems to be generalizable to any link that's supposed to jump more than one stock, and it happens on both next. and catcolab.

KevinDCarlson:
Fixes #231 by simply removing the misbehaving linear-path function for building links that would have been semantically bugged even if the implementation were corrected. Here's an example of a messy link situation after the fix:

Screenshot 2024-10-28 at 6 07 13 PM

### DIFF
--- a/packages/frontend/src/analysis/stock_flow_diagram.tsx
+++ b/packages/frontend/src/analysis/stock_flow_diagram.tsx
@@ -115,6 +115,7 @@ function StockFlowSVG(props: {
                     },
                 },
                 ({ srcId, tgtId }) => {
+                    console.log("nodeMap", nodeMap);
                     const srcNode = nodeMap.get(srcId);
                     const tgtEdge = edgeMap.get(tgtId);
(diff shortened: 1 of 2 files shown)


packages/frontend/src/analysis/stock_flow_diagram.tsx
```diff
@@ -115,17 +115,15 @@ function StockFlowSVG(props: {
                     },
                 },
                 ({ srcId, tgtId }) => {
+                    console.log("nodeMap", nodeMap);
```
epatters: Remove this line.

epatters (changes requested):
Thanks Kevin! To fix the CI, run `pnpm run format` and `pnpm run lint`. See also one small comment.

epatters (approved):

PR #233, opened 2024-10-29
Add nix config for dev server

shaoweilin:
- add nix deploy node for catcolab-next
- fix CORS duplicate header problem in nginx
- fix catcolab service script for deploying frontend with nix
- re-encrypt secrets to allow access by catcolab-next public key
- update dev backend URL
- update backend dev docs


epatters (approved):
Thanks Shaowei! I tested that this works for me locally.

PR #248, opened 2024-11-10
Diagrams in models of discrete double theories

### LINKED ISSUE #236: Data structure for diagrammatic instances of models of discrete theories
Implement a data structure in `catlog` for diagrams in a model of a discrete double theory. For the purposes of this issue, the diagram will be interpreted as presenting an instance of the model, in the sense outlined in #240. (Working out the mathematics of #240 is *not* required to implement this feature, only to fully understand it.)

We should simplify by assuming that the diagram is free, i.e., the indexing model is a free model. Thus, this data structure should consist of:

- a graph, indexing the diagram
- a graph mapping (`GraphMapping`) with intended domain the above graph

KevinDCarlson:
Resolves #236 by implementing discrete diagrams which own their domain, a free model, and their mapping, along with simple validation that the mapping is actually defined everywhere on the domain.

### DIFF
--- a/packages/catlog/src/dbl/mod.rs
+++ b/packages/catlog/src/dbl/mod.rs
@@ -5,5 +5,6 @@ pub mod diagram;
 pub mod pasting;
 
 pub mod model;
+pub mod model_diagram;
 pub mod model_morphism;
 pub mod theory;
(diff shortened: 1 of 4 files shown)


epatters (changes requested):
Thanks Kevin! I'll review this PR more thoroughly soon but a few quick comments now:

- To fix the CI failures, run `cargo fmt` and `cargo clippy`
- Remove the spurious `println` that crept into `model_morphism.rs` and the whitespace added to `column.rs`

KevinDCarlson:
@epatters thanks, should be fixed!

epatters (approved):

Issue #242, opened 2024-11-03
Traits for monoidal categories and PROs


epatters:
Implement the traits:

- `MonoidalCategory` for (non-symmetric) monoidal categories, extending the trait `Category`
- `PRO` for multisorted PROs ("product categories"), i.e., a strict monoidal category whose monoid of objects is freely generated by a finite set of distinguished objects

The product operations should have an unbiased-first API.

epatters:
Closing in lieu of the new approach #561.

PR #260, opened 2024-11-15
Strict monoidal category, PRO, AugSimplexCat

kris-brown:
Adds features described in https://github.com/ToposInstitute/CatColab/issues/242

Also adds a new Finset implementation: `SingletonSet` which has one element of unit type.

As part of AugSimplexCat I effectively implemented composition in SkelFinSet, which wasn't implemented anywhere else to my knowledge. Eventually we'll want to make that category explicit somewhere and move this code there.


epatters:
Closing since we have a new approach, started in #561, for theories of monoidal categories and the like.

PR #269, opened 2024-11-23
Theory-specific help

hamidahoderinwale:
Theory-specific help
- Help button directs to main help page when a theory hasn't been selected
- Otherwise it directs to a theory-specific help page

(PR is incomplete, routing isn't working)

### DIFF
--- a/packages/frontend/src/help/theory_documentation/theory_file_parser.tsx
+++ b/packages/frontend/src/help/theory_documentation/theory_file_parser.tsx
@@ -1,0 +1,46 @@
+import { readdirSync, statSync, writeFileSync } from "node:fs";
+import { join, resolve } from "node:path";
+import { For, type JSX } from "solid-js";
+import type { Plugin } from "vite";
+
+export const documentationFileParser = (): Plugin => {
+    return {
+        name: "documentation-file-parser",
+        async buildStart() {
+            processFiles();
+        },
+
+        configureServer(server) {
+            server.watcher.on("change", (filePath) => {
+                if (filePath.includes("./theory_documentation")) {
+                    processFiles();
+                }
+            });
+        },
+    };
+};
+
+export const theoryDocumentationPosts: { slug: string }[] = [];
+
+export const processFiles = () => {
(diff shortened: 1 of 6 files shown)


packages/frontend/src/model/theory_selector.tsx
```diff
@@ -45,6 +45,8 @@ export function TheorySelectorDialog(
     );
 }
 
+export const [selectedTheory, setSelectedTheory] = createSignal<string | undefined>(undefined);
```
epatters, line 48: I don't know if creating a signal outside of a component, for example at the top level like here, will work. More importantly, we shouldn't need to create a new signal at all. The selected theory is already available from the `LiveModelDocument`. See below.

packages/frontend/src/page/toolbar.tsx
```diff
@@ -33,13 +34,23 @@ const Brand = () => (
...
+/** Button that navigates to the root help page or theory documentation page. */
 export function HelpButton() {
     const navigate = useNavigate();
-
-    return (
-        <IconButton onClick={() => navigate("/help")} tooltip="Get help about CatColab">
-            <CircleHelp />
-        </IconButton>
-    );
+    if (selectedTheory() == null) {
```
epatters, line 40: Instead of using `selectedTheory()`, create a `TheoryHelpButton` that takes the theory ID (or undefined) as a prop. Then pass the theory as `props.liveModel.theory()` from inside the `ModelDocumentEditor`.

packages/frontend/src/help/theory_documentation/theory_file_parser.tsx
```diff
@@ -0,0 +1,46 @@
+import { readdirSync, statSync, writeFileSync } from "node:fs";
+import { join, resolve } from "node:path";
```
epatters, line 2: This will not work because you're using Node-specific APIs, which are not available in the browser. That's why the CI is failing. We can discuss alternative approaches.

epatters (changes requested):
Thanks for getting this started, Hamidah! I've provided some initial feedback below. Let's discuss more in person.

Issue #272, opened 2024-11-25
Backlink to anonymous model


KevinDCarlson:
If you build a diagram in an untitled model, the usual backlink is not populated due to an empty textbox.

KevinDCarlson:
<img width="1140" alt="Screenshot 2024-11-25 at 1 36 44 PM" src="https://github.com/user-attachments/assets/89030870-cab0-4926-a26a-575aa9d240d4">
<img width="1166" alt="Screenshot 2024-11-25 at 1 37 05 PM" src="https://github.com/user-attachments/assets/eb90830f-5536-48ca-b607-f9b21b76cfb4">

PR #273, opened 2024-11-25
Render empty model name as Untitled in backlinks

kris-brown:
Closes https://github.com/ToposInstitute/CatColab/issues/272 in as simple a way as possible: an empty model name defaults to "Untitled". 

If we end up discovering this code needs to get used in multiple places, it could be its own `name_or_untitled` method. Unsure if that's necessary at this point.


epatters (approved):
Thanks Kris!

PR #274, opened 2024-11-26
Categories with scalars

KevinDCarlson:
Implements the theory for "categories with scalars", which are pairs of categories on the same objects in which one category acts on the other, that is, categories sliced over the monoid freely generated by an idempotent, or...Ah, yes, I just remembered that this almost promonads, except for the unitor relations. So that's worth thinking over, at least for me.

Scalar morphisms are decorated at the theory level with a little proportionality twiddle. Unfortunately it looks at a glance like we don't yet have decorated arrows implemented at the diagram level, which is where this would be most relevant. I'll see about taking a glance at that, though other things are liable to be more urgent.


epatters (approved):
Thanks Kevin!

PR #279, opened 2024-11-29
Anon objects, scalar constants, and more DEC operators in Decapodes integration

quffaro:
Just need to move the `my_generate` function to inside the `System` struct and retest. i'll have to clean up `test` by separating some of functionality being tested


packages/algjulia-service/test/test_sim.jl
epatters, line 1: This file looks to be accidentally committed.

quffaro:
tested with this [catcolab analysis](http://localhost:5173/analysis/019379a1-39e4-76f2-b82c-99947ad750c8)

packages/frontend/src/stdlib/analyses/decapodes.tsx
```diff
@@ -85,6 +85,7 @@ export function Decapodes(props: DiagramAnalysisProps<DecapodesContent>) {
     const [result, { refetch: rerunSimulation }] = createResource(maybeKernel, async (kernel) => {
         // Construct the data to send to kernel.
         const simulationData = makeSimulationData(props.liveDiagram, props.content);
+        console.log(JSON.parse(JSON.stringify(simulationData))); // XXX
```
epatters: This line should be removed.

packages/algjulia-service/test/test_sim.jl
epatters: Unresolving because the file is still present. Am I correct that this file should be removed? It is machine generated, and usually such files should not be committed.

packages/algjulia-service/src/decapodes.jl
```diff
@@ -170,53 +203,91 @@ end
 
 This returns a Decapode given a jsondiagram and a theory.
 """
-function Decapode(diagram::AbstractVector{JSON3.Object}, theory::Theory)
+function Decapode(diagram::AbstractVector{JSON3.Object}, theory::Theory; scalars=[])
     # initiatize decapode and its mapping between UUIDs and ACSet IDs
-    pode = SummationDecapode(parse_decapode(quote end))
+    pode = SummationDecapode(parse_decapode(quote end));
```
epatters, line 208: Not important, but I'm wondering why all these lines end with semicolons.

packages/algjulia-service/src/decapodes.jl
```diff
@@ -226,86 +297,44 @@ struct SimResult
...
-function SimResult(sol::ODESolution, mesh::EmbeddedDeltaDualComplex2D)
+function SimResult(sol::ODESolution, system::System)
+
+    points = collect(values(system.mesh.subparts.point.m));
 
-    points = collect(values(mesh.subparts.point.m));
+    xlen = 51; ylen = 51;
+
+    statevars = infer_state_names(system.pode)
+    statevar = length(statevars) == 1 ? first(statevars) : error("$statevars must be length one")
```
epatters: Seems like the `System` struct should store the list of statevars (or the unique state variable, if that is being enforced) so it doesn't have to be inferred twice.

packages/algjulia-service/src/decapodes.jl
```diff
@@ -226,86 +297,44 @@ struct SimResult
...
+    SimResult(sol.t, state_vals, 0:xlen-1, 0:ylen-1)
 end
 # TODO generalize to HasDeltaSet
 
-struct System
-    pode::SummationDecapode
-    mesh::HasDeltaSet
-    dualmesh::HasDeltaSet
-    init::Any # TODO specify. Is it always ComponentVector?
+function generate(s, my_symbol; hodge=GeometricHodge())
```
epatters: Do you still need this function? It looks to be superseded by the closure `sys_generate` above, but I'm not sure.

epatters (changes requested):
Thanks Matt! A few questions and comments below.

packages/algjulia-service/src/decapodes.jl
quffaro: I've tested by sending text to terminal, so I left semicolons to suppress output

packages/algjulia-service/src/decapodes.jl
quffaro: Makes sense!

packages/algjulia-service/src/decapodes.jl
quffaro: No it's no longer necessary

quffaro:
The `test_sim.jl` was not automatically generated; it's produced by open-write code we specify so we can get line numbers for errors in our generated code.

PR #283, opened 2024-12-04
Improved Decapodes integration: domains, meshes, initial conditions, plot variables

epatters:
We can now do the Navier-Stokes vorticity equation in addition to the diffusion/heat equation. To accomplish this, we had to generalize the frontend support and Julia interop in several directions. A Decapodes simulation now analysis now requires that you specifiy:

- the geometric domain: rectangle or sphere)
- the mesh that discretizes the domain: several choices for both rectangle and sphere
- initital/boundary conditions for specific variables: again, several fixes choices available
- which variables to plot

The code is a quite raw in certain places and will need to be cleaned up later.


quffaro:
This latest push anticipates more integrations with AlgebraicJuliaService more and, concretely, generalizings *some* code from the Decapodes integration. AlgebraicJuliaService specifies integrations in an `AlgebraicJuliaIntegration` ADT and then has submodules through a basic yet vague interface specify how they will build theories/models/diagrams for CatColabsumption.

While I do not think this is entirely complete, I've

* added a `Geometry` struct to contain the domain, mesh, and dualmesh.
* used `ThDecapode <: AlgebraicJuliaIntegration` to tag generic-sounding structs like `Theory`, which is currently mapping names to `TheoryElements`, which may be `Obs` or `Homs` each responsible for storing data consumed from the incoming JSON. This interface is vague but, barring one failing test for unrelated reasons, does not currently break functionality.
* threaded the `system::PodeSystem` variable further into functions, which allows us to carry more data further into functions, keeping the number of method arguments smaller
* housekeeping such as reorganizing code into functions and relevant functionality into distinct files, added documentation, cleaned up tests

I'd like to generalize the methods in `Theory` so that further integration points would have an interface they integrate with, but I'm not sure if these packages would require different data structures (specifically, would the fields of `Theory` be different, or can we can away with type parameterizing the struct by `T<:AlgebraicJuliaIntegration`?)

epatters:
Thanks @quffaro! I like this reorganization and refactor.

packages/algjulia-service/src/decapodes-service/DecapodesService.jl
```diff
@@ -0,0 +1,179 @@
...
+# funcitons for geometry and initial conditions
+include("geometry.jl")
+include("initial_conditions.jl")
+include("theory.jl") ## theory-building
+include("model.jl") ## model-building (is this actually a diagram of a model?)
+
+struct PodeSystem
+    pode::SummationDecapode
+    plotvar::Vector{Symbol}
+    scalars::Dict{Symbol, Any} # closures # TODO rename scalars => anons
```
epatters: Why this TODO? The scalar constants are unrelated to the anonymous variables.

packages/algjulia-service/src/decapodes-service/DecapodesService.jl
```diff
@@ -0,0 +1,179 @@
...
+    init::ComponentArray # TODO Is it always ComponentVector?
+    generate::Any
+    uuiddict::Dict{Symbol, String}
+end
+export PodeSystem
+
+"""
+Construct a `PodeSystem` object from a JSON string.
+"""
+function PodeSystem(json_string::String,hodge=GeometricHodge())
```
epatters: I'd refactor this to have:

```julia
function PodeSystem(json_object::AbstractDict, ...)
  [...]
end

function PodeSystem(json_string::AbstractString, ...)
  json_object = JSON3.read(json_string)
  PodeSystem(json_object, ...)
end

File: packages/algjulia-service/src/decapodes-service/DecapodesService.jl [hunk 3]
```diff
@@ -0,0 +1,179 @@
...
+    json_object = JSON3.read(json_string);
```
epatters: All the semi-colons in the body of this function should be removed: that you sometimes copy-paste into the shell to test does not justify deviating from the standard Julia code style.

packages/algjulia-service/src/decapodes-service/DecapodesService.jl
```diff
@@ -0,0 +1,179 @@
...
+    dot_rename!(decapode)
+    uuid2symb = uuid_to_symb(decapode, vars);
+
+    # plotting variables
+    plotvars = [uuid2symb[uuid] for uuid in json_object[:plotVariables]];
+    
+    # extract the domain in order to create the mesh, dualmesh
+    geometry = Geometry(json_object);
+
+    # initialize operators
```
epatters, line 79: Not a job we should solve in this PR, but IMO all this nitty-gritty stuff initializing operators should be upstreamed to Decapodes. The glue layer in AlgebraicJuliaService should be as minimal as possible to reduce the maintenance burden for CatColab, which is not written in Julia.

packages/algjulia-service/src/decapodes-service/initial_conditions.jl
```diff
@@ -42,10 +43,10 @@ GaussianIC(r::Rectangle) = GaussianIC(r, GaussianData(r))
 TaylorVortexIC(d::Sphere) = TaylorVortexIC(d, TaylorVortexData())
 
 
-function initial_conditions(jo::JSON3.Object, uuid2symb::Dict{String, Symbol}, mesh_builder::String)
-    ic_specs = jo[:initialConditions];
+function initial_conditions(json_object::JSON3.Object, geometry::Geometry, uuid2symb::Dict{String, Symbol})
```
epatters: Should dispatch on `AbstractDict` instead of `JSON3.Object` to avoid unnecessary coupling to JSON3.jl.

packages/algjulia-service/src/decapodes-service/model.jl
```diff
@@ -0,0 +1,87 @@
+## MODEL BUILDING
+
+function add_to_pode! end
+export add_to_pode!
+
+function add_to_pode!(d::SummationDecapode, 
+        vars::Dict{String, Int}, # mapping between UUID and ACSet ID
+        theory::Theory, 
+        content::JSON3.Object,
```
epatters: Ditto about dispatching on `AbstractDict` here and below.

packages/algjulia-service/src/decapodes-service/plotting.jl
```diff
@@ -9,3 +9,20 @@ function Base.reshape(::Heatmap, data)
     l = floor(Int64, sqrt(length(data)))
     reshape(data, l, l)
 end
+
+# TODO size fixed
```
epatters: Commented code should be fixed and uncommented, or removed. I have no preference about which. If we're not producing heat maps in Julia, we could just delete this stuff.

packages/algjulia-service/src/decapodes-service/theory.jl
```diff
@@ -0,0 +1,138 @@
...
+        "Δ⁻¹" => :Δ⁻¹
+        "d*" || "d̃₁" => :dual_d₁
+        # \star on LHS
+        "⋆" => :⋆₁
+        "⋆⁻¹" || "⋆₀⁻¹" => :⋆₀⁻¹
+         # => :⋆₀⁻¹
+        # \bigstar on LHS
+        "★" || "★⁻¹" => :⋆₁
+         # => :⋆₀⁻¹
+        "diffusivity" => :diffusivity
```
epatters, line 28: Are we still hard-coding the list of constants? That should be fixed, though I won't insist that it happen in this PR if it's feels like too much work.

packages/algjulia-service/src/decapodes-service/theory.jl
```diff
@@ -0,0 +1,138 @@
...
+struct Theory{T<:ThDecapode}
+    data::Dict{String, TheoryElement}
+    function Theory(::T) where T
+        new{T}(Dict{String, TheoryElement}())
+    end
+end
+export Theory
+
+# TODO engooden
+Base.show(io::IO, theory::Theory) = show(io, theory.data)
```
epatters: Would suggest just removing this method if you're not going to improve it.

packages/algjulia-service/src/decapodes-service/theory.jl
```diff
@@ -0,0 +1,138 @@
...
+
+Base.values(theory::Theory) = values(theory.data)
+
+function add_to_theory! end; export add_to_theory!
+
+function add_to_theory!(theory::Theory{T}, content::Any, type::ObType) where T
```
epatters: `content` is `AbstractDict`?

epatters (commented):
Refactor is looking good, Matt! I know its WIP but I left some comments.

packages/algjulia-service/src/decapodes-service/DecapodesService.jl
quffaro: you're right, I'll correct this

packages/algjulia-service/src/decapodes-service/plotting.jl
quffaro: IIRC we rely on Julia to format the data for plotting, so I can see this design being useful if we want to specify different ways of plotting which would require different output formats. As of now it's unused and will be deleted.

packages/algjulia-service/src/decapodes-service/theory.jl
quffaro: I'll remove this and test, but to keep pushes more frequent I've pushed the rest of the feedback

packages/algjulia-service/src/decapodes-service/DecapodesService.jl
quffaro: My intuition is that operators may vary across simulations, so while the `default_dec_matrix_generate` captures a lot of the operators necessary to do physics simulations, it may still be necessary to have bespoke operators.

Consider
* [budyko_sellars_halfar](https://github.com/AlgebraicJulia/Decapodes.jl/blob/586915f0c24519ef0522dad29c564c1f7ebb7282/examples/climate/budyko_sellers_halfar.jl#L141) where the $$\sharp$$ operator is bespoke
*  [cism](https://github.com/AlgebraicJulia/Decapodes.jl/blob/586915f0c24519ef0522dad29c564c1f7ebb7282/docs/src/cism/cism.md#:~:text=x%20%2D%3E%20norm.(x)-,%3A%E2%99%AF%20%20%3D%3E%20begin,-sharp_mat%20%3D%20%E2%99%AF_mat) where the $$\sharp$$ operator is bespoke, in the sense that it's defined in the `generate` statement and not "OOTB" in Decapodes, however I think it's defined in `default_dec_matrix_generate` now.
* [bc](https://github.com/AlgebraicJulia/Decapodes.jl/blob/586915f0c24519ef0522dad29c564c1f7ebb7282/docs/src/bc/bc_debug.md#:~:text=%3A%E2%88%82C%20%3D%3E%20x%20%2D%3E%20begin%0A%20%20%20%20%20%20boundary%20%3D%20boundary_inds(Val%7B0%7D%2C%20sd)%0A%20%20%20%20%20%20x%5Bboundary%5D%20.%3D%200%0A%20%20%20%20%20%20x%0A%20%20%20%20end) where the boundary operator is defined inline.

Some of these could probably be up-streamed if they aren't already, but I think it remains to be a possibility that there are operators with nontrivial *bespoke* implementations that have historically been defined Julia-side. I'm not sure how CatColab will handle those cases, but they may be out of scope.

@lukem12345 please correct my understanding.

packages/algjulia-service/src/decapodes-service/theory.jl
epatters: Marking as resolved since this is now captured by #284.

packages/algjulia-service/src/decapodes-service/DecapodesService.jl
epatters: Resolving since separated into #286.

PR #287, opened 2024-12-09
Reorganize code in AlgebraicJulia service

quffaro:
This PR completes some code organization done shortly after the December 2024 Hackathon, attempting to rename objects to be closer to the theory subtending CatColab, including:
* renaming `theory.jl` to `model.jl` and appropriate renamings
* renaming `model.jl` to `diagram.jl` and appropriate renamings
* moving general code to the main source file to suggest how future AlgebraicJulia packages might integrate through AlgebraicJuliaService.jl


packages/algjulia-service/src/AlgebraicJuliaService.jl
```diff
@@ -7,27 +7,72 @@ using Reexport
 # the intent is that this is an interface for AlgebraicJulia code to interoperate with CatColab 
 @data AlgebraicJuliaIntegration begin
     ThDecapode()
-end
-export ThDecapode
+end; export ThDecapode
```
epatters: TBH, I prefer the previous style, which feels like more idiomatic Julia despite using slightly more whitespace.

packages/algjulia-service/src/AlgebraicJuliaService.jl
```diff
@@ -7,27 +7,72 @@ using Reexport
...
+@active patterns are MLStyle-implementations of F# active patterns that forces us to work in the Maybe/Option pattern. 
+Practically, yet while a matter of opinion, they make @match statements cleaner; a statement amounts to a helpful pattern
+name and the variables we intend to capture.
+=# 
+@active IsObject(x) begin; x[:tag] == "object" ? Some(x) : nothing end
+@active IsMorphism(x) begin; x[:tag] == "morphism" ? Some(x) : nothing end
+export IsObject, IsMorphism
+
+""" Obs, Homs """
+abstract type ModelElementValue end
```
epatters: Why use an abstract type here instead of `@data` as used for `ElementTag` above? Also, I'd suggest a more uniform naming scheme, such as:

- `ElementTag` -> `ModelElementTag`
- `ModelElementValue` -> `ModelElement`

epatters (commented):
Thanks Matt, this is a good refactor! I left a few very minor comments.

I assume that the Julia tests are passing. Have you tested in the frontend, say with a simple diffusion, to be sure everything still works end-to-end? That should be part of the workflow going forward since we don't have automated tests for the UI itself.

quffaro:
image
Verifying the simulation works on the front-end

packages/algjulia-service/src/AlgebraicJuliaService.jl
quffaro: I've changed them to use the `@data` macro, and renamed `ElementTag` -> `ModelElementTag`, but I've retained the naming `ModelElement` so its `val` parameter is still parameterized over something named `ModelElementValue`

epatters (approved):

PR #289, opened 2024-12-10
Script to build sys image for AlgebraicJulia service

quffaro:
This PR allows users to run a script for building a sysimage and installing it as a kernel. Currently this script builds a sysimage with all AlgebraicJuliaService dependencies, which can be time-consuming on slower machines. (a Thinkpad P15 with an i9 CPU can build this in ~2 minutes).

I need to make sure the Jupyter server chooses the installed kernel.


epatters:
Thanks for getting this started! Are you planning to add the test suites for CombinatorialSpaces and Decapodes to the sys image to the precompile script, as explained here? https://julialang.github.io/PackageCompiler.jl/dev/sysimages.html#tracing

quffaro:
I've added a `precompile.jl` file loading the test statements, and verified on the frontend that this runs, with the AlgebraicJuliaService loading in ~9 seconds. Should the `run_jupyter_service`, which is currently hardcoding julia-1.11 kernel, be dynamic?

packages/frontend/src/stdlib/analyses/jupyter.ts
```diff
@@ -39,9 +39,14 @@ export function createKernel(
...
+    // XXX: How do we know...
+    // - which Julia version to use?
+    // - whether to use the standard kernel or one with our custom sys image?
+    // For now, we are omitting the kernel name completely and thus assuming
+    // that the correct default kernel has been set on the Jupyter server.
+    // Obviously this approach will not extend to multiple languages.
     return createKernel(serverOptions, {
-        // XXX: Do I have to specify the Julia version?
-        name: "julia-1.11",
+        //name: "julia-1.11",
```
epatters, line 49: A comment on my own code: clearly this isn't ideal, but since in practice the only way to launch a Jupyter server that won't reject the frontend on cross-origin grounds is to run our script, this shouldn't be any less reliable than before, and it has the advantage of not baking the Julia version into the bundled JS.

epatters (approved):
Thanks Matt! I've confirmed that I can build the sys image and use it in the frontend.

At least on my machine, the performance benefits seem marginal, unfortunately. Loading the AlgJulia service is noticeably faster, but I can't tell a difference when running the simulation, which is far more time consuming. I haven't done a proper benchmark, though.

quffaro:
I think proper benchmarks are in order (I've created [this issue](https://github.com/ToposInstitute/CatColab/issues/297) to track it), but the sysimage won't improve simulation speed, just precompilation time. The simulation is generated code, so performance improvements must come from Decapodes itself.

epatters:
Right, at most it would improve simulation speed on the *first* simulation, which I didn't observe,

PR #290, opened 2024-12-11
Permissions for model sharing 

hamidahoderinwale:
This is a start to adding permissions for model documents. The goal for this initial PR is to have a link that can be copied when the widget is selected. Permission to edit or view will be handled in another PR. 

This is a WIP.


epatters:
Thanks for starting this work, Hamidah. I'm going to close it because, as we discussed, I'm currently working on handling this in a different way.

PR #301, opened 2024-12-20
Dev docs on free models of double theories

epatters:
It's been bugging me that, while we frequently single out the free models of various double theories, we haven't written down anywhere the definition of a free model of an arbitrary double theory. This is a first attempt to do so.

If this seems right, I'll probably expand this content in a future PR by fleshing out the examples, but I wanted to get some feedback first.


epatters:
The math could still benefit from another pair of eyes, but since it's no longer live for me, I'll go ahead and merge. We can revisit if/when it becomes pressing to understand this more deeply.

PR #315, opened 2025-01-03
Plot non-state variables in Decapodes

quffaro:
This feature allows us to plot non-state variables.

Currently, state variables are able to be retrieved from the ODESolution object, and therefore plotted, because they are passed in as initial conditions to the solver. However the variables specified for plotting are stored in a separate field, `plotVars`, and therefore may not have initial conditions.

The design decision made in this draft PR intends to also endow variables in plotVars with a default initial condition called "Zeros." The `initial_conditions` function dispatches on `Zeroes` initial condition type, returning `zeros(geometry.dualmesh[:point])`.

At the time of opening this draft PR, _all variables_ in the `uuid2symb` mapping are given Zeroes ICs if they are not specified in the `initialConditions` field in the JSON.

My question is
1. should AlgebraicJuliaService restrict giving ICs to just symbols in either initial conditions (`ic_specs`) or specified to be plotted, or should every variable be given ICs?
2. Is giving a default IC always valid? For instance if we give Zeroes, what if the model has a singularity at 0?
3. Are there ways of extracting the trajectories of variables where ICs are not provided? I've checked [ODESolution](https://github.com/SciML/SciMLBase.jl/blob/acabca532aa383477b0967b29e21a1a36ce0e0d8/src/solutions/ode_solutions.jl#L92) source and I'm doubtful.

### DIFF
--- a/packages/algjulia-service/src/decapodes-service/initial_conditions.jl
+++ b/packages/algjulia-service/src/decapodes-service/initial_conditions.jl
@@ -31,20 +31,24 @@ This IC contains the domain and the initial conditions data.
 While these are currently tightly-interlinked with InitialConditionsData, they are formally separated to distinguish between the initial conditions schema and the data it might be parameterized over. 
 =#
 @data InitialConditions begin
+    Zeroes(r::Domain)
     # planar
     GaussianIC(r::Rectangle, ξ::GaussianData)
-    # spherical
+    # spheric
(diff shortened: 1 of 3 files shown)


epatters:
Closing since this is an early draft and work on this feature has stopped. Feel free to re-open or a start a new PR if work resumes.

PR #328, opened 2025-01-15
Distinguish icons for different kinds of help

### LINKED ISSUE #319: Clarify interface for different kinds of help
Screenshot 2025-01-06 at 1 25 49 PM

This favicon means "get help about this theory" when in the top bar but "get help about CatColab" when in the hamburger menu.

hamidahoderinwale:
Closes #319 

To clarify interface for different kinds of help. This also includes a small change to have the syntax be the same for all the menu items.

### DIFF
--- a/packages/frontend/src/page/menubar.tsx
+++ b/packages/frontend/src/page/menubar.tsx
@@ -4,14 +4,13 @@ import { getAuth, signOut } from "firebase/auth";
 import { useAuth, useFirebaseApp } from "solid-firebase";
 import { type JSX, Show, createSignal } from "solid-js";
 
-import { Dialog, IconButton } from "../components";
-import { Login } from "../user";
-
-import CircleHelp from "lucide-solid/icons/circle-help";
+import Info from "lucide-solid/icons/info";
 import LogInIcon from "lucide-solid/icons/log-in";
 import LogOutIcon from "lucide-solid/icons/log-out";
 import MenuIcon from "lucide-solid/icons/menu";
 import SettingsIcon from "lucide-solid/icons/settings";
+import { Dialog, IconButton } from "../components";
+import { Login } from "../user";
 
 import "./menubar.css";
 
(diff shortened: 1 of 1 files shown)


packages/frontend/src/page/menubar.tsx
```diff
@@ -4,14 +4,13 @@ import { getAuth, signOut } from "firebase/auth";
...
-import { Login } from "../user";
-
-import CircleHelp from "lucide-solid/icons/circle-help";
+import Info from "lucide-solid/icons/info";
 import LogInIcon from "lucide-solid/icons/log-in";
 import LogOutIcon from "lucide-solid/icons/log-out";
 import MenuIcon from "lucide-solid/icons/menu";
 import SettingsIcon from "lucide-solid/icons/settings";
+import { Dialog, IconButton } from "../components";
+import { Login } from "../user";
```
epatters: Can you put these last two imports back where they were? I like to keep the icon imports in their own "section" but unfortunately the editor tools that automatically insert imports are too dumb to put things in exactly the right place.

packages/frontend/src/page/menubar.tsx
```diff
@@ -83,8 +82,8 @@ function HelpMenuItem() {
 
     return (
         <MenuItem onSelect={() => navigate("/help")}>
-            <CircleHelp />
-            <MenuItemLabel>Help</MenuItemLabel>
+            <Info />
+            <MenuItemLabel>{"About"}</MenuItemLabel>
```
epatters: The top-level help page does function a bit like an "About" page but it's more than that since it links to other docs. Perhaps call this "About & documentation"?

epatters (changes requested):
Thanks Hamidah!

packages/frontend/src/page/menubar.tsx
hamidahoderinwale: Made the changes for both!

epatters: Thanks! Looks like only one change was made though.

hamidahoderinwale: > Thanks! Looks like only one change was made though.

Sorry about that, I took the second message as two changes and missed the first. The imports are back!

epatters (approved):

PR #329, opened 2025-01-15
Duration parameter for Decapodes simulations

quffaro:
This PR adds another table, "Duration," to capture a timestep that the simulation runs in. I've tested this with [diffusion](http://localhost:5173/analysis/0193b10e-e3a9-7772-8cf2-db157bc796d8) by setting duration to be 5 and 50 respectively, observing that t=50 appears to take an amount of time I'd expect to be 10x longer.

Setting the duration to t=100 gave me a "kernel failed to return data" (probably not exact wording) error. So it might be the case that setting longer durations would break the sim.

### DIFF
--- a/packages/frontend/src/stdlib/analyses/decapodes.tsx
+++ b/packages/frontend/src/stdlib/analyses/decapodes.tsx
@@ -1,5 +1,4 @@
-
-     0.0
+0.0;
 import { For, Match, Show, Switch, createMemo } from "solid-js";
 import { isMatching } from "ts-pattern";
 
@@ -38,7 +37,7 @@ export type DecapodesContent = {
     initialConditions: Record<string, string>;
     plotVariables: Record<string, boolean>;
     scalars: Record<string, number>;
-	duration: number;
+    duration: number;
 };
 
 export function configureDecapodes(options: {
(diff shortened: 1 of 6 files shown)


packages/frontend/src/stdlib/analyses/decapodes.tsx
```diff
@@ -1,3 +1,4 @@
+0.0;
```
epatters: Somehow this snuck in here.

packages/frontend/src/stdlib/analyses/decapodes.tsx
```diff
@@ -343,7 +362,7 @@ const makeSimulationCode = (data: SimulationData) =>
 
     f = simulator(system.geometry.dualmesh, system.generate, DiagonalHodge());
 
-    soln = run_sim(f, system.init, 50.0, ComponentArray(k=0.5,));
+    soln = run_sim(f, system.init, ${data.duration}, ComponentArray(k=0.5,));
```
epatters: This works but having to do more code generation/interpolation, minor though it is, is a code smell.

My suggestion would be add a `duration` parameter to `PodeSystem` and then replace this interpolation by the literal `system.duration`.

epatters (changes requested):
Thanks Matt! Two comments below.

epatters (approved):

PR #332, opened 2025-01-17
Import/export JSON for models and diagrams

### LINKED ISSUE #303: Import and export JSON in frontend
CatColab currently provides no facilities for users to import or export their data. The frontend should allow the user to import and export JSON data for the objects stored by the system, starting with models.

When importing data, the JSON should be validated using JSON schema before the user is allowed to proceed (requires #302). Initially we can just do the validation in the frontend. Eventually it will be better to also validate in the backend, but I consider that a separate issue (#304).

KevinDCarlson:
This PR implements the ability to download a JSON dump of a model or a diagram and to upload it back via either the file browser or copy-paste.

A little skeleton code is present for analyses but there are design questions there and I would guess we should just merge this without worrying about analyses yet.

Closes #303.


packages/frontend/src/components/json_import.tsx
KevinDCarlson, line 1: I've put this in `components` and export in `util` because export doesn't actually build a component but it needs to be used by `model_menu`, `diagram_menu`, and `analysis_menu`. Not sure if that's the right file structure.

packages/frontend/src/diagram/diagram_menu.tsx
```diff
@@ -48,7 +51,13 @@ export function DiagramMenuItems(props: {
         navigate(`/diagram/${newRef}`);
     };
 
-    return (
+    //Can this be less repetitive?
```
KevinDCarlson: i.e. I'm wondering whether we should have an overarching `Menu` type that extends to `DiagramMenu` etc. Seems like a fair amount of duplicated boilerplate otherwise.

packages/frontend/src/diagram/diagram_menu.tsx
```diff
@@ -48,7 +51,13 @@ export function DiagramMenuItems(props: {
...
+    const onDownloadJSON = (diagram: DiagramDocument) => {
+        downloadJson({ data: JSON.stringify(diagram), filename: `${diagram.name}.json` });
```
KevinDCarlson: The straight `stringify` of a diagram *contains the Automerge hash* of the diagram's model. What this means is that this simplest possible implementation has the behavior that if you download a diagram D in model M and then re-upload it, you get a diagram D' in *the same model* M. I think this is plausible behavior but worth thinking over.

packages/frontend/src/components/json_import.tsx
```diff
@@ -0,0 +1,105 @@
...
+import type { DiagramDocument } from "../diagram";
+import type { ModelDocument } from "../model";
+
+interface Props {
+    onImport: (data: ModelDocument | DiagramDocument | AnalysisDocument) => void;
+    validate?: (data: ModelDocument | DiagramDocument | AnalysisDocument) => boolean | string; // Return true if valid, error message if invalid
+}
+
+export const JsonImport: Component<Props> = (props) => {
+    const [error, setError] = createSignal<string | null>(null);
```
KevinDCarlson, line 19: This reactive error isn't currently being used. I'm not sure whether this is the paradigm we want here or, if so, quite how best to make use of it.

KevinDCarlson:
I've added some css for the import menu. I was hoping to make the text box auto-widen after pasting in json but couldn't quite figure out the right way to do this, so it just gets taller at the moment. I suppose it doesn't matter much, people aren't likely to be looking closely at their raw JSON. Possibly the ideal solution would be to have an in-client JSON parser display the pasted JSON nicely formatted? But it seems pretty unimportant.

packages/frontend/src/util/json_export.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// Download as file
+export const downloadJson = (props: ExportProps) => {
+    try {
+        const jsonString = props.data;
+        const blob = new Blob([jsonString], { type: "application/json" });
+        const url = URL.createObjectURL(blob);
+
+        const link = document.createElement("a");
```
KevinDCarlson: Surely there's a less weirdo way to do this.

packages/frontend/src/components/json_import.tsx
KevinDCarlson: This error is now being used! I'm not sure if it's the right design for a component that can have errors generate at several points.

packages/frontend/src/page/menubar.tsx
```diff
@@ -154,3 +165,14 @@ function SettingsMenuItem() {
...
     );
 }
+
+function ImportMenuItem(props: {
+    showImport: () => void;
+}) {
+    return (
+        <MenuItem onSelect={props.showImport}>
+            <UploadIcon />
+            <MenuItemLabel>{"Import notebook"}</MenuItemLabel>
```
KevinDCarlson, line 174: I put this at the very top because I think it should be next to the "new" buttons which are handled by document type and otherwise I'd have to split up the child.props below.

KevinDCarlson:
@epatters Ready for your consideration. A couple more questions that remain:

- Should there be a loading state displayed during import? It can take a little bit.
- Can we handle menus overall in a smarter way? There's a fair bit of duplicative work and menus getting built across many files.

packages/frontend/src/components/json_import.tsx
```diff
@@ -0,0 +1,109 @@
...
+            {/* JSON paste */}
+            <div class="flex">
+                <label>Or paste JSON:</label>
+                <textarea
+                    value={pasteValue()}
+                    onInput={handleInput}
+                    onPaste={handleInput}
+                    placeholder="Paste your JSON here..."
+                />
+                <button onClick={handlePaste} aria-label="Import JSON">
```
jmoggr: `handlePaste` is likely not the right name, given that it's actually handling a click and there is an onPaste right above it. It took me a moment to recognize that this was not the `onPaste` handler.

I might suggest `handleClick`

packages/frontend/src/components/json_import.tsx
```diff
@@ -0,0 +1,109 @@
...
+import "./json_import.css";
+
+interface JsonImportProps<T extends string> {
+    onImport: (data: Document<T>) => void;
+    validate?: (data: Document<T>) => boolean | string;
+}
+
+export const JsonImport = <T extends string>(props: JsonImportProps<T>) => {
+    const [error, setError] = createSignal<string | null>(null);
+    const [pasteValue, setPasteValue] = createSignal("");
```
jmoggr: nit: textAreaValue (or similar) is likely a more appropriate name. This is not necessarily the value directly from the users clipboard.

packages/frontend/src/components/json_import.tsx
```diff
@@ -0,0 +1,109 @@
...
+            }
+
+            const text = await file?.text();
+            validateAndImport(text);
+
+            // Reset file input
+            input.value = "";
+        } catch (e) {
+            handleError(e);
+        }
```
jmoggr: this try/catch block is superfluous, nothing in it can throw unexpectedly if you change to the following pattern:

```
           if (!(file?.type === "application/json") && !file?.name.endsWith(".json")) {
                handleError(new Error("Please upload a JSON file"));
                return;
            }
```

and `validateImport` won't throw either since it is itself a try/catch block.

packages/frontend/src/components/json_import.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+        try {
+            const file = input.files[0];
+
+            // Validate file type
+            if (!(file?.type === "application/json") && !file?.name.endsWith(".json")) {
+                throw new Error("Please upload a JSON file");
+            }
+
+            const MAX_FILE_SIZE = 5 * 1024 * 1024; // 5MB
```
jmoggr: note: you are absolutely going to want to enforce this on the backend as well, and for that it would be best to use the same constant. but that is a later problem

packages/frontend/src/components/json_import.tsx
```diff
@@ -0,0 +1,109 @@
...
+    // Handle file upload
+    const handleFileUpload = async (event: Event) => {
+        const input = event.target as HTMLInputElement;
+        if (!input.files?.length) return;
+
+        try {
+            const file = input.files[0];
+
+            // Validate file type
+            if (!(file?.type === "application/json") && !file?.name.endsWith(".json")) {
```
jmoggr: it's best to simplify where possible  `file?.type !== "application/json" && !file?.name.endsWith(".json")`

put that PhD to use! /s :)

packages/frontend/src/components/json_import.tsx
jmoggr: I think it's fine. It's worth noting that `json_exports` could be a component, more on that below.

jmoggr (commented):
At a high level this looks good! There are a number of nits, I'll open a PR with fixes for all of them.

I'm trying to gauge my pickiness off of the cleanliness of the rest of the codebase (very clean!), let me know if you'd like to apply a different amount of rigor.

packages/frontend/src/components/json_import.tsx
jmoggr: I think that this is correct. For a user facing workflow that's likely to have error it's best to not take the user out of the workflow with a popup, redirect, etc...

packages/frontend/src/util/json_export.tsx
jmoggr: Not yet unfortunately. We're waiting on the new File System API to land in all browsers, specifically [showSaveFilePicker](https://developer.mozilla.org/en-US/docs/Web/API/Window/showSaveFilePicker)

It is worth noting that some people have security setting such that .click() is not allowed. Nothing will happen for those people. The setting is super rare.

There are other ways of doing this, but on review I think this is still likely the best option.

packages/frontend/src/diagram/diagram_menu.tsx
jmoggr: Edit: I took a shot at this, it was less scary than I initially thought. PR is [here](https://github.com/ToposInstitute/CatColab/pull/407)

### old:
Based off of my limited understanding, I feel like we should hold off on merging them.

- The menus are different enough that the abstraction currently would not be much more than an if statement for each menu item, which I don't think is worth the additional complexity.
       - We could fix that by defining generic interfaces that work for both models and diagrams (like what you've done for the import), but that's a non-zero amount of work and...
- It's not obvious how much similarity there will be between the two menus going forward, and if they keep diverging an abstraction would become increasingly costly to maintain.

An alternative might be to create reusable components for menu items which can be easily shared. In this case it would be something like <ExportOrCopyMenuItem document={props.liveDiagram.liveDoc.doc} /> which add the 2 menu items and there handler.

Composition over inheritance and premature optimization is generally bad

packages/frontend/src/components/json_import.tsx
KevinDCarlson: I added an issue for this. #408

packages/frontend/src/components/json_import.tsx
```diff
@@ -0,0 +1,106 @@
+import { createSignal } from "solid-js";
+import type { Document } from "../api";
+import "./json_import.css";
+
+interface JsonImportProps<T extends string> {
+    onImport: (data: Document<T>) => void;
+    validate?: (data: Document<T>) => boolean | string;
+}
+
+export const JsonImport = <T extends string>({ onImport, validate }: JsonImportProps<T>) => {
```
epatters, line 18: I think having a doc-comment on the top-level component/primary export is a nice touch. It doesn't have to be long.

packages/frontend/src/components/json_import.tsx
```diff
@@ -0,0 +1,106 @@
...
+                    setError(validationResult);
+                    return;
+                }
+            }
+
+            // Clear any previous errors and import
+            setError(null);
+            onImport(data);
+            setImportValue(""); // Clear paste area after successful import
+        } catch (e) {
```
epatters, line 43: Following up on #398, should we use a local `ErrorBoundary` here instead of having to manually wrap everything in a try-catch?

packages/frontend/src/components/json_import.tsx
```diff
@@ -0,0 +1,106 @@
...
+            handleError(e);
+        }
+    };
+
+    // Handle file upload
+    const handleFileUpload = async (event: Event) => {
+        const input = event.target as HTMLInputElement;
```
epatters: You could save yourself the explicit type cast of `event.target` by typing `handleFileUpload` as follows: https://docs.solidjs.com/configuration/typescript#basics

packages/frontend/src/components/json_import.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const handleTextareaSubmit = () => {
+        if (!importValue().trim()) {
+            setError("Please enter some JSON");
+            return;
+        }
+        validateAndImport(importValue());
+    };
+
+    const handleInput = (event: Event) => {
+        const textarea = event.target as HTMLTextAreaElement;
```
epatters: Ditto about types here.

epatters (changes requested):
Thanks Kevin and Jason, this is looking great! I left a few small comments. I will merge once those are addressed.

packages/frontend/src/components/json_import.tsx
KevinDCarlson: It's my understanding that `ErrorBoundary` will only catch rendering errors, not errors thrown by event handlers like `validateAndImport`. So I'm leaving this in place for now unless somebody can show I'm confused/some other better way.

KevinDCarlson: It seems like it'd be reasonable to also add an ErrorBoundary on the component, but OTOH there's not much being rendered here, so I'll also leave it out unless someone has a strong opinion otherwise.

epatters: I see. That seems to be confirmed by this comment: https://github.com/solidjs/solid/discussions/1174#discussioncomment-3469578

Thanks for clarifying!

KevinDCarlson:
@epatters ready for you again

epatters (approved):
Thanks Kevin! I'm going to squash down some of these commits and then merge.

packages/frontend/src/components/json_import.tsx
jmoggr: Kevin is correct about `ErrorBoundary`. It's worth noting that async callbacks in components like `createResource` count as 'rendering', which is nice for us.

Another reason for the try catch is so that we can provide precise errors to the user without taking them out of their workflow. I think this is the best approach for flows where user errors are highly likely.

I'm slightly for adding an error boundary here. It's done defensively and because the component can fail independently of the rest of the app. I've updated this branch with an error boundary.

Issue #112, opened 2024-08-18
Mass action-style ODE semantics for stock-flow diagrams


epatters:
Inspired by the mass action kinetics for Petri nets, this semantics for primitive stock and flow diagrams (not described in the literature AFAIK) would assign flow functions equal to a constant times the upstream stock times the product of all stocks linked to the flow.

Not what the system dynamics practitioners will want but good enough for a proof of concept and enough to produce SIR-type dynamics.

PR #336, opened 2025-01-22
Remove `mathexpr` module for now

epatters:
Some house cleaning in preparation for #112. I am removing the `mathexpr` module in the core because

1) we are not currently using it, and the planned #112 can use a polynomial vector field based on a simple data structure for polynomials, thus avoiding the need to generate and then parse arbitrary mathematical expressions

2) while we may well want something like this in the future, @olynch has said that `mathexpr` is obsoleted by his interesting new work on "F-expressions," inspired by the idea of [bicameral syntax](https://parentheticallyspeaking.org/articles/bicameral-not-homoiconic/)


olynch:
Sounds good to me: also it definitely makes sense to generate polynomials directly instead of expressions. When we again need to parse arbitrary math expressions, we can compile them to polynomials.

PR #346, opened 2025-01-27
Links for embedding models elsewhere

hamidahoderinwale:
When button is selected (CodeXml icon) users can copy iframe code with settings preset.

In the future, maybe users could change style settings and select specific parts of their model (via notebook cells).


epatters (changes requested):
Thanks for starting this PR! Things to do:

- Lots of unrelated stuff has been accidentally committed
- Rather than having a separate embed button and dialog, a button to copy the link should be adding to the sharing/permissions dialogs where appropriate. See, for instance, how Google Docs does it, or ask me if you have questions.

Issue #318, opened 2025-01-06
Systematize types for documents and links between them


epatters:
We now have several different kinds of documents, such as model documents, diagram documents, and analysis documents. All of these documents are JSON-able and stored as document `refs` in the database. Moreover, there can be links/references/foreign keys between them. For example, a diagram references the model that it is a diagram in. These links need not be at the top-level of the JSON. Soon models will be able to import other models using special cells in the notebook.

In preparation for the v0.2 release, I would like to better systematize the notions of document and link in the hopes of avoiding breaking changes or ad hoc workarounds later.

In the original Node backend, @olynch had a notion of `extern` references in documents. I removed it when I rewrote the backend in Rust (#211) because we weren't using it for anything yet and so I wasn't sure whether/how I should re-implement it. However, it persists in the frontend type defs as [`ExternRef`](https://github.com/ToposInstitute/CatColab/blob/fadef8f979a139ad1ab8688b07147051dc5e3ecd/packages/frontend/src/api/types.ts#L12). I'd like to bring something like this back.

## Working design

Certain special JSON keys, prefixed by the `@` symbol, will be recognized by the backend and hoisted into a typed graph (actually, I would rather think of it as a model of some double theory) in the SQL DB. The objects will be documents:

```typescript
interface Document<T extends string> {
  "@id": UUID;
  "@type": T;
}
```

Before a document is stored in a JSONB column, these special keys will be stripped and stored in their own columns.

The arrows will be links:

```typescript
interface Link<T extends string> {
  "@repo": string;
  "@id": UUID;
  "@type": T;
}
```

The database will have a new `links` table into which the special keys are hoisted. Note that `@type` here is the type of the link, not of the target document.

For example, the document type for a diagram will look like:

```typescript
interface DiagramDocument extends Document<"diagram> {
  name: string;
  inModel: Link<"in_model">;
  notebook: Notebook<DiagramJudgment>;
}
```

with a typical instance:

```typescript
const diagram: DiagramDocument = {
  "@id": [...],
  "@type": diagram;
  name: "Evan's diagram";
  inModel: {
    "@repo": "next.catcolab.org",
    "@id": [...],
    "@type": "inModel",
  },
  notebook: [...]
};
```

## Prior art

People have obviously thought about [linking JSON](https://www.mnot.net/blog/2011/11/25/linking_in_json) before. In fact, I was rather annoyed that this wasn't properly standardized a decade ago. The best known candidates are:

[...]

Issue #291, opened 2024-12-11
Published flag for documents


epatters:
In addition to its permissions/access control list, a document should have a `published` flag stating whether it is publicly listed. Once a document is published, it cannot be unpublished, and a published document can depend only on other published documents.

This requires:

- adding a boolean column `published` to the document `ref` table
- enforcing the constraints above in the backend
- adding a toggle for publishing to the UI for permissions (#276)

For further discussion, see Sec 3.1.2 of the [n-dim report](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=e538fffc635c40315279ba59c323f3297edd8722).

KevinDCarlson:
I find the semantic distinction between published versus viewable-by-all to be rather subtle. Presumably any published document must be at least viewable-by-all, so these can't quite be independent columns. Should we just force permissions up to at least "all viewable" if the user checks published?

epatters:
I agree that it's subtle. I think the distinction is that a published document broadcasts its existence (e.g., it shows up in public lists of a user's models or in searches of the public library of models), whereas as a non-published document, even if readable by anyone, can only be found if you have the link.

Here's what the n-dim report (Sec 3.1.2) says about publishing:

> When an object is first created, it is malleable, but only in so far as its owner is concerned. Until it is *published*, no one but its owner can even know of its existence. The act of publishing an object is similar to the notion of publishing a paper; once the paper is published, its author cannot go to all of the libraries in the world and remove it from circulation. In the same way, publishing an object makes it visible to the rest of the world and, at the same time, immutable, even to its original author. If one wants to change a published object, it must be copied, and the copy revised, at which time the pedigree-maintenance mechanisms described above are activated. In this way, the path any published object has taken through its many copies can always be traced.
>
> To publish an object, everything that it depends on must be published, namely, all targets of part links with the object as source. One is thus guaranteed that when a published object is manipulated, even some time (e.g. years) after it was published, it will behave exactly as it did *at the time of publication*. This is a critical property of published
objects, and one reason for the way in which it has been formulated in n-dim.
>
> ...
>
> Published objects can never be destroyed, although n-dim does have a notion of archival vs. active storage, in the same way that libraries move little-used books and material to less easily accessed but more efficient storage, such as microfiche.

[...]

PR #347, opened 2025-01-27
Print files owned by current user to console

KevinDCarlson:
This PR adds a route `/user/:username` accessing a public profile page for user `username` which lists their documents. Currently it simply lists all owned models but the intended design is that it lists all public documents of any class. The display is currently just of hyperlinked titles; the intent is to include more metadata.


packages/frontend/src/user/profile.tsx
```diff
@@ -8,37 +8,98 @@ import { FormGroup, TextInputItem } from "../components";
...
     return (
         <div class="growable-container">
             <BrandedToolbar />
             <div class="page-container">
                 <LoginGate>
                     <h2>Public profile</h2>
-                    <UserProfileForm />
+                    <UserProfileForm onRefsLoaded={setMyRefs} />
+                    <div style="margin-top: 1rem;">
+                        <h3
```
epatters: No inline styles. Use the external stylesheet.

packages/frontend/src/user/profile.tsx
```diff
@@ -8,37 +8,98 @@ import { FormGroup, TextInputItem } from "../components";
...
+                                            <Fragment key={id}>
+                                                <a
+                                                    href={`${import.meta.env.VITE_SERVER_URL}/model/${id}`}
+                                                    class="filebutton"
+                                                    onMouseOver={(e) => {
+                                                        e.currentTarget.style.boxShadow =
+                                                            "0 2px 8px rgba(0,0,0,0.1)";
+                                                    }}
+                                                    onMouseOut={(e) => {
+                                                        e.currentTarget.style.boxShadow = "none";
```
epatters: Ditto about inline styles, and also you shouldn't use mouse over events to change styles. Use the appropriate CSS selectors.

packages/frontend/src/user/profile.tsx
```diff
@@ -8,37 +8,98 @@ import { FormGroup, TextInputItem } from "../components";
...
+                            Your Documents
+                        </h3>
+                        <Show when={myRefs()} fallback={<div> Loading user files... </div>} keyed>
+                            {(items) => {
+                                return (
+                                    <div class="files">
+                                        {items.map(([id, title]: [string, string]) => (
+                                            <Fragment key={id}>
+                                                <a
+                                                    href={`${import.meta.env.VITE_SERVER_URL}/model/${id}`}
```
epatters: Shouldn't need this env variable. Use Solid Router's `A` component.

packages/frontend/src/user/profile.tsx
```diff
@@ -8,37 +8,98 @@ import { FormGroup, TextInputItem } from "../components";
 import { BrandedToolbar } from "../page";
 import { LoginGate } from "./login";
 
+import "./profile.css";
+
 /** Page to configure user profile. */
 export default function UserProfilePage() {
+    const [myRefs, setMyRefs] = createSignal<[string, string][] | null>(null);
```
epatters: Having to pass these refs around is a code smell. Also, I don't understand the types. This isn't how you usually create a ref.

epatters (changes requested):
Thanks for starting this, Kevin. I left a few quick comments, but there are bigger designs issue that we need to address:

1. We shouldn't list just any documents that the user's owns: what if they're supposed to be private? We should implement #291 first and then only list documents that have been marked as published. Sub has emphasized this and n-dim had a similar approach, which I think we should follow.

2. The user profile editable by a logged-in user should be a separate page from the public display of a profile visible to anyone. This will also avoid that weird stuff with refs.

KevinDCarlson:
Thanks @epatters. I agree that there should be a public user page that shows only public docs. Maybe the page where you can edit your profile can be a "private" profile page where you can also see your private owned docs?

packages/frontend/src/user/profile.tsx
KevinDCarlson: The refs are going to be a list of `[string,string]`s giving a document's ref and name. Is the problem that it's already been stringified rather than passing around actual document objects, or something like that?

packages/frontend/src/user/profile.tsx
KevinDCarlson: Should/can we lint directly for the keyword `style`, then, if we're not allowing even one-liners?

packages/backend/pkg/src/index.ts
KevinDCarlson, line 1: I think `RpcResult<Array<[string, string]>>>` is probably wrong here and reflects the weird front-end types you couldn't parse earlier.

packages/frontend/src/user/profile_public.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { For, Show, createEffect, createSignal } from "solid-js";
+import { useApi } from "../api";
+import "./profile_public.css";
+import invariant from "tiny-invariant";
+
+export default function PublicProfilePage() {
+    const api = useApi();
+    const params = useParams();
+
+    const [userModels, setUserModels] = createSignal<[string, string][]>(); //QQQ: unclear what type the RPC call should return to; Documents?
```
KevinDCarlson, line 11: See first comment for the Rust side of this question.

KevinDCarlson:
OK, I think I was able to sensibilize the work so far pretty easily. But there are some todos:

- [ ] Should the model list display more than just the title? Probably the theory, maybe the most recently edited time, and it should probably be sorted by the latter (later, options for how to sort, etc)
- [ ] Need to let people choose which docs are public as in #291 
- [ ] After deciding what the RPC function should really be returning rather than just the stringified UUID and name (the whole content JSON blob?), that needs to be reimplemented.
- [ ] Partly because of the insufficient design above, this currently only lists models. The page should list the document type. Conceivably these points depend on #318 , to convert the JSON blobs into general `Document`s as they come into the front-end?
- [ ] Probably future work but the document list should probably be a standalone component which can be configured with permissions settings and then display public documents on the public profile page and all documents on the private profile page. 
- [ ] There should be a link to your own public profile when you're logged in.

epatters:
Replaced by #465.

PR #352, opened 2025-01-29
Warning for incomplete signup form submission

hamidahoderinwale:
Users get an alert in the signup form if they try to signup/login while leaving the fields blank.

### DIFF
--- a/packages/frontend/src/user/login.tsx
+++ b/packages/frontend/src/user/login.tsx
@@ -81,24 +81,32 @@ export function Login(props: {
             <Form onSubmit={onSubmit}>
                 <Field name="email">
                     {(field, props) => (
-                        <input
-                            {...props}
-                            type="email"
-                            required
-                            value={field.value}
-                            placeholder="Email address"
-                        />
+                        <>
+                            <input
+                                {...props}
+                                type="email"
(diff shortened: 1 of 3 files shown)


epatters (changes requested):
Hi Hamidah, thanks for identifying this problem and starting on a fix for it!

To implement this, we should use the validation API that comes with Modular Forms rather than manually hooking up a signal. Specifically, see this page, which has an example very similar to this situation which you should be able to adapt: https://modularforms.dev/solid/guides/validate-your-fields

packages/frontend/src/user/login.tsx
```diff
@@ -30,6 +30,10 @@ type EmailAndPassword = {
 export function Login(props: {
     onComplete?: (user: User) => void;
 }) {
+    const [userAdded, setUserAdded] = createSignal<
+        "Please complete login" | "Please complete signup" | ""
```
epatters: This should become irrelevant in my proposed approach, but for the sake of learning, I should point out that there is hardly ever a situation where you want to elevate human-readable text like this into the type level. This requires that the message be repeated exactly, which is itself generally a code smell.

epatters (approved):

PR #353, opened 2025-01-29
Downsampling Decapodes simulations on large meshes

quffaro:
This PR allows us to downsample the mesh a simulation is running on so we avoid hitting browser string limitations.


quffaro:
image

So far we can run [Navier-Stokes inviscid](http://localhost:5173/analysis/01938869-4114-73c0-b412-42d86d941296) in Chromium on Ico7@t=10. I'd like to use the browser information to set up a constrained linear problem between timesteps, mesh size, downsampling radius (the radius form the center of the mesh that downsampling takes place), and rate of downsampling (the distance between downsampling)

epatters:
We're abandoning this approach because it's hard to get right. Instead, we will produce the videos in Julia (like we've been doing) and send those over the wire.

PR #389, opened 2025-02-05
Help pages for all theories

### LINKED ISSUE #86: Theory-specific documentation
There aren't many instructions; probably we should at least have something like a sidebar reminder of what keyboard commands are available and, dependent on the theory, of what kinds of cells fit into what other kinds of cells.

### LINKED ISSUE #296: Stable list of example models
For ease of demoing and later for other purposes, we should have a library of example models. These could be linked from the theory-specific documentation pages (#86) or collected somewhere else.

KevinDCarlson:
As described, with some new examples and references to go with the Wren release.

Closes #86 and, for present purposes, also closes #296.

### DIFF
--- a/packages/frontend/README.md
+++ b/packages/frontend/README.md
@@ -9,7 +9,7 @@ Install Rust and [pnpm](https://pnpm.io/), then run
 
 ```sh
 cd packages/frontend
-pnpm run install
+pnpm install
 pnpm run build
 ```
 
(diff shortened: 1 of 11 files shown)


epatters (approved):
Thanks Kevin! I made a few small changes and additions, and I also fixed a bug I noticed while moving between the docs pages.

Issue #375, opened 2025-02-04
Migrate models along monomorphisms of theories


epatters:
As a sequel to #374, it should be possible to push forward a model along a monomorphism of theories. Abstractly, this is the same thing as #374 but in practice it requires more work since (1) the theory map must be given explicitly instead of implictly and (2) it can't be accomplished in the model data structure by a pointer change but requires changing object and morphism types.

epatters:
General pushforward migration (for models of discrete theories) was implemented in #660.

PR #391, opened 2025-02-07
Trivial migrations in the frontend and core

kris-brown:
This PR addresses https://github.com/ToposInstitute/CatColab/issues/375 

In the core, there is a method `pushforward` which takes a `DblModelMorphism` and uses it to act on a model (this may in general not produce a valid model without something like the chase, but right now we're just concerned with cases where `pushforward` does produce a valid model). 

In the frontend, there is an attribute `inclusions` for each `Theory` which is a `Map` keyed by the ids of possible codomain theories, with the value being a pair of renamings for the generating objects and morphisms. These are hardcoded in `theories.tsx`. 

For any such monomorphism we could also pushforward in the *opposite* direction so long as we have no formal cells which are outside the image of the monomorphism. This has the effect of performing a $$\Delta$$ migration. The app forbids you from doing such a reverse migration if it would involve loss of data.

## Room for improvement

These things could be addressed in this PR but I'm also open to them being future PRs as needed.

- Theories which could be "reverse sigma migrated" to in principle (if only certain cells were deleted) could be visible but greyed out in the theory selection menu.

- Right now, the theory "group" (e.g. "Knowledge and Data") is still present even if there are no available theories within that group. Unsure if that is a problem.

- Changing the theory is in general an irreversible and mutating operation. Some sugar could be added in the future to combine the theory change with a duplication of the model.

- There are places where for tabulator theories I resort to `panic` but in principle should be doable. We don't _yet_ have cases where we want to migrate a tabulator theory. 

- One can pushforward from `CausalLoop` into `CausalLoopWithDelays` or `CasualLoopIndeterminates`. When one does that (but hasn't yet added anything else), one is able to pushforward *back* into `CausalLoop`, but one cannot go from `CausalLoopWithDelays` directly to  `CausalLoopIndeterminates` (or vice versa). If one wanted to do that, you'd have to first switch back to `CasualLoop`. This is because there are no simple privileged migrations between those theories. It's not clear this is a problem, though it does feel a little weird to have to switch back to `CausalLoop`.

### DIFF
--- a/packages/catlog-wasm/src/model.rs
+++ b/packages/catlog-wasm/src/model.rs
@@ -262,30 +262,40 @@ impl MapData {
             codhom: codhom.iter().map(|v| MorType::Basic(ustr(v))).collect(),
         }
     }
+
     #[wasm_bindgen(getter)]
     pub fn domob(&self) -> Vec<ObType> {
         self.domob.clone()
     }
+
     #[wasm_bindgen(getter)]
     pub fn codob(&self) -> Vec<ObType> {
         self.codob.clone()
     }
+
     #[wasm_bindgen(getter)]
     pub fn domhom(&self) -> Vec<MorType> {
         self.domhom.clone()
     }
+
     #[wasm_bindgen(getter)]
     pub fn codhom(&self) -> Vec<MorType> {
         self.codhom.clone()
     }
+
     #[wasm_bindgen]
     pub fn includes_ob(&self, ob: &ObType) -> bool {
         self.codob.contains(ob)
     }
+
     #[wasm_bindgen]
     pub fn includes_mor(&self, mor: &MorType) -> bool {
-        self.codhom.contains(mor)
+        match mor {
+            MorType::Basic(_) => self.codhom.contains(mor),
+            MorType::Hom(x) => self.codob.contains(x),
+        }
     }
+
     #[wasm_bindgen]
     pub fn swap(&self) -> MapData {
         MapData {
--- a/packages/catlog-wasm/src/model.rs
+++ b/packages/catlog-wasm/src/model.rs
@@ -16,8 +16,8 @@ use catlog::dbl::functor::{DblFunctor, DiscreteDblTheoryMorphism, FgDiscreteDblT
 use catlog::dbl::model::{
     self as dbl_model, DblModel as CoreDblModel, FgDblModel, InvalidDblModel, MutDblModel,
 };
-use catlog::dbl::theory::{DiscreteDblTheory, TabMorType, TabObType};
-use catlog::one::fin_category::{FinCategory, FinMor, UstrFinCategory};
+use catlog::dbl::theory::DiscreteDblTheory;
+use catlog::one::fin_category::{FinCategory, UstrFinCategory};
 use catlog::one::{Category as _, FgCategory, Path};
 use catlog::validate::Validate;
 use catlog::zero::MutMapping;
@@ -237,6 +237,66 @@ pub(crate) type DiscreteDblModel = dbl_model::DiscreteDblModel<Uuid, UstrFinCate
 pub(crate) type DiscreteTabModel =
     dbl_model::DiscreteTabModel<Uuid, Ustr, BuildHasherDefault<IdentityHasher>>;
 
+#[wasm_bindgen]
+#[derive(Clone, PartialEq, Eq, Hash, Debug, Serialize, Deserialize)]
+pub struct MapData {
+    domob: Vec<ObType>,
+    codob: Vec<ObType>,
+    domhom: Vec<MorType>,
+    codhom: Vec<MorType>,
+}
+
+#[wasm_bindgen]
+impl MapData {
+    #[wasm_bindgen(constructor)]
+    pub fn new(
+        domob: Vec<String>,
+        codob: Vec<String>,
+        domhom: Vec<String>,
+        codhom: Vec<String>,
+    ) -> MapData {
+        MapData {
+            domob: domob.iter().map(|v| ObType::Basic(ustr(v))).collect(),
+            codob: codob.iter().map(|v| ObType::Basic(ustr(v))).collect(),
+            domhom: domhom.iter().map(|v| MorType::Basic(ustr(v))).collect(),
+            codhom: codhom.iter().map(|v| MorType::Basic(ustr(v))).collect(),
+        }
+    }
+    #[wasm_bindgen(getter)]
+    pub fn domob(&self) -> Vec<ObType> {
+        self.domob.clone()
+    }
+    #[wasm_bindgen(getter)]
+    pub fn codob(&self) -> Vec<ObType> {
+        self.codob.clone()
+    }
+    #[wasm_bindgen(getter)]
+    pub fn domhom(&self) -> Vec<MorType> {
+        self.domhom.clone()
+    }
+    #[wasm_bindgen(getter)]
+    pub fn codhom(&self) -> Vec<MorType> {
+        self.codhom.clone()
+    }
+    #[wasm_bindgen]
+    pub fn includes_ob(&self, ob: ObType) -> bool {
+        self.codob.contains(&ob)
+    }
+    #[wasm_bindgen]
+    pub fn includes_mor(&self, mor: MorType) -> bool {
+        self.codhom.contains(&mor)
+    }
+    #[wasm_bindgen]
+    pub fn swap(&self) -> MapData {
+        MapData {
+            domob: self.codob(),
+            codob: self.domob(),
+            domhom: self.codhom(),
+            codhom: self.domhom(),
+        }
+    }
+}
+
 /** A box containing a model of a double theory of any kind.
 
 See [`DblTheoryBox`] for motivation.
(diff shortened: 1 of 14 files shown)


epatters (commented):
Thanks Kris! I was going to make a few line-level comments, but then I realized there is a design issue that we haven't fully discussed but I think should be addressed before proceeding further.

Currently, the situation is that you can't push forward models except in the *most* trivial case, namely from the initial theory to another theory. However, provided you don't have any formal cells in the notebook, you're free to keep the changing the theory, even after choosing a non-initial theory. A mathematical gloss on this intuitively correct behavior is that as long as you have no formal content, starting from any theory, you can pull back to the initial theory then pushforward to any other theory without loss of data.

With this PR, once you change to a theory, you can never undo that. You can only keep changing to bigger theories. As you say in your description, "changing the theory is in general an irreversible and mutating operation." This is quite punishing to the user: if they make a wrong click, they can never get back to the previous state. The aim, I think, should be make these migrations as fluid as possible.

I can imagine some possible strategies to address this problem and I'm sure you can too. Let's discuss more in person.

kris-brown:
This PR _does_ recover the old behavior (as long as there are only informal cells, then you can switch between any theories reversibly). It's only when you have formal content that changing theory is irreversible. Still, I agree that if one is only using formal cells from a simpler theory that it should be possible to move from more complex to the simpler theory, so happy to talk about ways of doing that!

epatters:
Oh good, I missed that. Thanks for clarifying. But yes, still worth talking more.

KevinDCarlson:
I suppose we should just ask for both deltas and sigmas along inclusions, yes? That’ll make the menu design a bit trickier as you need both ancestors and descendants available

kris-brown:
Just to update on progress, I've got a working `pushforward` in the core, and I'm very close to getting     

```rust
#[wasm_bindgen]
pub fn pushforward
```

working, though I'm stuck at an error 

```
mismatched types
expected struct `catlog::dbl::model::DiscreteDblModel<Uuid, _>`
   found struct `catlog::dbl::model::DiscreteDblModel<Ustr, _>`
```

when trying to call that core method. Will continue working on it!

kris-brown:
Ok! This is now working as intended, so it's ready for others to look at it! Check out in the preview URL: you can add a type+aspect to an olog and then switch to a ThSchema model. You can switch back iff you don't add any attrtypes/attrs. Likewise with causal loop diagrams being migrated into ones with delays (you can switch back iff you don't actually add any delays, or if you've deleted the ones you added).  

I'm unsure how to pass the rust lint check. If I don't import `MutMapping`, I get an error 

```
no method named `set` found for struct `column::HashColumn` in the current scope
```

But if I include the import then the linter warns `unused import: MutMapping`

packages/catlog-wasm/src/model.rs
```diff
@@ -308,6 +360,46 @@ impl DblModel {
         })
     }
 
+    /// What is the type of this object? Expects `has_ob(ob) == true`
+    #[wasm_bindgen(js_name = "obType")]
+    pub fn ob_type(&self, ob: Ob) -> Result<ObType, String> {
+        match &self.0 {
```
epatters: The best way I've found to handle this situation is to use the `all_the_same` macro. If you search around the crate you'll find examples. In this case, the method body should be something like this (*warning*: not tested):

```rust
all_the_same!(match &self.0 {
      DblModelBox::[Discrete, DiscreteTab](model) => {
          let ob = ob.try_into()?;
          Ok(model.ob_type(&ob).into())
      }
  })
```

packages/catlog-wasm/src/model.rs
```diff
@@ -308,6 +360,46 @@ impl DblModel {
...
+                    TabObType::Tabulator(_) => panic!("Not yet treated"),
+                })
+            }
+        }
+    }
+
+    /// What is the type of this object? Expects `has_ob(ob) == true`
+    #[wasm_bindgen(js_name = "morType")]
+    pub fn mor_type(&self, mor: Mor) -> Result<MorType, String> {
+        match &self.0 {
```
epatters: Ditto about `all_the_same` here: it should be possible to simplify this code and eliminate most of the boilerplate.

packages/catlog-wasm/src/model.rs
```diff
@@ -257,6 +263,52 @@ impl DblModel {
...
+    fn theory_arc(
+        &self,
+    ) -> Arc<DiscreteDblTheory<FinCategory<Ustr, Ustr, BuildHasherDefault<IdentityHasher>>>> {
+        match &self.0 {
+            DblModelBox::Discrete(dblmod) => dblmod.theory_arc(),
+            DblModelBox::DiscreteTab(_) => panic!("Tab theories not yet supported"), // dblmod.theory_arc() // HAS A DIFFERENT RETURN TYPE
+        }
+    }
+
+    /// Find negative feedback loops in a model.
```
epatters: Wrong docstring.

packages/catlog-wasm/src/model.rs
```diff
@@ -257,6 +263,52 @@ impl DblModel {
...
+    #[wasm_bindgen]
+    pub fn pushforward(
+        &self,
+        codtheory: &DblTheory,
+        domob: Vec<String>,
```
epatters: Why have these be `Vec<String>` instead of `Vec<ObType>` and `Vec<MorType>`? Even if you're specializing to discrete double theories for now in the body, seems better to handle the type coercion safely in the Rust code instead of doing untyped stuff with strings in TypeScript.

You might even be able to write the all body generically using `all_the_same` as sketched in a simpler case below, but that can wait for another time if it seems hard.

packages/frontend/src/model/document.ts
```diff
@@ -19,6 +21,44 @@ export type ModelDocument = Document<"model"> & {
...
+UUIDs in `updated_model` not found in the current document)
+*/
+export function updateFromCatlogModel(
+    changeDoc: (f: ChangeFn<ModelDocument>) => void,
+    updated_model: DblModel,
+): void {
+    const object_types = new Map<string, string | MorType>();
+    const mor_types = new Map<string, string | ObType>();
+    for (const ob of updated_model.objects()) {
+        object_types.set(ob.content.toString().substring(0, 16), updated_model.obType(ob).content);
```
epatters: I am alarmed that you're having to call `content.toString()` and then extract a substring here and below.

I assume that this is to work around the sad fact that JavaScript Maps can only have strings as keys. A few comments:

- A heuristic I try to follow is that any algorithm requiring a hash map with non-string keys is too elaborate for JavaScript and should be in the Rust code
- However, I haven't always followed my own advice: if you look in `catlog-wasm/src/theory.rs` you will find helper types `ObTypeIndex` and `MorTypeIndex` that wrap `HashMap<ObType, usize>` and `HashMap<MorType, usize>`. Using these types is another strategy to clean up this code.

packages/frontend/src/model/model_editor.tsx
```diff
@@ -76,6 +77,19 @@ export function ModelPane(props: {
...
+            .doc.notebook.cells.filter((cell) => cell.tag === "formal")
+            .map((cell) => {
+                if (cell.tag === "formal" && cell.content.tag === "object") {
+                    return cell.content.obType.content.toString();
+                }
+                if (cell.tag === "formal" && cell.content.tag === "morphism") {
+                    const j = JSON.parse(JSON.stringify(cell.content.morType.content.valueOf()));
+                    return typeof j === "string" ? j : j.content.toString();
+                }
+                return "";
```
epatters: Echoing my comment above, this coercion to strings and to/from JSON is a code smell. I'm hoping that we can eliminate all of this.

packages/frontend/src/model/theory_selector.tsx
```diff
@@ -48,22 +47,69 @@ export function TheorySelector(props: TheorySelectorProps) {
...
-
     return (
         <div class="theory-selector">
             <For each={groupedTheories()}>
                 {([group, theories]) => (
                     <div class="group">
                         <div class="group-name">{group}</div>
-                        <For each={theories}>
+                        <For
+                            each={theories.filter((t) => {
```
epatters: As a matter of style, I think it's more readable to define long helper functions like this before the JSX rather than inline, which due to the depth tends to force the formatter to chop up your code in unfortunate ways.

packages/frontend/src/model/theory_selector.tsx
```diff
@@ -48,22 +47,69 @@ export function TheorySelector(props: TheorySelectorProps) {
...
+                        >
                             {(meta) => (
                                 <div class="theory">
                                     <input
                                         type="radio"
                                         name="theory"
                                         id={meta.id}
                                         value={meta.id}
-                                        onchange={(evt) => props.setTheory(evt.target.value)}
+                                        onchange={(evt) => {
```
epatters: Ditto about big event handlers: better to make the separate helper functions/closures rather than inline JSX.

packages/frontend/src/theory/types.ts
```diff
@@ -6,6 +6,66 @@ import type { DiagramAnalysisComponent, ModelAnalysisComponent } from "../analys
 import { uniqueIndexArray } from "../util/indexing";
 import type { ArrowStyle } from "../visualization";
 
+/** Data about privileged migrations a frontend theory can have.
+ */
+export class MapData {
```
epatters: I'm not completely sure but my instinct is that this whole type should be define in Rust and exposed via Wasm. We can discuss the design more.

epatters (changes requested):
Thanks Kris! It's great to see this functionality really being fleshed out. I've left some comments below, most to do with the interface between Rust and TypeScript, a tricky business. Let's chat more when we're both in the office.

packages/frontend/src/theory/types.ts
kris-brown: Thanks I like this idea! I wasn't sure where the data structure belonged in the code itself but I defined it just above where it gets used (in `model.rs`) to start with.

packages/catlog-wasm/src/model.rs
kris-brown: I switched to ObType and MorType, which made the code much nicer in the frontend, too. However, I'm now getting an error (`Error: null pointer passed to rust wasm`) that I can't really resolve (when I print out the values before passing them to rust, they certainly don't look null to me). So this latest commit is broken.

kris-brown: Ah I changed the argument to `DblModel.pushforward` in the wasm from a `MapData` to a `&MapData` and that solved this!

packages/frontend/src/model/model_editor.tsx
kris-brown: Ok this one is now gone!

packages/frontend/src/model/document.ts
kris-brown: So the PR is now behaving correct - (EDIT: another problem which is resolved now)

kris-brown: Ok, the types are a bit funky if I don't coerce to string, but it works, at least until we start handling tabulator theories:

```typescript 
    const object_types = new Map<string | Mor, string | MorType>();
    const mor_types = new Map<
        string | { dom: Mor; cod: Mor; pre: Mor; post: Mor },
        string | ObType
    >();
```

epatters (commented):
Sorry that I let this PR languish for so long, @kris-brown. I'll take responsibility for moving this feature forward from here.

Recent developments are making me rethink the approach to model migration:

- Motivated by the open notebooks feature, @olynch is currently refactoring the Wasm boundary so that the whole notebook will cross from JS to Rust. While we must still handle the problem of lifting the theory change from the model in catlog to the model notebook, we will at least be able to perform the whole computation in Rust. Much of the challenge and churn in this PR had to do with the delicate interplay of data moving b/w JS and Rust. I expect that to become much cleaner after Owen's refactor is complete.
- A discrete double theory is a wrapper around a `FgCategory`; likewise, I think a morphism b/w such theories should be a newtype wrapper around an `FgFunctor` (yet to be defined). I'd like to start working on that in parallel with Owen's work. The trait sketched here for a general mapping b/w double theories is a good start. I'll probably borrow from that.

In short, I'm going to pause this PR again (sorry!) but resume working on a key piece of it very soon.

epatters:
I'm adding functionality related to this PR one piece at a time:

- #552
- #568 
- #660

PR #397, opened 2025-02-12
Have menu bar float with scroll in model

### LINKED ISSUE #358: Have menu bar float with scroll in model
This would be cute.

jmoggr:
Closes #358 

If the screen is less than 1100px wide there was an issue with the toolbar overlapping the content:

image

To work around this the toolbar is only sticky if the screen is at least 1100px wide


epatters (approved):
Thanks Jason! Did a quick test, works for me.

Issue #335, opened 2025-01-20
Show an error page when unauthorized access is attempted


epatters:
Currently, an authorization error is logged to the console and a (mostly) blank page is displayed.

Follow up to #330.

PR #398, opened 2025-02-12
Show an error message when unauthorized access is attempted

jmoggr:
Implements #335

I could not find any existing general purpose user facing error handling in the the repo, so I took the liberty of adding some.

This will show a modal for any uncaught error and shows a more specific message for permissions errors.

image

I'm somewhat nervous introducing several new patterns (~5) hours into a new job, but here we are :)

why a dialog over a page?
- it's more transparent in the code how errors are handled
- it allows parts of the page to keep working (when we eventually get other parts)
- we don't have to mess with client side routing

Let me know what you think!

### DIFF
--- a/packages/frontend/src/diagram/diagram_editor.tsx
+++ b/packages/frontend/src/diagram/diagram_editor.tsx
@@ -1,6 +1,6 @@
 import { MultiProvider } from "@solid-primitives/context";
 import { A, useParams } from "@solidjs/router";
-import { ErrorBoundary, Match, Show, Switch, createResource, useContext } from "solid-js";
+import { Match, Show, Switch, createResource, useContext } from "solid-js";
 import invariant from "tiny-invariant";
 
 import { useApi } from "../api";
--- a/packages/frontend/src/diagram/diagram_editor.tsx
+++ b/packages/frontend/src/diagram/diagram_editor.tsx
@@ -1,6 +1,6 @@
 import { MultiProvider } from "@solid-primitives/context";
 import { A, useParams } from "@solidjs/router";
-import { Match, Show, Switch, createResource, useContext } from "solid-js";
+import { ErrorBoundary, Match, Show, Switch, createResource, useContext } from "solid-js";
 import invariant from "tiny-invariant";
 
 import { useApi } from "../api";
@@ -30,6 +30,7 @@ import {
     newDiagramMorphismDecl,
     newDiagramObjectDecl,
 } from "./types";
+import { ErrorBoundaryDialog } from "../util/errors";
 
 import "./diagram_editor.css";
 
@@ -53,6 +54,7 @@ export function DiagramDocumentEditor(props: {
 }) {
     return (
         <div class="growable-container">
+            <ErrorBoundary fallback={(err) => <ErrorBoundaryDialog error={err} />}>
             <Toolbar>
                 <DiagramMenu liveDiagram={props.liveDiagram} />
                 <span class="filler" />
@@ -65,6 +67,7 @@ export function DiagramDocumentEditor(props: {
             <Show when={props.liveDiagram}>
                 {(liveDiagram) => <DiagramPane liveDiagram={liveDiagram()} />}
             </Show>
+            </ErrorBoundary>
         </div>
     );
 }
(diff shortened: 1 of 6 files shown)


jmoggr:
aside: is there any reason we're not having ci automatically apply `biome format --write && biome lint --write && biome check --write` with a new commit, if necessary?

packages/frontend/src/diagram/diagram_editor.tsx
```diff
@@ -53,18 +54,20 @@ export function DiagramDocumentEditor(props: {
...
-                <TheoryHelpButton theory={props.liveDiagram?.liveModel.theory()} />
-                <MaybePermissionsButton
-                    permissions={props.liveDiagram?.liveDoc.permissions}
-                    refId={props.liveDiagram?.refId}
-                />
-            </Toolbar>
-            <Show when={props.liveDiagram}>
-                {(liveDiagram) => <DiagramPane liveDiagram={liveDiagram()} />}
-            </Show>
+            <ErrorBoundary fallback={(err) => <ErrorBoundaryDialog error={err} />}>
```
epatters: A big question I have is where should the error boundaries go? Currently we've added them to the model and diagram editor components. Is there a reason not to just stick an error boundary at the top level of the app?

packages/frontend/src/util/errors.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+    let heading: string;
+    let message: string;
+
+    if (props.error instanceof PermissionsError) {
+        heading = "Permissions Error";
+        message = "You are not permitted to view this resource.";
+    } else {
+        heading = "Error";
+        message = "An unknown error occurred.";
```
epatters, line 23: I figure that once the known error types (currently just permissions error, maybe in the future additional types) have been exhausted, the error presumably arises from a bug in our application rather than any expected behavior. In this case, perhaps we should warn the user that something bad has happened and tell them how to file a bug report (file an issue on GitHub). That come could in a later PR, though.

epatters (commented):
Thanks Jason, this is great! As you gathered, we haven't had any systematic approach to error handling, which has been biting us in different. I like how clean and simple this one is.

epatters:
> aside: is there any reason we're not having ci automatically apply `biome format --write && biome lint --write && biome check --write` with a new commit, if necessary?

No particular reason, besides the fact that I've never used a workflow where the CI changes my commits, as opposed to me modifying the code to appease the CI. I'm open to trying it though if you're willing to tinker with [CI](https://github.com/ToposInstitute/CatColab/tree/main/.github/workflows). I know it would make other people happy as well to avoid having to remember to run the formatter and linter.

packages/frontend/src/diagram/diagram_editor.tsx
jmoggr: They should go at the top level of the app, and adding more of them further down the component hierarchy can help improve UX when errors happen.

Outside of the top level, the heuristic I use for wrapping a component in an error boundary is to consider whether users still have viable actions on the page if the component fails. This typically means wrapping the root of navigation components like panels, and "app" components like the model/diagram editor. This increases the chance that the user can easily recover without reloading or closing the page. In my experience pages typically end up with somewhere between 1 and 4 error boundaries.

For now I have moved the error boundary to the top level.

packages/frontend/src/util/errors.tsx
jmoggr: I created an issue for instructing users to file bug reports [here](https://github.com/ToposInstitute/CatColab/issues/403)

On this topic, it can be beneficial to log client side errors on the backend for a number of reasons, I created a ticket for that [here](https://github.com/ToposInstitute/CatColab/issues/404).

jmoggr:
> > aside: is there any reason we're not having ci automatically apply `biome format --write && biome lint --write && biome check --write` with a new commit, if necessary?
> 
> No particular reason, besides the fact that I've never used a workflow where the CI changes my commits, as opposed to me modifying the code to appease the CI. But I'm certainly open to trying it if you're willing to tinker with [CI](https://github.com/ToposInstitute/CatColab/tree/main/.github/workflows). I know it would make other people happy as well to avoid having to remember to run the formatter and linter.

I'll spend 30-45min, which should be sufficient. I've done this a couple of times before and have always really appreciated having it.

packages/frontend/src/diagram/diagram_editor.tsx
epatters: Thanks for that explanation, Jason. Makes sense.

PR #399, opened 2025-02-12
Mitigate failed test cleanups by logging in if the user already exists

### LINKED ISSUE #395: Vite test runner cleanup unreliable
Jason and Evan have both experienced 
```
FAIL  src/api/user_rpc.test.ts [ src/api/user_rpc.test.ts ]
FirebaseError: Firebase: Error (auth/email-already-in-use).
```
when a previous run of the tests didn't get cleaned up correctly, making the test user already present in the back-end.

It would be great to figure out why the cleanup doesn't always work. Barring that, we could at least randomize the test user's email to avoid this particular bug

jmoggr:
Closes #395 

This does cause the following test to sometimes fail:

```
    const defaultProfile = unwrap(await rpc.get_active_user_profile.query());
    test.sequential("should get empty profile after user creation", () => {
        assert.strictEqual(defaultProfile.username, null);
        assert.strictEqual(defaultProfile.displayName, null);
    });
```

ideas:
- this test might not be that important, so it could be removed
- conditionally perform this test if the account was actually created


packages/frontend/src/api/document_rpc.test.ts
```diff
@@ -1,10 +1,11 @@
...
 import { BrowserWebSocketClientAdapter } from "@automerge/automerge-repo-network-websocket";
 import { type FirebaseOptions, initializeApp } from "firebase/app";
-import { createUserWithEmailAndPassword, deleteUser, getAuth, signOut } from "firebase/auth";
+import { deleteUser, getAuth, signOut } from "firebase/auth";
 import * as uuid from "uuid";
 import { assert, afterAll, describe, test } from "vitest";
 
+import { initTestUserAuth } from "../util/test_util.ts";
 import { createRpcClient } from "./rpc.ts";
 import { unwrap, unwrapErr } from "./test_util.ts";
```
epatters: I can't think of a good reason to have two different modules for testing utils at this stage, so I'd suggest moving the content of `api/test_util.ts` to your new `util/test_util.ts` and then deleting the former file.

epatters (changes requested):
Thanks, this will notably improve the robustness of the tests!

- To your question about the test that now sometimes fails, I'm fine with just deleting it since `get_active_user_profile` will still be tested elsewhere, which seems the most important thing.
- As a GitHub thing, if you write "Closes [my issue number]" instead of "Addresses [my issue number]", merging the PR will automatically close the issue, which is convenient

packages/frontend/src/api/document_rpc.test.ts
jmoggr: I:
- Edited the PR to match conventions
- consolidated test utils in `util/test_util.ts`
- Removed the empty user profile test

Do you have a convention for who marks these kinds of conversations as resolved?

epatters: Thanks for the changes! No fixed convention. If you find it helpful to resolve them yourself to keep track of what you've done, that's fine; otherwise, I'll do it.

epatters (approved):

PR #400, opened 2025-02-13
Help page for quick introduction

KevinDCarlson:
A demo page! Should there be some more text, a wren, or a Topos logo on here, or is this fair enough?


epatters (approved):
Thanks Kevin. I like Brendan's idea that this page should involve to one that is generally shareable, say with an embed of your to-be-recorded five minute intro video. With that in mind, I've renamed it. So yes, in the future the content should be bulked up slightly but this seems fine for tomorrow.

PR #401, opened 2025-02-13
Links for embedding models elsewhere

hamidahoderinwale:
Button to copy the link added to the permissions dialog.


epatters:
Now that we have a permissions dialog, this should amount to a small and localized change: add a "Copy Link" button to the permissions dialog. The "share" dialog in Google Docs is a good example of what this should look like.

PR #407, opened 2025-02-14
Merge menu components for the model and diagram editor

jmoggr:
I'm not sure if we actually want to do this right now, but this is how you'd do it.


packages/frontend/src/analysis/analysis_editor.tsx
```diff
@@ -142,24 +141,28 @@ export function AnalysisDocumentEditor(props: {
...
+            case "diagram":
+                return props.liveAnalysis.liveDiagram;
+            case "model":
+                return props.liveAnalysis.liveModel;
+            default:
+                return undefined;
+        }
+    })();
+
+    console.log("here");
```
KevinDCarlson: clear this

packages/frontend/src/model/document.ts
```diff
@@ -32,6 +32,9 @@ export const newModelDocument = (theory: string): ModelDocument => ({
 Contains a live document for the model, plus various memos of derived data.
  */
 export type LiveModelDocument = {
+    /** discriminator for use in union types */
+    type: "model";
```
KevinDCarlson, line 36: For a `LiveModelDocument` `d`, this information should already be available in `d.liveDoc.doc.type`. Is there any reason we can't drag it out of there and avoid the duplication?

packages/frontend/src/diagram/document.ts
```diff
@@ -40,6 +40,9 @@ export const newDiagramDocument = (modelRef: StableRef): DiagramDocument => ({
 /** A diagram document "live" for editing.
  */
 export type LiveDiagramDocument = {
+    /** discriminator for use in union types */
+    type: "diagram";
+
```
KevinDCarlson, line 45: Same complaint as for `LiveModelDocument`

packages/frontend/src/components/json_import.tsx
KevinDCarlson, line 1: Remark that these changes are orthogonal to the menu merge task and should be merged into `json_io` either way.

packages/frontend/src/analysis/analysis_editor.tsx
KevinDCarlson, line 1: This feels nontrivially nicer to me.

packages/frontend/src/components/live_document_menu.tsx
```diff
@@ -0,0 +1,152 @@
...
+    return (
+        <AppMenu disabled={props.liveDocument === undefined}>
+            <Show when={props.liveDocument}>
+                {(liveDocument) => <LiveDocumentMenuItems liveDocument={liveDocument()} />}
+            </Show>
+        </AppMenu>
+    );
+}
+
+/** Menu items for a diagram in a model. */
```
KevinDCarlson: This and the above docstring are inaccurate, aren't they? Looks like it's for a diagram *or* a model.

packages/frontend/src/components/live_document_menu.tsx
```diff
@@ -0,0 +1,152 @@
...
+    const onNewDiagram = async () => {
+        const modelRefId = (() => {
+            switch (props.liveDocument.type) {
+                case "diagram":
+                    return props.liveDocument.liveModel.refId;
+                case "model":
+                    return props.liveDocument.refId;
+                default:
+                    assertExhaustive(props.liveDocument);
+            }
```
KevinDCarlson: I'm curious about the context for this convention for exhausting switches. Is the idea that you'd otherwise be either hand-writing the default case every time or else letting default cases fall through? I don't actually know what TS does in a failed switch.

KevinDCarlson (approved):
I like this change all in all. It's not a huge shift but it seems more reasonable and is slightly negative in LOC and in number of files.

KevinDCarlson (changes requested):
Oops, didn't mean to approve here.

packages/frontend/src/components/live_document_menu.tsx
jmoggr: This allows missing cases to be caught at compile time, if you replaces this block with:
```
        const modelRefId = (() => {
            switch (props.liveDocument.type) {
                case "diagram":
                    return props.liveDocument.liveModel.refId;
                default:
                    assertExhaustive(props.liveDocument);
            }
        })();
```

It would produce a type error `Argument of type 'LiveModelDocument' is not assignable to parameter of type 'never'.ts` because TS knows that there are some unhanleded cases and tries to pass them as arguments to `assertExhaustive` which is only allowed to accept the `never` type (which is what remains after all cases have been exhausted). The behavior of this at runtime is exactly the same as throwing, but it allows problems to be caught during development.

packages/frontend/src/model/document.ts
jmoggr: `d.liveDoc.doc.type` does not discriminate between the members of the union `LiveDiagramDocument | LiveModelDocument` for reasons that are not immediately obvious to me. I'm pretty sure it has to do with some interaction between templates and union types.

If you'd like, I could try to re-work the types to fix this, but I'm only 75% confident that it is solvable without creating extra mess. If it's solvable nicely it'd probably take me between 5 and 30min.

packages/frontend/src/components/live_document_menu.tsx
jmoggr: Yes, this was a cut and paste error.

packages/frontend/src/model/document.ts
KevinDCarlson: @epatters Your call on whether to just merge this as is or chase down why my suggestion doesn't immediately work. (Or ignore this branch for now because of this annoyance.)

epatters: It would be nice if TypeScript were smart enough to infer this. OTOH, even if it were, it might be kind of annoying to reach down three levels to discriminate the type. So all-in-all I'm fine to lift the discriminator field up to this level if that seems more convenient.

This situation suggests to me that we ought to refactor the `Live[*]Document` types so that there is a `LiveDocument` base type with a common `type` field among others, just like we have for (non-live) `[*]Document` types. But let's save that for later to avoid creating a big diff here.

packages/frontend/src/components/live_document_menu.tsx
epatters, line 1: With apologies that it's not documented anywhere (please feel free to correct that), my intention was that `src/components` be for more-or-less generic components that could, in principle, be used in other apps, whereas `src/page` is for stuff specific to the CatColab app like menu bars and toolbars. On that logic, this file should be in `src/page`.

Open to other ideas about how to organize things if you have them!

epatters (changes requested):
Thanks Jason, I like this change! A left one comment to address now, and another that we can consider for the future. Once the first one is addressed, feel free to merge.

packages/frontend/src/components/live_document_menu.tsx
jmoggr: I moved the component and created and issue for the `Live[*]Document` refactor: https://github.com/ToposInstitute/CatColab/issues/417

epatters (approved):

PR #420, opened 2025-02-21
ts_vector column with content from database

hamidahoderinwale:
This PR adds a new column for TS_VECTOR that's automatically updated with data from the `content` column in the database.

Includes GIN Index.


packages/backend/build.rs
```diff
@@ -0,0 +1,5 @@
+// generated by `sqlx migrate build-script`
+fn main() {
+    // trigger recompilation when a new migration is added
+    println!("cargo:rerun-if-changed=migrations");
+}
```
jmoggr, line 5: What is this for? I tried adding a new migration and nothing happened.

packages/backend/migration_status.txt
```diff
@@ -0,0 +1,4 @@
+20241004010448/installed document refs
+20241025030906/installed users
+20250114010822/installed test first five view
+20250121193606/installed testing
```
jmoggr, line 4: What is this file for? I can't find any reference to it.

It's somewhat concerning to be tracking the status of migrations outside of the db, the authoritative source for migration status is the `_sqlx_migrations` table. Let me know if you need any help with migration stuff.

packages/backend/migrations/20241004010448_document_refs.up.sql
```diff
@@ -11,5 +11,6 @@ CREATE TABLE refs (
     created TIMESTAMPTZ NOT NULL
 );
 
-ALTER TABLE snapshots
-    ADD FOREIGN KEY (for_ref) REFERENCES refs (id) DEFERRABLE INITIALLY DEFERRED;
+
+
+
```
jmoggr, line 24: Why are we modifying a migration? That should be avoided whenever possible, the migration files are hashed and stored in `_sqlx_migrations.checksum`, and migrations will fail to apply if the checksums don't match.

It is possible to do this, but it's a real pain.

packages/backend/migrations/20250114010822_test_first_five_view.down.sql
```diff
@@ -0,0 +1,12 @@
...
+FROM users
+LIMIT 2;
+
+SELECT * 
+FROM permission_level
+LIMIT 2; 
+
+SELECT * 
+FROM permissions
+LIMIT 2;
```
jmoggr, line 12: Why is this in a down migration script? It looks like a test of some kind, but that doesn't make sense to put in a migration.

packages/backend/migrations/20250121193606_testing.down.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM _sqlx_migrations WHERE version = '20250121193606';
```
jmoggr, line 1: Why is this in a migration? This is very concerning, especially without any comment on why it's needed.

If you want to rewrite migration history it's not advisable to do it from inside a migration, it typically should be done manually. It can technically work sometimes, but you can run into unfun dependency problems.

In general the only time you want to rewrite migration history is when there is an ongoing incident due to a botched migration. It should be seen as a last resort when things are actively on fire.

packages/backend/migrations/20250121200530_testing.down.sql
```diff
@@ -0,0 +1 @@
+-- Add down migration script here
```
jmoggr, line 1: why do we have empty migrations?

packages/backend/src/test.rs
```diff
@@ -0,0 +1,21 @@
...
+        .fetch_all(&pool)
+        .await?;
+
+    // Print each row
+    for row in rows {
+        println!("{:?}", row);
+    }
+
+    Ok(())
+}
```
jmoggr, line 21: Where does the 'Catcolab' table come from? (also why is there a capital in a table name, and if there is, it needs to be wrapped in quotes)

This test will always pass, which isn't much of a test.

packages/backend/src/search.rs
```diff
@@ -0,0 +1,14 @@
...
+pub fn 
+
+// take an enum for each instance of a model, subtyped based on the theory 
+
+pub enum modelInstance {
+    title, 
+    theory, 
+
+}
+// taking each title of model as keyword / tokenizing
```
jmoggr: This seems like it was maybe not meant to be included?

jmoggr (commented):
The migration for adding the gin index looks fine (I haven't tested it yet), however I'm very confused about everything outside of that.

It seems like you had some issues with managing the database while testing migrations and that may have lead to a lot of unnecessary changes.

Developing database migrations can be really tricky, I don't intend to be unjustly critical, and am happy to help walk you through any questions or confusions you might have.

hamidahoderinwale:
> The migration for adding the gin index looks fine (I haven't tested it yet), however I'm very confused about everything outside of that.
> 
> It seems like you had some issues with managing the database while testing migrations, and that may have lead to a lot of unnecessary changes.
> 
> Developing database migrations can be really tricky, I don't intend to be unjustly critical and am happy to help walk you through any questions or confusions you might have.

I think the confusion comes from merging what was originally two seperate PRs into a single one. I have the GIN index and started implemented the full-text search (more info on the project issue). There were some migrations that I tried to edit and had to rerun/make deletions, etc.

epatters:
Closing since this PR has gotten quite stale. We can revive certain pieces as needed in a fresh branch.

PR #429, opened 2025-03-01
Rename AlgebraicJuliaService to CatColabInterop and refactor

quffaro:


epatters (changes requested):
Thanks Matt! Although we're naming the Julia package `CatColabInterop`, I think the folder in this repo should be called `packages/algjulia-interop`. The discrepancy might seem strange but from the perspective of the Julia general registry, the Julia package is about interoperation with CatColab, whereas from the perspective of this repo, the package is about interoperating with AlgebraicJulia! What do others think about this?

packages/algjulia-interop/jupyter_server.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-KERNEL_NAME="julia-1.11"
-#KERNEL_NAME="julia-ajaas-1.11"
+KERNEL="julia-ccl-interop-1.11"
```
epatters: IMO, by default, we should not assume that the sys image is being used.

More important, though also more work, would be to make this script take command-line flags:

- `--kernel` (default: output of `julia -e 'print("julia-$(VERSION.major).$(VERSION.minor)")'`)
- `--mode`: one of `production`, `staging`, `development` (default: `production`), resulting in `ORIGIN` be set appropriately
- `--origin`: direct specification of `ORIGIN` if `mode` is omitted

Then the user should never have to edit this script.

packages/algjulia-interop/make_sysimage.jl
```diff
@@ -8,16 +8,17 @@ using PackageCompiler
...
                 precompile_execution_file="sysimage_precompile.jl")
 sysimg_path=joinpath(@__DIR__, sysimg);
 
 @info "Adding $sysimg_path to IJulia kernel"
 Pkg.activate(); Pkg.add("IJulia")
 using IJulia
+Pkg.build("IJulia")
 
-installkernel("Julia AJaaS", "--project=@.", "--sysimage=$sysimg_path")
+installkernel("Julia CCL Interop", "--project=@.", "--sysimage=$sysimg_path")
```
epatters: Minor point, but I'd prefer the clearer if slightly more verbose: "Julia CatColab Interop"

packages/algjulia-interop/src/decapodes-service/analysis/initial_conditions.jl
```diff
@@ -84,7 +79,7 @@ function initial_conditions(ics::Dict{Symbol,<:InitialConditions}, geometry::Geo
 end
 
 function initial_conditions(x::InitialConditions, args...)
-    throw(ImplError("These initial conditions ($(x)) are"))
+    throw(ImplError("These initial conditions ($(x)) are")) # TODO
```
epatters: Lingering TODO: can we just delete this method? IMO, in Julia it's usually better to not define a method at all than to define and throw a not-implemented error.

packages/algjulia-interop/src/decapodes-service/model.jl
```diff
@@ -70,15 +70,20 @@ function add_to_model!(model::Model{ThDecapode}, content::AbstractDict, type::Ho
...
 # for each cell, if it is...
 #   ...an object, we convert its type to a symbol and add it to the modeldict
 #   ...a morphism, we add it to the modeldict with a field for the ids of its
 #       domain and codomain to its
-function Model(::ThDecapode, model::AbstractVector{<:AbstractDict}) # AbstractDict is the JSON
+function Model(::ThDecapode, json_model::JSON3.Object) # AbstractDict is the JSON
     newmodel = Model(ThDecapode())
-    foreach(model) do cell
+    __name = json_model[:name] # TODO unused
+    foreach(json_model[:notebook][:cells]) do cell
```
epatters: More idiomatic to replace `foreach` + do notation with a simple `for` loop.

packages/algjulia-interop/src/decapodes-service/model.jl
```diff
@@ -88,3 +93,15 @@ function Model(::ThDecapode, model::AbstractVector{<:AbstractDict}) # AbstractDi
     return newmodel
 end
 export Model
+
+function Model(::ThDecapode, json_array::JSON3.Array{T}; name="model") where T
+    newmodel = Model(ThDecapode())
+    foreach(json_array) do cell
```
epatters: Ditto here and elsewhere.

packages/algjulia-interop/src/decapodes-service/analysis/simulation.jl
```diff
@@ -100,7 +118,9 @@ end
 
 """ for the variables in a system, associate them to their state values over the duration of the simulation """
 function variables_state(soln::ODESolution, system::PodeSystem)
-    Dict([ system.uuiddict[var] => state_entire_sim(soln, system, var) for var ∈ system.plotvar ])
+    plottedVars = [ k for (k, v) in system.plotVars if v == true ]
+    uuid2symb = Dict([ v => k for (k, v) in system.uuiddict]) # TODO why reverse again?
```
epatters, line 120: Lingering TODO: seems resolvable before merging

packages/frontend/src/stdlib/analyses/decapodes.tsx
```diff
@@ -195,7 +195,10 @@ export function Decapodes(props: DiagramAnalysisProps<DecapodesContent>) {
                 </IconButton>
             </Match>
             <Match when={kernel.error || options.error}>
-                <IconButton onClick={restartKernel} tooltip="Restart the AlgebraicJulia service">
+                <IconButton
```
epatters: Since the user is gonna have to know about kernels anyway, perhaps it's clear to replace "AlgebraicJulia Interop service" with simply "Julia kernel" here and below?

packages/algjulia-interop/testsim.jl
epatters, line 1: This file seems to have been accidentally committed.

packages/algjulia-interop/test/test_jsons/diffconst_payload1.json
epatters, line 1: Was this filed exported from CatColab? Why is "payload" in the filename? Why isn't it in a subfolder like the others?

packages/algjulia-interop/test/test_jsons/payload.json
epatters, line 1: Same comment about this file.

packages/algjulia-interop/src/decapodes-service/diagram.jl
```diff
@@ -0,0 +1,149 @@
...
+function Base.nameof(model::Model, content::AbstractDict)
+    if isnothing(content[:over])
+        :no_name
+    else
+        Symbol(model.data[content[:over][:content]].name)
+    end
+end
+
+# TODO we are restricted to Op1
+function add_to_pode!(d::SummationDecapode,
```
epatters: This and the other method `add_to_pode!` have a bunch of arguments, but most of them are already in the `DecapodeDiagram` struct defined below. I would suggest making these be methods `add_to_pode!(d::DecapodeDiagram, ...)`. This will require slightly refactoring your `Diagram` code below so that the `DecapodeDiagram` is created *before* you have to call `add_to_pode!` but I think it will ultimately simplify this code and make it easier to read.

epatters (changes requested):
Thanks Matt, this is shaping up well! I left some comments below.

packages/algjulia-interop/test/test_jsons/diffconst_payload1.json
quffaro: the "payloads" verify that the data we pass into the `Analysis` function on `decapodes.tsx:362` work. they're not exported from CatColab; instead, I grabbed them from the console.

Should I put them into a `payloads` (or better name) folder?

epatters (approved):
Thanks Matt, looks great! I will rebase this and merge.

PR #436, opened 2025-03-03
Add Jason's ssh public key to catcolab-next

jmoggr:
Instructions: https://catcolab.org/dev/unstable-264.xml

my understanding:
- the instructions will re-encrypt .env.next.age using my public key as well
- that will allow me to decrypt the .env.next secrets locally
- .env.next will have the ssh keys for the catcolab-next AWS instance


epatters:
Hi Jason, you should now be set up as a user on the staging server, also with permissions to log in as the main user `catcolab`. Can you confirm that the following works from your machine?

```bash
ssh catcolab@backend-next.catcolab.org 
```

jmoggr:
> Can you confirm that the following works from your machine?

I confirm that the command works from my machine.

Thanks!

epatters (approved):
Great, you should be good to go then!

PR #445, opened 2025-03-10
Landing page

hamidahoderinwale:
This PR is an overlay/landing page.


packages/frontend/src/notebook/landing_page.css
```diff
@@ -0,0 +1,43 @@
+.overlay {
```
jmoggr: I think this should be renamed to  `walkthrough_overlay.css`, since there is no landing page component it's not obvious from the file name what this file relates to.

packages/frontend/src/notebook/notebook_editor.css
```diff
@@ -23,7 +23,284 @@
...
+.overlay {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    position: fixed;
+    opacity: 0;
+    visibility: hidden;
+    transition: all 0.3s ease;
+    z-index: 1000;
+}
```
jmoggr: Is there any reason why much of this CSS was not included in `landing_page.css`? That seems like a much more natural spot for it and it would avoid splitting `.overlay` selector over two files.

packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -36,6 +37,260 @@ import {
 
 import "./notebook_editor.css";
 
+export function WalkthroughOverlay(props: { isOpen: boolean; onClose: () => void }) {
```
jmoggr: I think this component should get moved into it's own file. Both the walkthrough the notebook editor are going to get a lot bigger before this project is over, also their implementation seems almost entirely orthogonal.

jmoggr (commented):
Code looks good! I just have some nits about organisation. And the build should work of course.

packages/frontend/src/notebook/notebook_editor.css
hamidahoderinwale: There's an overlay for the theory selection overlay and the landing page overlay with different styling so I separated them.

jmoggr: I see. Then they should probably use different css classes. The two overlays seem completely different and having them use the same class is going to cause confusion in the future.

hamidahoderinwale:
Made updates, thank you!

jmoggr:
Thanks for the changes! The changest for the landing page looks good.

The merge conflicts are unfortunate, I imagine they are caused by the frequent force pushes. Personally I would cut and paste the changes into a new branch rather than deal with the merge conflicts or try to untangle git...

jmoggr (approved):

hamidahoderinwale:
Should be fixed and made some small edits! 

> Thanks for the changes! The changest for the landing page looks good.
> 
> The merge conflicts are unfortunate, I imagine they are caused by the frequent force pushes.

KevinDCarlson (commented):
- I'm not sure about the unpredictably rotating model images on the

quffaro:
Hey @KevinDCarlson 

> I'm not sure about the unpredictably rotating model images on the first page of the carousel. Maybe they could be scrollable inside the overlay. Also, maybe a video that shows something easy to help people get started rather than just the coolest stuff. The vorticity image, in particular, resizes the overlay so that buttons can become inaccessible.

We can coerce the modal dialog to be a fixed size, but since videos are recorded with different aspect ratios, they would be unhappily smushed. I recorded a demo of making an SIR model+analysis and made the only item on the carousel for now. The slides on the dialog are different sizes still, which I think I can now fix, but I found it to be less bothersome now that the first slide was also not dynamically sizing to its content.

> Font sizes for "Next" and "Get Started" buttons are different in my view.

The buttons used different css classes with different font sizes, so i've moved the font size specification to a parent class, which makes the font 18px. I've also switched the ordering so "Next" goes in front, to signify progressing through the dialog.

> Somewhat ambivalent about the second page, I'm tempted to have these buttons go into a bit of exposition of what we actually mean to have them not come across as vacuous. (That would be written by a non-Hamidah, probably!)

C.f., [this issue](https://github.com/ToposInstitute/CatColab/issues/559)

> The third page seems pretty close but having slightly more informative labels would be nice. Maybe "Example models" instead of "Introduction", "Mathematical Documentation" instead of "Developer Documentation".

I changed "Introduction" to "Example models," but the "Developer Documentation" button links to a page titled "Developer Documentation," so I didn't change it. Shifting outside scope, should the documentation title be changed  to express that its audience are both mathematicians and developers?

quffaro:
After some discussion yesterday, I agree that the overlay appearance should be more conservative. I think typical behavior for landing pages is that they appear whenever a **new user** opens a new notebook, so I've added an overlay condition that the user must be not logged in. I think modal dialogs should communicate, implicitly or explicitly, why they appear to the user, I recommend adding a Sign Up button to the overlay.

Some modern websites have more subtler ways of introducing themselves to the user which I think should be punted to a future PR.

packages/frontend/src/model/model_editor.tsx
```diff
@@ -145,6 +145,7 @@ export function ModelNotebookEditor(props: {
                 cellConstructors={cellConstructors()}
                 cellLabel={judgmentLabel}
                 duplicateCell={duplicateModelJudgment}
+                notebookType={"model"}
```
epatters: We shouldn't introduce this new attribute `notebookType`. By design, notebooks are generic UI component that doesn't know about the larger ontology of the app.

(And if we *did* introduce it, it's missing for diagram notebooks.)

packages/frontend/src/notebook/notebook_editor.css
```diff
@@ -1,10 +1,10 @@
 .notebook-container {
     --gutter-width: 50px;
-
     max-width: calc(100ex + var(--gutter-width));
     margin: 0 auto;
     padding-left: var(--gutter-width);
     padding-right: 5px;
+    font-family: var(--main-font);
```
epatters: Why is this added? Does this PR change now fonts appear in the notebook? I wouldn't expect that.

packages/frontend/src/notebook/notebook_editor.css
```diff
@@ -23,7 +23,20 @@
...
+    gap: 1rem;
+}
+
+.resource-link:last-child {
+    grid-column: 1 / -1;
+    grid-row: 3;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    background-color: #f0f0f0;
```
epatters: This shouldn't be here. If it's needed, it should go with the other CSS for this new component.

packages/frontend/src/notebook/walkthrough_overlay.tsx
epatters, line 1: I'd suggest calling this component `WelcomeOverlay` since it's no longer really a walkthrough.

packages/frontend/src/notebook/walkthrough_overlay.css
```diff
@@ -0,0 +1,270 @@
...
+    padding: 0.75rem 1rem;
+    background-color: #f9f9f9;
+}
+
+.topos-logo {
+    height: 35px;
+    width: auto;
+}
+
+.media-container {
```
epatters: I assume that `media-container` is no longer relevant since the video/gif is gone.

Can you comb through this CSS file, which has gotten pretty big, and remove any other cruft that is no longer needed at this stage of the design?

packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -78,8 +83,15 @@ export function NotebookEditor<T>(props: {
...
     noShortcuts?: boolean;
+    notebookType?: string;
 }) {
+    const firebaseApp = useFirebaseApp();
+    const auth = useAuth(getAuth(firebaseApp));
+
     const [activeCell, setActiveCell] = createSignal(props.notebook.cells.length > 0 ? 0 : -1);
+    const [isOverlayOpen, setOverlayOpen] = createSignal(
+        props.notebookType === "model" && props.notebook.cells.length === 0 && auth.data == null,
+    );
```
epatters: If the welcome overlay should only be shown in models, then the component should be added to the `ModelEditor` rather than the `NotebookEditor`. Then you can avoid this `notebookType` business as noted above.

packages/frontend/src/notebook/notebook_editor.css
epatters, line 1: For the sake of a clean diff and to avoid accidental changes, let's ensure that there are *zero* diffs to the `notebook` folder (so, to this file and `notebook_editor.tsx`).

packages/frontend/src/model/welcome_overlay.css
```diff
@@ -0,0 +1,198 @@
...
+}
+
+.navigation-buttons {
+    display: flex;
+    justify-content: center;
+    padding: 10px 0 15px 0;
+    gap: 15px;
+}
+
+.nav-button {
```
epatters: Class names like `header-container`, `footer-container`, `navigation-buttons`, and `nav-button` are too generic to appear at the top-level of the CSS file for the welcome overlay. Two standard fixes are to either

- manually prefix them with something like `welcome-`, or
- turn this file into a CSS module to automatically mangle the names

epatters (changes requested):

tim-at-topos:
I like this a lot! My suggestion would be to change the layout (and content) of the 5 links to the following

```
|         CatColab overview       |
| Example models |   Help pages   |
|  Source code   | Zulip chatroom |
```

PR #446, opened 2025-03-10
Automerge storage adapter for Postgres DB

jmoggr:
I've only been able to think of one way where we might end up with a race conditions: if you expect an automerge doc and a snapshot to be on the same version. But that should be pretty easy to avoid.


packages/automerge-doc-server/package.json
```diff
@@ -7,23 +7,29 @@
...
         "socket.io-client": "^4.8.0",
         "ws": "^8.18.0"
     },
     "devDependencies": {
         "@biomejs/biome": "^1.8.3",
         "@types/express": "^4.17.21",
         "@types/node": "^22.7.4",
+        "@types/pg": "^8.11.11",
         "@types/ws": "^8.5.12",
+        "automerge-repo": "^0.1.0",
```
epatters: Added by accident, I assume.

packages/automerge-doc-server/src/postgres_storage_adapter.ts
```diff
@@ -0,0 +1,75 @@
...
+    private pool: Pool;
+
+    constructor(pool: Pool) {
+        this.pool = pool;
+    }
+
+    // Load the data for an exact key match
+    async load(key: StorageKey): Promise<Uint8Array | undefined> {
+        const result = await this.pool.query(
+            "SELECT data FROM storage WHERE key = $1",
```
epatters: I figure there are database migrations that should go with this PR. Based on a quick look, we've got:

- a new table called `storage` with the columns `data` and `key`
- a new column in `refs` called `doc_id`

packages/backend/src/document.rs
```diff
@@ -95,24 +97,42 @@ pub async fn save_snapshot(state: AppState, data: RefContent) -> Result<(), AppE
...
-    } else {
-        // Otherwise, fetch the content from the database and create a new
-        // Automerge doc handle.
-        let content = head_snapshot(state.clone(), ref_id).await?;
-        let data = RefContent { ref_id, content };
-        let ack = automerge_io.emit_with_ack::<Vec<String>>("create_doc", data).unwrap();
-        let response = ack.await?;
-        Ok(response.data[0].to_string())
+
+    // Expecting an array of responses instead of a single response for unknowable reasons
```
epatters: Haha, yes, when I writing the original code I wondered why `emit_with_ack` returns an array. I wish I knew.

packages/backend/src/document.rs
```diff
@@ -95,24 +97,42 @@ pub async fn save_snapshot(state: AppState, data: RefContent) -> Result<(), AppE
...
+    let ack = automerge_io
+        .emit_with_ack::<Vec<GetDocSocketResponse>>("get_doc", ref_id)
+        .unwrap();
+    let response_array = ack.await?.data;
+
+    // Extract the first response
+    let response = response_array
+        .into_iter()
+        .next()
+        .ok_or_else(|| AppError::AutomergeServer("Empty ack response".to_string()))?;
```
epatters: A good idea to catch this case rather than just unwrapping like I did before.

packages/automerge-doc-server/src/postgres_storage_adapter.test.ts
epatters, line 1: Just to check my understanding: for the time being,

- this file is a stub
- the file `storage_adapter_test.ts` is a stub copied from upstream

epatters (commented):
Thanks Jason, this looks like a great start.

I'm thinking about the overall design. As we've discussed, a two-process design is inevitable unless and until `automerge-repo-rs` gets more love. So the only question is what logic goes in which process. Here's what I'm thinking:

- It makes sense to have the storage adapter write directly to the database from the Node process to minimize latency. Also, the interface is already defined in TS by automerge-repo, so we might as well localize the logic in the implementation of the interface
- The logic and DB queries to associate refs with Automerge doc IDs, here implemented in `setDocIdForRef` and `getDocIdForRef`, should be moved to the Rust backend

The advantage of this design is that it creates a clear separation of concerns: the main Rust backend is the only process that touches the `ref` table, and the Automerge document server is the only process that touches the `storage` table. So we shouldn't have to worry about the processes stomping on each other. In fact, since neither table links to the other, they could live in separate databases! I'm not suggesting that we put them in separate databases right now but having this flexibility down the road might be useful.

What do you think about this approach?

epatters:
Coincidentally, while perusing the Automerge Discord this afternoon, I found that someone else has already implemented an Automerge storage adapter for Postgres: <https://www.npmjs.com/package/automerge-repo-storage-postgres>

The code is pretty similar to what you have. We might consider using this one to reduce our code surface, though it doesn't have any tests either.

packages/automerge-doc-server/package.json
jmoggr: not entirely, but it was unintentionally left in and I have removed it.

I was attempting to use the storage adapter tests from the repo directly. Unfortunately the teardown handler wasn't async so I had to cut and paste the tests.

packages/automerge-doc-server/src/postgres_storage_adapter.ts
jmoggr: yup, they are now added

"it works on my machine"... Ideally we'd be testing the migrations in CI, but given the infrequency of db changes I doubt that's worth it right now.

packages/automerge-doc-server/src/postgres_storage_adapter.test.ts
jmoggr: Yup, this file runs the tests that were copied from upstream. I have merged the two files, I don't think it makes sense to keep them separate.

jmoggr:
> What do you think about this approach?

I strongly agree.

I had a second go at it and feel much better about the result. We are no longer lazily initializing the automerge docs and they should now have the same lifetime as the rest of the ref. This greatly simplifies things and addresses your concerns.

My original idea was to consolidate everything to do with the creation and management of the automerge docs, but that was messier and had fewer advantages than I thought it would.

I wrote a migration script that can populate the `storage` table. It can be run via `npm run migrate-storage` from the `automerge-doc-server`. After we do that we can add a `NOT NULL` constraint to the `refs.doc_id` column, which is nice.

packages/backend/src/document.rs
```diff
@@ -95,24 +102,70 @@ pub async fn save_snapshot(state: AppState, data: RefContent) -> Result<(), AppE
...
+
+    match response {
+        Ok(_) => Ok(()),
+        Err(err) => Err(AppError::AutomergeServer(err)),
+    }
+}
+
+async fn create_automerge_doc(automerge_io: &SocketIo, content: Value) -> Result<String, AppError> {
```
jmoggr: moved the code for calling SocketIo to a function so that my eyes would bleed less

jmoggr:
> Coincidentally, while perusing the Automerge Discord this afternoon, I found that someone else has already implemented an Automerge storage adapter for Postgres: https://www.npmjs.com/package/automerge-repo-storage-postgres
> 
> The code is pretty similar to what you have. We might consider using this one to reduce our code surface, though it doesn't have any tests either.

I am mildly against this. I personally think that the lifetime tech debt of ~300 lines of code (including tests) is less than that of a dependency with no tests and a bus factor of 1. However I'm fairly biased towards LoC over dependencies compared to most devs.

In my own time I'll open a PR with the tests, if that gets accepted then I would be mildly in favour of switching.

> Discord
That helps explain why I couldn't find anything through web search... this is a sad state of affairs.

epatters:
> I am mildly against this. I personally think that the lifetime tech debt of ~300 lines of code (including tests) is less than that of a dependency with no tests and a bus factor of 1. However I'm fairly biased towards LoC over dependencies compared to most devs.

I think that's a perfectly reasonable perspective. My only comment is that if we go with our version, then we owe it our future selves to write some tests/adapt the tests copied from upstream, since we can no longer comfort ourselves by outsourcing the testing to someone else :)

epatters (commented):
Anyway, things are looking nice and clean! Before merging, I suppose we ought to:

- finish fixing the CI (#452)
- finish adapting the tests for the Postgres storage adapter

jmoggr:
> finish adapting the tests for the Postgres storage adapter

I figured out how to call the `runStorageAdapterTests` function from `automerge-repo`. It looks like I was hitting some kind of bug with vitest. The result is that the tests should now be comfortably outsourced.

In the future if the document keys used by our cleanup function stop matching the ones used upstream, then the tests should fail because the cleanup function would no longer be cleaning up the correct keys.

epatters (commented):
I'm getting back to this PR now. Apologies for the long delay, @jmoggr.

I'm happy with the general design but I think one omission is how it interacts with snapshots. It's understandable that this wasn't very clear since this feature of the backend is not yet exposed in the UI, but the idea is that the user could tag a new snapshot of a ref at a particular time, possibly even giving it a version number. This will become important once models can depend on other models.

Since we are following the Automerge devs' advice that the Automerge doc, rather than its JSON-ification, should be the source of truth, we should:

- Not add a `doc_id` to a ref; thus, the `refs` table will remain unchanged
- Instead, associate a `doc_id` to each snapshot
- When a new snapshot is created from the current head, we call [`Repo.clone`](https://automerge.org/automerge-repo/classes/_automerge_automerge_repo.Repo.html#clone) to create a cloned doc handle from the current one and add the new doc to the storage.

Tagging @olynch for thoughts since he was the architect of the refs/snapshots design for the original Node backend, which later underwent some minor changes when I ported the backend to Rust.

jmoggr:
Thanks Evan. Admittedly I gave little consideration to snapshots, as they have no existing functionality to test against (we only ever reference the head snapshot) and look somewhat similar to version control.

Are snapshots meant to be immutable? If so, then the all mappings from `refs.id` to `snapshot.doc_id` would reference ostensibly immutable documents. In this case, what `doc_id` would be used by the clients for live editing the current 'head'?

If they are meant to be mutable, then that opens another can of worms.

epatters:
Thanks for trying to clarify this confusing point. So far, we've worked under the idea that the head snapshot is "live" and mutable, whereas any other snapshot is "archived" and thus should not be mutated. When a snapshot is saved, a new head is created and the old head ceases to be mutable.

Though it's not yet implemented, it should be possible to restore from an old snapshot. That implies that the snapshots should be Automerge docs, since otherwise we would lose any rich text annotations.

So there are two questions:

1) Does the intended semantics of snapshots described above seem right?
2) If it is, should we take any precautions to ensure that old snapshots don't get mutated?

jmoggr:
1. Yup!
2. Some enforcement is necessary. Otherwise, references (dependencies) in other documents would be implicitly unstable.

### Enforcing Snapshot Immutability
There doesn't seem to be a straightforward way to enforce this. Documents must be loaded via automerge-doc-server to be viewed, but it doesn't expose a public API for enforcing read-only mode.

Two possible approaches:
1. Export the document on the server and import it on the client (this avoids a "live" connection)
2. Enforce read-only mode client-side.

Option 2 seems fine for now and it would be easy to implement option 1 later.

### Snapshot Overhead
Snapshots would ideally be lightweight (just referencing a point in a documents history), but are currently deep copies. While this costs DB space, it's unlikely to become a major issue soon.

I briefly explored lightweight snapshots. My understanding is that the Automerge storage API support referencing a specific versions by ID, but this isn’t exposed in the repo API. I may be misunderstanding something, or it could be a relatively simple missing feature.

### Mutable Snapshots
I initially reacted negatively to the idea that the latest snapshot is mutable. On reflection, I can see how this might simplify things. Just flagging that it was briefly confusing - not objecting.

epatters:
Great, we're on the same page.

> Snapshots would ideally be lightweight (just referencing a point in a documents history), but are currently deep copies. While this costs DB space, it's unlikely to become a major issue soon.
> 
> I briefly explored lightweight snapshots. My understanding is that the Automerge storage API support referencing a specific versions by ID, but this isn’t exposed in the repo API. I may be misunderstanding something, or it could be a relatively simple missing feature.

Yes, my impression is that the Automerge devs regard an regard an Automerge doc as like a repo in a version control system, but even if that's the case, VCS features like branching and tagging don't seem to be exposed in the public APIs. So I agree that for now we should proceed by cloning (deep copying) the document, even if it would preferable for snapshots to be more lightweight.

jmoggr:
When we create a new snapshot we update `refs.head` to point to the newly created snapshot, that new snapshot will have a different doc id. All clients accessing the doc when it is snapshotted will still be connected to the old doc_id, which is now "read only".

I'm wondering if @epatters had thoughts on this, if not I could enumerate possible solutions.

(this is a non blocking for now, everything still technically "works" despite this problem)

packages/backend/src/rpc.rs
```diff
@@ -17,7 +17,7 @@ pub fn router() -> Router<AppState> {
         .handler(new_ref)
         .handler(get_doc)
         .handler(head_snapshot)
-        .handler(save_snapshot)
+        .handler(create_snapshot)
```
jmoggr, line 20: I renamed this since we are no longer passing content to be saved. However it might make sense to revert to `save_snapshot`.

packages/backend/src/rpc.rs
```diff
@@ -94,12 +94,13 @@ async fn _head_snapshot(ctx: AppCtx, ref_id: Uuid) -> Result<Value, AppError> {
...
-    _save_snapshot(ctx, data).await.into()
-}
-async fn _save_snapshot(ctx: AppCtx, data: doc::RefContent) -> Result<(), AppError> {
-    auth::authorize(&ctx, data.ref_id, PermissionLevel::Write).await?;
-    doc::save_snapshot(ctx.state, data).await
+async fn create_snapshot(ctx: AppCtx, ref_id: Uuid) -> RpcResult<()> {
+    async {
+        auth::authorize(&ctx, ref_id, PermissionLevel::Write).await?;
+        doc::create_snapshot(ctx.state, ref_id).await
+    }
```
jmoggr, line 101: I found this easier to read compared to the two similarly named functions we had before, however it breaks that convention. I should probably either follow the existing convention, or update the other examples to follow this. Thoughts?

packages/automerge-doc-server/src/socket.ts
```diff
@@ -1,37 +1,60 @@
...
-/** Encapsulates socket.io for internal communication with the backend.
+// Convert the type of each property from `(...args) => ret` to `(...args, (ret) => void) => void` to
+// match the signature expected for SocketIO listeners
+type Callbackify<T> = {
+    [K in keyof T]: T[K] extends (...args: infer Args) => infer Ret
+        ? (...args: [...Args, (res: Promise<Ret>) => void]) => void
+        : never;
+};
+
+function registerHandler<
```
jmoggr, line 26: I got annoyed at all the plumbing I had to do when modifying the handlers so removed the biolerplate.

packages/backend/src/rpc.rs
epatters: Good point, thanks!

packages/backend/src/rpc.rs
epatters: This is definitely better! I just didn't know I could create standalone `async` blocks, so I did a workaround. Gotta RFTM :)

This would be a great change to make consistently in a follow-up PR.

packages/automerge-doc-server/package.json
```diff
@@ -5,24 +5,30 @@
     "main": "dist/main.js",
     "scripts": {
         "main": "tsx src/main.ts",
+        "migrate-storage": "tsx src/populate_automerge_storage.ts",
```
epatters: I assume we don't want to leave this script around forever. What is its intended lifecycle? Do we run it first on staging and the eventually on production, then remove it?

Should I run it on staging immediately after merging this PR?

epatters (approved):
Thanks Jason, I think is ready to merge! My remaining question concerns the workflow for doing so (see below).

jmoggr:
The commit history was really messy. I squashed all of the commits made before Evans last comment. All changes since then are in the most recent commit.

PR #450, opened 2025-03-14
Automate the deployment of the backend for catcolab-next

### LINKED ISSUE #349: Update backend code upon push to main
It's currently quite hard to work on the backend because there's no automatic updating of the code on which the staging (nor production) backend runs, even when `main` is updated. There are fancier things one could imagine but at the very least, we'd like the CI to turn off the process running the staging backend, update the backend Rust, rebuild, and restart upon pushed to `main`. 

Evan says

> It's something like:
>
> Connect to staging server with SSH
> Pull latest changes into the copy of the CatColab git repo there
> Run `catcolab-build` (and fail CI if this command fails)
> Assuming the above worked, run `catcolab-restart`
> 
> The `catcolab-` commands were added by Shaowei as a convenience to the Nix scripts.

jmoggr:
This creates a new user `deployuser` which is restricted to running a single command `catcolab-update`. When the `deployuser` user connects via SSH the `catcolab-update` script will start running and the ssh session will terminate when the script exits. If the script exists unsuccessfully the gitub action will be marked as failed.

We need to give the `deployuser` the same access to the sudo command as the `catcolab` user because `deployuser` needs to run the same `systemctl` commands to manage services. 

Additional configuration required from Eval:
- Create a new ssh keypair
- assign the public key to the `catcolabnextDeployuserKey` variable in `infrastructure/hosts/catcolab-next/backend.nix`
- assign the private key to a new github action repository secret named `CATCOLAB_NEXT_DEPLOYUSER_KEY`

Notes:
- I'm only 60% confident that this will work on the first go. These changes were cut and pasted from a hacked together config targeting a separate deployment. It is most likely that something trivial will go wrong in `catcolab-update`
- This does not automate the backend deployment to the production catcolab backend, however it should be easy to cut and paste the changes after we've confirmed that they work on staging.

Closes #349.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -193,7 +193,7 @@ jobs:
           mkdir -p ~/.ssh
           echo "${{ secrets.CATCOLAB_NEXT_DEPLOYUSER_KEY }}" > ~/.ssh/id_ed25519
           chmod 600 ~/.ssh/id_ed25519
-          ssh-keyscan ec2-18-217-205-18.us-east-2.compute.amazonaws.com >> ~/.ssh/known_hosts
+          ssh-keyscan backend-next.catcolab.org >> ~/.ssh/known_hosts
 
       - name: Deploy via SSH
-        run: ssh deployuser@ec2-18-217-205-18.us-east-2.compute.amazonaws.com
+        run: ssh deployuser@backend-next.catcolab.org
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -179,3 +179,21 @@ jobs:
         with:
           message: |
             Preview url: ${{ steps.url_preview.outputs.NETLIFY_PREVIEW_URL }}
+
+  deploy_backend:
+    name: Deploy backend to AWS
+    runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' && github.ref_name == 'main'
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up SSH key
+        run: |
+          mkdir -p ~/.ssh
+          echo "${{ secrets.CATCOLAB_NEXT_DEPLOYUSER_KEY }}" > ~/.ssh/id_ed25519
+          chmod 600 ~/.ssh/id_ed25519
+          ssh-keyscan ec2-18-217-205-18.us-east-2.compute.amazonaws.com >> ~/.ssh/known_hosts
+
+      - name: Deploy via SSH
+        run: ssh deployuser@ec2-18-217-205-18.us-east-2.compute.amazonaws.com
--- a/infrastructure/hosts/catcolab-next/backend.nix
+++ b/infrastructure/hosts/catcolab-next/backend.nix
@@ -1,6 +1,8 @@
 { inputs, pkgs, config, ... }:
 
 let
+    catcolabnextDeployuserKey = "";
+
     automergePort = "8010";
     backendPort = "8000";
 
@@ -73,6 +75,30 @@ let
         /var/lib/catcolab/infrastructure/scripts/build.sh
     '';
 
+    updateScript = pkgs.writeShellScriptBin "catcolab-update" ''
+        #!/usr/bin/env bash
+
+        set -e
+
+        echo -e "\n#### stoping services...\n"
+        catcolab-stop
+
+        echo -e "\n#### pulling changes...\n"
+        cd /var/lib/catcolab
+        git pull --force
+
+        echo -e "\n#### applying migrations...\n"
+        catcolab-migrate
+
+        echo -e "\n#### building...\n"
+        catcolab-build
+
+        echo -e "\n#### starting services...\n"
+        catcolab-start
+
+        echo -e "\n#### update finished...\n"
+    '';
+
     packages = with pkgs; [
         rustup
         nodejs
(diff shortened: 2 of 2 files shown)


infrastructure/hosts/catcolab-next/backend.nix
```diff
@@ -183,17 +210,28 @@ in {
...
-        users = [ "catcolab" ]; 
+        users = [ "catcolab" "deployuser" ];
         commands = [
             { command = "/run/current-system/sw/bin/systemctl start automerge"; options = [ "NOPASSWD" ]; } 
             { command = "/run/current-system/sw/bin/systemctl stop automerge"; options = [ "NOPASSWD" ]; } 
             { command = "/run/current-system/sw/bin/systemctl restart automerge"; options = [ "NOPASSWD" ]; }
             { command = "/run/current-system/sw/bin/systemctl start backend"; options = [ "NOPASSWD" ]; } 
             { command = "/run/current-system/sw/bin/systemctl stop backend"; options = [ "NOPASSWD" ]; } 
             { command = "/run/current-system/sw/bin/systemctl restart backend"; options = [ "NOPASSWD" ]; }
+            { command = "${updateScript}/bin/catcolab-update";  options = [ "NOPASSWD" ]; }
```
jmoggr, line 221: The way the existing scripts are created and used is strange blend of manual unixy install steps and declarative nixy configuration. Mixing them seems wrong so I added the new one the nix way. I'm not sure if this was the right call

infrastructure/hosts/catcolab-next/backend.nix
```diff
@@ -183,17 +210,28 @@ in {
...
         ]; 
     }];
 
+
+    users.users.deployuser = {
```
jmoggr, line 225: I was unable to put this user with the other users in `default.nix` without adding a layer of abstraction to the nix configs (there is no easy/clean way to access `updateScript` from `default.nix`)

infrastructure/hosts/catcolab-next/backend.nix
```diff
@@ -183,17 +210,28 @@ in {
...
+        isNormalUser = true;
+        openssh.authorizedKeys.keys = [
+            ''
+                command="${updateScript}/bin/catcolab-update",no-port-forwarding,no-agent-forwarding,no-X11-forwarding,no-pty ${catcolabnextDeployuserKey}
```
jmoggr, line 230: Declaration and use of the existing scripts is strange blend of manual unixy install steps and declarative nixy configuration. Mixing them seems wrong so I added the new one the nix way. I'm not sure if this was the right call.

infrastructure/hosts/catcolab-next/backend.nix
```diff
@@ -194,6 +221,16 @@ in {
         ]; 
     }];
 
+
+    users.users.deployuser = {
+        isNormalUser = true;
+        openssh.authorizedKeys.keys = [
+            ''
+                command="${lib.getExe updateScript}",no-port-forwarding,no-agent-forwarding,no-X11-forwarding,no-pty ${catcolabnextDeployuserKey}
```
jmoggr, line 229: Declaration and use of the existing scripts is strange blend of manual unixy steps and declarative nixy configuration. Mixing them seems wrong so I added the `updateScript` the nix way. I'm not sure if this was the right call.

.github/workflows/deploy.yml
```diff
@@ -179,3 +179,21 @@ jobs:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up SSH key
+        run: |
+          mkdir -p ~/.ssh
+          echo "${{ secrets.CATCOLAB_NEXT_DEPLOYUSER_KEY }}" > ~/.ssh/id_ed25519
+          chmod 600 ~/.ssh/id_ed25519
+          ssh-keyscan ec2-18-217-205-18.us-east-2.compute.amazonaws.com >> ~/.ssh/known_hosts
```
epatters: Would it still work if we replace this line with

```sh
ssh-keyscan backend-next.catcolab.org >> ~/.ssh/known_hosts
```

.github/workflows/deploy.yml
```diff
@@ -179,3 +179,21 @@ jobs:
...
+
+      - name: Deploy via SSH
+        run: ssh deployuser@ec2-18-217-205-18.us-east-2.compute.amazonaws.com
```
epatters: And ditto here:

```sh
ssh deployuser@backend-next.catcolab.org
```

I think that should work.

epatters (commented):
Thanks Jason, this will be a nice quality of life improvement!

I left a question/comment on the GitHub action. Regarding the Nix config, I remain a Nix noob so I'm happy to defer to your judgment on how to structure the files. If you want another opinion, we can ask Owen.

infrastructure/hosts/catcolab-next/backend.nix
epatters: Again, I'm not the right person to advise here, but if you see other ways to make the Nix config more idiomatic and/or declarative, please go ahead and do so!

.github/workflows/deploy.yml
jmoggr: yup! I pushed the changes

.github/workflows/deploy.yml
jmoggr: That works, I applied the change

epatters (approved):

PR #452, opened 2025-03-15
Fixes to automatic deployment of staging backend

epatters:
Fixes to #450.

### DIFF
--- a/infrastructure/hosts/catcolab-next/backend.nix
+++ b/infrastructure/hosts/catcolab-next/backend.nix
@@ -1,7 +1,7 @@
 { lib, inputs, pkgs, config, ... }:
 
 let
-    catcolabnextDeployuserKey = "";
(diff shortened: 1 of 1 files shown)


epatters:
@jmoggr, the SSH keys are now properly set up, I think. The next error is:

```
#### pulling changes...

fatal: detected dubious ownership in repository at '/var/lib/catcolab'
To add an exception for this directory, call:

	git config --global --add safe.directory /var/lib/catcolab
```

I could follow this advice but I haven't looked into the implications. I'll let you take it here.

jmoggr:
I fixed a couple of permissions issues relating to `deployuser` not being the same as the `catcolab` user. I had neglected to test this with existing data owned by catcolab...

@epatters could you re-encrypt `.env.next.age` with the private key for `deployuser`? `deployuser` is currently unable to decrypt `.env.next.age`. I've already added the `deployuser` public key to `secrets.nix`, so you should only have to do the re-encrytion: `agenix -r -i /deployuser/privatekey .env.next.age`

epatters:
@jmoggr, I've re-keyed the file and deployed catcolab-next. I can't tell if `deployuser` has actually now has the desired permissions, so please let me know if not.

jmoggr:
These changes were included in the nixification branch. Closing as stale.

PR #455, opened 2025-03-24
Manage backend services with nix and share configuration across hosts

jmoggr:
Sorry for the size... this basically turned into a rewrite of the nix configs

#### Nginx replaced with Caddy
I was unable to get certbot to setup SSL certs when changing the hostnames used by nginx. I'm unsure how the certs were originally configured. Caddy configures certbot for us and provides a nice wrapper for creating reverse proxies, this greatly simplifies the webserver config.

#### Removal of the nix `lib` files
Both myself and a number of LLMs were unable to understand the .nix files in `infrastructure/lib` enough to successfully modify them. They were very clever but were only used to simplify the definitions `nixosConfigurations` and `deploy.nodes` which saved ~20 lines of code in flake.nix.

#### Consolidated users
Removed individual logins for users, users may still login as `catcolab` or `root`.

#### Moved flake.nix to repo root
This makes my life a lot easier by letting me initialise the nix dev shell from the repo root.

### Todo
- The backup service will need to be tested when we deploy to prod, I was unable to test this so there could be some issues there
- Prod will need a deployuser associated with it, it would probably be fine to reuese the one for catcolab-next
- Update documentation
- Remove catcolab-jmoggr references
  - I'd like to hold off on this until after we have deployed production. I really appreciate having a development environment that I can break without worry.

You can see the a successful example of the deploy running in a cloned repo [here](https://github.com/jmoggr/CatColab/actions/runs/14041844690/job/39313626333) (note that the netlify build is failing)


infrastructure/hosts/catcolab-next/backend.nix
epatters, line 1: Wondering about this file. I see that `../catcolab/backend.nix` is removed but this file still exists.

packages/automerge-doc-server/package.json
```diff
@@ -14,6 +13,7 @@
     "dependencies": {
         "@automerge/automerge-repo": "^1.2.1",
         "@automerge/automerge-repo-network-websocket": "^1.2.1",
+        "esbuild": "^0.25.1",
```
epatters: Should this be in `devDependencies`?

packages/automerge-doc-server/default.nix
```diff
@@ -0,0 +1,54 @@
...
+    # the wasm file for automerge
+    automerge_wasm_path=$(find node_modules/.pnpm -path "*/wasm_bindgen_output/nodejs/automerge_wasm_bg.wasm" 2>/dev/null | head -n 1)
+    if [ -z "$automerge_wasm_path" ]; then
+      echo "❌ Error: Node.js automerge WASM file not found!"
+      exit 1
+    fi
+    cp "$automerge_wasm_path" "$out/"
+  '';
+
+  # TODO: I have been unable to bust this hash, and I don't know why
```
epatters: I don't know what this means or whether it is important.

packages/backend/src/document.rs
```diff
@@ -13,6 +13,18 @@ pub async fn new_ref(ctx: AppCtx, content: Value) -> Result<Uuid, AppError> {
...
+    let query = sqlx::query!(
+        "
+        INSERT INTO users(id, created, signed_in)
+        VALUES ($1, NOW(), NOW())
+        ON CONFLICT (id) DO UPDATE
+        SET signed_in = EXCLUDED.signed_in
+        ",
+        user_id,
+    );
+    query.execute(&mut *transaction).await?;
```
epatters, line 26: I'm surprised to see a nontrivial code change in this PR. A couple questions/comments:

- This looks like mostly a copy-paste from `user::sign_up_or_sign_in`, which would be nice to avoid
- Why is it needed here? IIRC, in the current design it is the frontend's responsibility to call the above RPC procedure to notify the backend that such a sign-up/sign-in event has occurred

infrastructure/hosts/catcolab-next/backend.nix
jmoggr: Good catch, this was accidentally left in and has been removed.

packages/automerge-doc-server/package.json
jmoggr: Yes, but I just removed it since it's not being used. Nix is providing esbuild in `automerge-doc-server/defualt.nix`

packages/automerge-doc-server/default.nix
jmoggr: Rules around when the hash needs to be updated are looser than I thought. I removed this comment and pushed instructions to the `automerge-doc-server/README.md` for updating the hash.

packages/backend/src/document.rs
jmoggr: Thanks for noticing that. On a clean DB when the user first loads the site and is sent to a new blank document, the user is not yet created when `new_ref` is called. This would be a pre-existing bug exposed by using a clean DB.

This was only meant as a temporary fix while I addressed the problem upstream. I'll put together a fix in a separate PR.

.github/workflows/deploy.yml
```diff
@@ -193,7 +193,7 @@ jobs:
           mkdir -p ~/.ssh
           echo "${{ secrets.CATCOLAB_NEXT_DEPLOYUSER_KEY }}" > ~/.ssh/id_ed25519
           chmod 600 ~/.ssh/id_ed25519
-          ssh-keyscan backend-next.catcolab.org >> ~/.ssh/known_hosts
+          ssh-keyscan backend-next.jmoggr.com >> ~/.ssh/known_hosts
```
olynch: why are we now deploying to jmoggr.com?

infrastructure/modules/host.nix
```diff
@@ -0,0 +1,101 @@
...
+    #!/usr/bin/env bash
+    set -ex
+
+    branch="nixification"
+
+    if [ -d "catcolab" ]; then
+      rm -rf ./catcolab
+    fi
+
+    git clone -b "$branch" https://github.com/ToposInstitute/CatColab.git catcolab
```
olynch: If this is running in CI, it should take the git hash of the commit that it is running on

.github/workflows/deploy.yml
olynch: Is this just for testing?

packages/automerge-doc-server/default.nix
```diff
@@ -0,0 +1,55 @@
...
+  ];
+
+  buildInputs = with pkgs; [
+    nodejs_23
+  ];
+
+  installPhase = ''
+    mkdir -p $out/
+
+    # We use esbuild instead of tsc for building, as it bundles all required JavaScript into a single
```
olynch, line 27: Ah, excellent! I was wondering how you were going to deal with that

olynch (approved):
Overall, this seems very good. 

The lib/*.nix stuff was just cruft I've been copypasta-ing around as I setup new servers so it's probably good to get rid of it.

I have one major and one minor concern. The minor concern is that the deploy script for CI doesn't seem to use the current hash. The major concern is how this all impacts deploy speed. One benefit of the old approach is that if one step broke, you could iterate on it interactively fairly quickly. Nix makes you do everything from scratch; if I have to wait ~5 min every time before the problem that I'm trying to debug even shows up, then that's a big pain. Perhaps the esbuild trick addresses this somehow? What are the average build times?

.github/workflows/deploy.yml
epatters: Yeah, this is just for testing. It should be changed before merging. But thanks for noticing!

PR #459, opened 2025-03-31
Log users out if their session is stale

### LINKED ISSUE #458: Remove code duplicated in `new_ref` from `sign_up_or_sign_in`
The code was duplicated because, on a clean database, when a user first loads the site and is redirected to a new blank document, the user hasn't been created yet when new_ref is called.

By the time new_ref is invoked, the user should already be signed in—so this likely indicates a bug in the authentication flow or some client-side initialization code.

original discussion:
https://github.com/ToposInstitute/CatColab/pull/455/files/1aa6c7f78a207a37ffaaadf095c9eb916d6dffcf#r2010900419

jmoggr:
closes #458

This SHOULD fix the issues with the deployment of the `nixification` branch

#### Stale sessions
A reset or new/empty DB will have no user information from it, but the app will not invalidate the users session in the browser even though it now references a user_id which no longer exists. This results in requests failing when `ctx.user.user_id`, which is provided by the session, fails to match any `id` in the `users` table. 

#### 458 hack
The workaround introduced in #458 was implemented without understanding the root cause (stale sessions) and introduced a new problem: it caused the query to fail when the user is unauthenticated, since `user.user_id` will be null, which violates the `NOT NULL` constraint on `users.id` in the DB.

It is likely that anyone with a valid session would still have been able to connect to `next.catcolab.org` would not have experience any issue. And the issue could have been worked around by navigating directly to `next.catcolab.org/profile` and logging in from there.

### Analysis
It is very likely that some issue in the future will require us to invalidate users sessions, and it is better to have something in place to do that before it is needed. It also makes the developer experience nicer by removing something that is often forgotten about when developing/testing.


packages/frontend/src/App.tsx
```diff
@@ -36,6 +38,24 @@ const Root = (props: RouteSectionProps<unknown>) => {
 
     const api: Api = { serverHost, rpc, repo };
 
+    const [sessionInvalid, setSessionInvalid] = createSignal(false);
+    createEffect(() => {
+        (async () => {
```
epatters: This method of getting an async function into an effect handler seems questionable. Would it work to use a resource instead, something like

```typescript
const [isSessionValid] = createResource(async () => {
  [...]
  return result.tag === "Ok"
}, {
  initalValue: true,
});
```

jmoggr: I agree, that was some react brain rot leaking through. I've replaces the signal with a resource.

epatters (approved):
Looks great, thanks!

PR #460, opened 2025-03-31
Remove `vite-plugin-top-level-await` to fix source maps

jmoggr:
For some unknown reason the vite plugin `topLevelAwait` was stripping the source mapping comments (`//# sourceMappingURL=...`) from the compiled .js files, and this is what was causing source mapping to not work.

The `top-level-await` feature is included in `es2022`, so we can drop the polyfill and target the newer version. Vite by default targets `es2020` so this would remove support for browsers from 2020-2022, however I don't think we should be putting any effort into supporting browsers that old anyway.

I'll add a discussion point about which browsers we should support to the next meeting.

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -70,7 +70,6 @@
         "typescript": "^5.2.2",
         "vite": "^5.3.1",
         "vite-plugin-solid": "^2.10.2",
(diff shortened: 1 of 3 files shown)


epatters (approved):
Thanks for the fix, Jason!

PR #462, opened 2025-04-02
Remove deployuser script and run build commands over ssh

jmoggr:
tl;dr: the deploy process no longer uses a script on the remote host and instead copies the repo and runs commands over SSH.

#### deploy script
Remove the deploy script managed by nix in favour of running commands over ssh from github actions.

I think using Nix to manage the deploy script was a mistake. It made it too easy to misreason about the script, since the one being executed was different from the one in source control. (#461 is similar to this).

#### `deployuser`
`deployuser` is no longer restricted and is functionally equivalent to root (similar to the `catcolab` user). In practice locking down `deployuser` added little extra security, anybody who was able to access the secrets for it would also be able to commit changes which could then be deployed. I initially set it up this way under a (possibly misguided) notion of defence-in-depth and to keep as much of the process in nix as possible with the script. 

We could simplify the setup slightly be removing `deployuser` and just using `root` instead, but should work for now and we can easily change it in the future.

#### Fixing the issue with the commit SHA not being available publicly
no idea what going on here, it never seems to work on TopoInstitute, and works most(?) of the time on my fork.

#### Using deploy-rs from the github action
This would be ideal, but I was unable to get caching to work for /nix/store in the github action. You'd think this would be easy, but I was unable to get any 3rd party github actions to work and managing the caching myself (which I've done before) resulted in cryptic errors.

Caching in github actions is fairly limited, and it seems like most people using nix+github actions are using some sort of 3rd party service for handling caching. That seems overly complicated for what we need.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -184,8 +184,11 @@ jobs:
     name: Deploy backend to AWS
     runs-on: ubuntu-latest
     if: github.event_name != 'pull_request' && github.ref_name == 'main'
+    env:
+      TARGET_HOST: backend-next.catcolab.org
+      HOST_CONFIG: catcolab-next
     steps:
-      - name: Checkout repository
+      - name: Checkout
         uses: actions/checkout@v4
 
       - name: Set up SSH key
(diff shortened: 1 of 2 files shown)


epatters (approved):
Thanks for the detailed explanation!

I would support removing the `deployuser` user in the future if that will simplify things without meaningful change in security.

PR #463, opened 2025-04-07
Wait until document is loaded to use fully defined document object

jmoggr:
This moves the responsibility of checking if the document is loaded to the page root. The goal is to reduce the number of states page components have to handle. Previously they all had to separately handle the case where the document wasn't loaded yet and implement their own approach for doing nothing.

This does not change the current behaviour of the app, however it would make it harder to implement certain loading effects in the future. I argue that this trade the tradeoff of simplicity now vs polish later favours simplicity: this would be easy to rollback in the future when/if that level of polish is required.


packages/frontend/src/page/document_loading_screen.tsx
```diff
@@ -0,0 +1,13 @@
+import { TheoryHelpButton, Toolbar } from "./toolbar";
+
+export default function DocumentLoadingScreen() {
+    return (
+        <div class="growable-container">
+            <Toolbar>
+                <span class="filler" />
+                <TheoryHelpButton />
+            </Toolbar>
+            <div class="notebook-container">Loading...</div>
```
epatters: The one thing that I want to insist on is that we not flash "Loading..." text while the document is loading. I find it quite distracting and it highlights, rather than minimizes, what we prefer the user to not think about, namely that the document is loading.

We should also make this stub screen look a bit more like the pages that will be loaded. An easy thing would be to put the hamburger menu icon in the top left.

epatters (changes requested):
Thanks! I agree that this is the right tradeoff for us at this stage. Plus we're barely taking advantage of the extra flexibility of the current approach right now anyway.

packages/frontend/src/page/document_loading_screen.tsx
KevinDCarlson: Couldn't we add a dancing colorful sprite looking slightly sheepish and saying "Gosh, sorry this is taking so long!"

jmoggr: I removed the "Loading..." and added a hamburger menu. I think it will make sense and be easier to add more skeleton components after some more of the big picture UI design stuff has been resolved.

> Couldn't we add a dancing colorful sprite looking slightly sheepish and saying "Gosh, sorry this is taking so long!"

The most "correct" thing to do here is probably adding a ~0.5s timeout that shows some sort of loading indicator. This prevents a flash when loading is quick and gives some sort of user feedback when loading is slow. I don't think this is worth doing, but only because we don't already have a dancing colorful sprite :)

epatters: Thanks, this is already a lot better! And I agree, we can improve this further down the road when UI idioms are more stable.

epatters (approved):

Issue #447, opened 2025-03-12
Notion-like UI for navigation


quffaro:
I'm using this issue to open a discussion about improving file navigation in CatColab. Currently, a user navigates to, say, the model subtending a diagram by a link in the upper-right corner.

<img width="320" alt="Image" src="https://github.com/user-attachments/assets/02dc5b8e-9764-4672-b2e0-294b82b2be39" />

Since since references to the model and equations page are only visible to one another through a link, and not presented on the screen simultaneously, the user is not led to think of them as both "active," but instead you're "leaving" the page. Oppositely, modern UI features like tabbed browsing (which gives the impression that content exists but is just not in focus) might be a good feature for representing multiple different models open at once. Likewise, breadcrumb or filetree navigation might be better suited for guiding the user through the process of defining a model, its analysis, and maybe any "assets" that would be eventually be associated to an analysis.

Notion features some useful UI navigation widgets, including breadcrumbs and a file navigator. I think some of these features can be inspiration for ways to improve file navigation.

epatters:
Agreed. And I'm marking this as strategic since ideally we'd use it consistently across of the app, instead of doing one thing for analyses and another for diagrams like we are right now.

KevinDCarlson:
Reading of this issue from today's meeting: want the aspect of Notion behavior in which opening a new document pops up a half-width pane which can then be maximized if desired; not immediately obvious how to handle nesting three things at once.

epatters:
Closed by #749.

PR #464, opened 2025-04-07
Document breadcrumbs

jmoggr:
This is stacked on #463, but it would be trivial to rebase it onto main.

I tried ~2 other ways of loading the the information necessary for the breadcrumbs more efficiently, and they were both far more complicated than this. I did not feel that the additional complexity was worth it.

The document objects not having access to their own ref id was a continuous pain point. In a separate issue/PR it might be worth considering adding a documents id to the document object during it's creation.

### DIFF
--- a/packages/frontend/src/analysis/analysis_editor.tsx
+++ b/packages/frontend/src/analysis/analysis_editor.tsx
@@ -106,6 +106,7 @@ export function AnalysisDocumentEditor(props: {
                         >
                             <Toolbar>
                                 <AnalysisMenu liveAnalysis={props.liveAnalysis} />
+                                <DocumentBreadcrumbs document={props.liveAnalysis} />
                                 <span class="filler" />
                                 <TheoryHelpButton theory={theoryForAnalysis(props.liveAnalysis)} />
                                 <IconButton
@@ -134,8 +135,10 @@ export function AnalysisDocumentEditor(props: {
                             onExpand={() => setSidePanelOpen(true)}
                         >
                             <div class="notebook-container">
+                                <div class="toolbar">
+                                    <div class="toolbar-spacer" />
+                                </div>
                                 <h2>Analysis</h2>
-                                <DocumentBreadcrumbs document={props.liveAnalysis} />
                                 <AnalysisNotebookEditor liveAnalysis={props.liveAnalysis} />
                             </div>
                         </Resizable.Panel>
--- a/packages/frontend/src/analysis/analysis_editor.tsx
+++ b/packages/frontend/src/analysis/analysis_editor.tsx
@@ -22,7 +22,13 @@ import {
     NotebookEditor,
     newFormalCell,
 } from "../notebook";
-import { DocumentLoadingScreen, DocumentMenu, TheoryHelpButton, Toolbar } from "../page";
+import {
+    DocumentBreadcrumbs,
+    DocumentLoadingScreen,
+    DocumentMenu,
+    TheoryHelpButton,
+    Toolbar,
+} from "../page";
 import { TheoryLibraryContext } from "../stdlib";
 import type { AnalysisMeta } from "../theory";
 import { assertExhaustive } from "../util/assert_exhaustive";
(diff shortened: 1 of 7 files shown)


packages/frontend/src/page/document_breadcrumbs.tsx
```diff
@@ -0,0 +1,77 @@
...
+        // biome-ignore lint/style/noNonNullAssertion: the array initializer guarantees that there will always be at least one item in the array
+        const parentRefId = getParentRefId(documentChain[0]?.document!);
+        if (!parentRefId) {
+            break;
+        }
+
+        // In a worst case this results in sequential round trips to the server. However it should be
+        // reasonable to hope that either the parents are already in the local automerge repo, or that
+        // they will be needed by the app at some point in the near future. The alternative is picking
+        // apart a JSON blob in postgres, and that sounds neither fun nor maintainable.
```
epatters, line 68: Agreed that this seems the best approach for now. One nice thing about it that we wouldn't get by doing the query in the backend/DB is liveness/reactivity: if you edit the title of a parent document anywhere, it instantly updates in the breadcrumb. Pretty cool.

packages/frontend/src/page/document_breadcrumbs.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const [documentChain] = createResource(() => props.document, getDocumentChain);
+
+    return (
+        <div class="breadcrumbs-wrapper">
+            <Show when={documentChain()} fallback={<span>Loading...</span>}>
+                <For each={documentChain()}>
+                    {(doc, index) => (
+                        <>
+                            {index() > 0 && <span class="breadcrumb-spacer">/</span>}
+                            <a class="breadcrumb-link" href={`${doc.document.type}/${doc.refId}`}>
```
epatters: The link is broken. It should be absolute:

```typescript
<a class="breadcrumb-link" href={`/${doc.document.type}/${doc.refId}`}>
```

packages/frontend/src/page/document_breadcrumbs.tsx
```diff
@@ -0,0 +1,77 @@
...
+};
+
+export function DocumentBreadcrumbs(props: {
+    document: AnyLiveDocument;
+}) {
+    const [documentChain] = createResource(() => props.document, getDocumentChain);
+
+    return (
+        <div class="breadcrumbs-wrapper">
+            <Show when={documentChain()} fallback={<span>Loading...</span>}>
```
epatters: As in #463, I want to insist that we don't show the "Loading..." text as a fallback. In this case, I think we can just leave it blank while loading.

packages/frontend/src/analysis/analysis_editor.tsx
```diff
@@ -129,6 +135,7 @@ export function AnalysisDocumentEditor(props: {
                         >
                             <div class="notebook-container">
                                 <h2>Analysis</h2>
+                                <DocumentBreadcrumbs document={props.liveAnalysis} />
```
epatters: Currently, due to the inconsistent UI idioms around analyses (not having editable titles, and always appearing in a side bar), this looks a little weird. While I very much appreciate that you tested the breadcrumbs in this case too, I think that until we make the UI more consistent (#447), we should show the breadcrumb for *the document being analyzed* in the top-left, that is, [around here](https://github.com/ToposInstitute/CatColab/blob/6371b27c82e9800919d76a31b7e423ae9c2cdf3e/packages/frontend/src/analysis/analysis_editor.tsx#L102). What do you think?

epatters (changes requested):
Thanks Jason, both the code and the interface are very clean! A few comments below.

packages/frontend/src/analysis/analysis_editor.tsx
jmoggr: I agree strongly. It did look weird and this way the breadcrumbs are in a consistent place. Also the hamburger menu of the left pane is for the analysis, so putting the analysis breadcrumbs there follows that.

One of the reasons I tried putting the breadcrumbs on the right was so that the spacing at the top of the page would be more consistent, instead I've just added a spacer:

2025-05-12-133032_hyprshot

epatters (approved):

Issue #378, opened 2025-02-04
Look up documents by name and type


epatters:
UUIDs are perfectly unique but are also unwieldy and unfriendly. It should be possible to look up documents by a combination of

- type, e.g., "model" or "diagram,
- human-readable name

Fortunately, as of #369, the latter is now standardized across document types.

Implementing this involves:

- Add an RPC method the backend, backed by a suitable SQL query
- Exposing this through a nice entry widget in the frontend, with completion menu

epatters:
We've had the first part of this issue, an RPC method for querying documents, for some time now. It has not yet been integrated into the frontend for importing cells, but at this point we should track that through a more precise and timely issue.

PR #465, opened 2025-04-07
Add backend for quering refs and a documents page 

jmoggr:
This is my attempt to create backend functions that generalizes across a number of issues and anticipates future requests.

https://github.com/ToposInstitute/CatColab/issues/378
https://github.com/ToposInstitute/CatColab/pull/347
https://github.com/ToposInstitute/CatColab/issues/291

There two different sorts of queries that we need:

1. querying the documents associated with a user, which shows the relationship between the user and documents which they have permissions for
2. querying all documents, which shows documents without any relationship (outside of ownership)

1 would be used for a users profile, 2 would be used as a more general search function.

To make things simpler for consumers I use the same interface and techniques for both queries. This may be a bit overkill for something like the user profile right now, but I think the uniformity is nice and could allow for more powerful searches later.

Both queries only show documents that the user who is searching is allowed to see (determined by the user id in the session). This can be easily extended to account for publish documents.

I still need to do more testing and to add tests, but I would like to get some feedback before spending more time on this.


packages/backend/src/document.rs
```diff
@@ -123,3 +125,192 @@ pub struct RefContent {
...
+/// refs.
+#[derive(Clone, Debug, Serialize, Deserialize, TS)]
+pub struct RefStub {
+    pub name: String,
+    pub type_name: String,
+    pub ref_id: Uuid,
+    // permission level that the current user has on this ref
+    pub permission_level: PermissionLevel,
+    pub owner: UserSummary,
+    pub created_at: String,
```
epatters: Sad that JSON has no native support for datetimes. Given that:

- How do we know what the format of this datetime string will be?
- On the Rust side, should we be using [`chrono`](https://crates.io/crates/chrono) or [`time`](https://crates.io/crates/time)? Both are supported by [`sqlx`](https://docs.rs/sqlx/latest/sqlx/postgres/types/index.html), and at least `chrono` is supported by `ts-rs`.

packages/backend/src/document.rs
```diff
@@ -123,3 +125,192 @@ pub struct RefContent {
...
+        .as_ref()
+        .map(|user| user.user_id.clone())
+        .ok_or(AppError::Unauthorized)?;
+
+    // selects the ref id and ref name (from the most recent snapshot of that
+    // ref) for all refs that a user (the searcher) has permission to access.
+    // Optionally filter the results by the owner of the refs.
+    let results = sqlx::query!(
+        r#"
+        WITH latest_snapshots AS (
```
epatters: The latest snapshot for a ref can be obtained from the `head` field in the `refs` table. Would that simplify this query?

packages/backend/src/document.rs
```diff
@@ -123,3 +125,192 @@ pub struct RefContent {
...
+        LEFT JOIN users owner ON owner.id = (
+            SELECT p_owner.subject 
+            FROM permissions p_owner 
+            WHERE p_owner.object = ls.ref_id 
+            AND p_owner.level = 'own' 
+            LIMIT 1
+        )
+        WHERE (
+            (p_searcher.subject = $1
+            AND p_searcher.level IN ('read', 'write', 'maintain', 'own'))
```
epatters: I think this line can be replaced with `p_searcher.level IS NOT NULL`

packages/backend/src/document.rs
```diff
@@ -123,3 +125,192 @@ pub struct RefContent {
...
+                display_name: row.owner_display_name,
+            },
+        })
+        .collect();
+
+    Ok(stubs)
+}
+
+/// Gets a list of RefStubs that a user has permissions for and match the search parameters
+pub async fn search_ref_stubs_related_to_user(
```
epatters: This function is similar enough to the previous one that:

- I can't easily tell from the docstrings how they are different (but the difference *was* clear from the PR comments, so thanks for explaining it there!)

- I wonder whether the two functions should be consolidated into a single function, say by adding an extra field `individual_level: Option<PermissionLevel>` to `RefQueryParams` that would restrict the results to documents for which the searcher has *at least* that level (set individually, not inherited from global permissions). Then, if I understand correctly, in one case you pass `Some(Read)` or `Some(Own)` (depending on how strongly the document is supposed to be "associated" with the user) and in the other case, you pass `None`.

epatters (commented):
Thanks Jason, looks like a great start on a much-needed feature. I haven't tried it out in the frontend yet, but I left some questions/comments on the backend piece below.

packages/backend/src/document.rs
jmoggr: > How do we know what the format of this datetime string will be?

Great question! We are casting `TIMESTAMPZ` to text with `::TEXT` however I have been unable to find any docs on that behaviour (!), a quick check shows that the text is NOT ISO 8601. JS Date still accepts the string and works as expected, but this behaviour is not well defined. It turns out I've been doing this wrong for almost a decade...

> On the Rust side

`time` is not supported by `ts-rs`, so I went with `DateTime<Utc>` from `chrono`. This also simplifies the code a non-trivial amount.

packages/backend/src/document.rs
jmoggr: I agree with the second point. I had accidentally left a comment unsent yesterday with a similar sentiment:

> The queries for both `search_ref_stubs` and `search_ref_stubs_related_to_user` are fairly similar and fairly complicated. Having worked with this code a bit, I now believe that the overall maintenance burden would be reduced if these queries were combined. (I made cut+paste errors a couple of times, which bodes ill).

Consolidating the queries will also address the first point.

jmoggr:
> I haven't tried it out in the frontend yet

note: the 'My documents' option in the hamburger menu is only available if you are logged in, and it does not seem like you can log in on the netfily previews.

packages/backend/src/document.rs
epatters: Thanks for looking into it. At least using `chrono` should help keep the Rust code clean even if we ultimately have to coerce to a string at the JSON interface.

packages/backend/migrations/20250409171833_add_permissions_object_subject_idx.up.sql
```diff
@@ -0,0 +1,5 @@
+-- no-transaction
```
epatters, line 1: FYI, @jmoggr, I had to add this to fix the migrations:

```sh
~/Local/topos/catcolab/packages/backend> sqlx database reset
Drop database at postgres://epatters:catdev@localhost:5433/catcolab? (y/n): y
Applied 20241004010448/migrate document refs (34.13997ms)
Applied 20241025030906/migrate users (56.077989ms)
error: while executing migration 20250409171833: error returned from database: CREATE INDEX CONCURRENTLY cannot run inside a transaction block
```

packages/backend/src/document.rs
```diff
@@ -123,3 +126,116 @@ pub struct RefContent {
...
 }
+
+/// A subset of user relevant information about a ref. Used for showing
+/// users information on a variety of refs without having to load whole
+/// refs.
+#[derive(Clone, Debug, Serialize, Deserialize, TS)]
+pub struct RefStub {
+    pub name: String,
+    pub type_name: String,
+    pub ref_id: Uuid,
```
epatters, line 139: Though it requires a little extra boilerplate, we should rename the fields in `RefStub` and `RefQueryParams` to match TypeScript case conventions where appropriate. E.g., this field should become:

```rust
#[serde(rename = "refId")]
pub ref_id: Uuid,
```

packages/frontend/src/user/documents.tsx
```diff
@@ -0,0 +1,157 @@
...
+            </td>
+        </tr>
+    );
+}
+
+function getUrlForRefStub(refStub: RefStub): string {
+    const hostname = window.location.hostname;
+    const protocol = window.location.protocol;
+    const port = window.location.port;
+    return `${protocol}//${hostname}:${port}/${refStub.type_name}/${refStub.ref_id}`;
```
epatters: I'm surprised we have to do this, namely construct the full URL here and then run `window.location.href = url` above. Can we use [`useNavigate`](https://docs.solidjs.com/solid-router/reference/primitives/use-navigate), something like:

```typescript
const navigate = useNavigate();

[...]

navigate(`/${refStub.type_name}/${refStub.ref_id}`);
```

epatters (changes requested):
Thanks, this is looking great!

Two small comments to address and then this should be ready to merge. I also made a few small changes myself.

packages/frontend/src/user/documents.tsx
jmoggr: Yup, I changed it to this. I initially was using a `<a href=...>` element and it grew out poorly.

packages/backend/src/document.rs
jmoggr: done!

packages/backend/migrations/20250409171833_add_permissions_object_subject_idx.up.sql
jmoggr: thanks. I had only tested it by running the query directly. Lesson learned.

epatters (approved):

PR #473, opened 2025-04-16
Refactor frontend-core interface to process notebooks in Rust

olynch:
This is a barebones prototype to demonstrate some of the ideas from https://forest.topos.site/ocl-00E3.xml.


packages/catlog-wasm/src/model_old.rs
epatters, line 1: Accidentally committed file.

packages/catlog-wasm/Cargo.toml
```diff
@@ -18,8 +18,9 @@ derive_more = { version = "1", features = ["from", "try_into"] }
 getrandom = { version = "0.2", features = ["js"] }
 js-sys = "0.3.77"
 nonempty = { version = "0.10", features = ["serialize"] }
+notebook-types = { version = "0.1.0", path = "../notebook-types" }
 serde = { version = "1", features = ["derive"] }
-tsify-next = { version = "0.5.4", features = ["js"] }
+tsify-next = { version = "0.5.6", features = ["js"] }
```
epatters: There are merge conflicts because `tsify-next` is back to being `tsify`, see #495.

packages/notebook-types/src/v0/model_judgment.rs
epatters, line 1: To avoid having to de-structure object/model declarations "too early" in the Wasm bindings (leading to unnecessary verbosity), perhaps `ObjectDecl` and `MorphismDecl` should become independent structs and then `ModelDecl` becomes:

```rust
#[serde(tag = "tag")]
#[tsify(into_wasm_abi, from_wasm_abi, missing_as_null)]
pub enum ModelDecl {
    #[serde(rename = "object")]
    Object(ObjectDecl),
    #[serde(rename = "object")]
    Morphism(MorphismDecl)
}
```

You're already doing something similar for `Document`s.

packages/catlog-wasm/src/model.rs
```diff
@@ -1,123 +1,158 @@
...
+#[wasm_bindgen]
+pub struct DblModel(#[wasm_bindgen(skip)] pub DblModelBox);
+
+/** Elaboration is the process of transforming notation (as declared in
+notebook-types) into syntax and values. This can possibly fail. Eventually,
+this struct may have some role to play in accumulating errors, but for now it is
+a singleton. */
+pub struct Elaborator;
+
+pub trait CanElaborate<T, S> {
```
epatters, line 48: I assume the traits for elaboration and quoting are intended to ultimately live somewhere else?

packages/frontend/src/model/document.ts
```diff
@@ -140,7 +139,7 @@ export async function createModel(
         init = initOrTheoryId;
     }
 
-    const result = await api.rpc.new_ref.mutate(init as JsonValue);
+    const result = await api.rpc.new_ref.mutate(init as any as JsonValue);
```
epatters: The intermediate cast to `any` is unfortunate. Can TS not infer that the new `Document` type is coercible to `JsonValue`?

packages/notebook-types/src/v0/notebook.rs
epatters, line 1: Can the type defs in `frontend/src/notebook/types.ts` be replaced with imports from here? So far that file hasn't been touched.

epatters (commented):
Thanks Owen, this is really nice! I don't have any comments about the high-level design. We've already discussed it a fair amount, and I think it's a big improvement. I especially like seeing the code eliminated from the frontend type defs.

In addition to the small-ish comments below, before merging some general cleanup will be needed, e.g. there's at least one block of old commented-out tests. You can take a pass at cleanup and then I'll do a final review for that stuff.

packages/catlog-wasm/src/model.rs
olynch: So, this is a bit tricky because of the rules for traits. Because both the types in `notebook-types` and the types in `catlog` have already been defined, this trait has to be defined in the same crate as its instances for these previously-defined types are declared. I'm not sure if it has to literally be in the same file though; I'll look into this. But I think eventually I'd like to put the elaboration and quoting in `catlog` itself.

packages/notebook-types/src/v0/notebook.rs
olynch: Ah yes, didn't see these!

packages/frontend/src/model/document.ts
olynch: Apparently not; I'm not sure why.

packages/notebook-types/src/v0/model_judgment.rs
olynch: For backwards compatibility, it makes more sense to have `ObjectDecl = ModelDecl & { type: "object" }` in the typescript. But once we are refactoring this later on, we can change that.

packages/catlog-wasm/src/model.rs
epatters: Ah yes, you cannot define a foreign trait on a foreign type, a limitation that is annoying but understandable. We can ignore this for now. (It does not have be in the same file, just in the same crate.)

packages/frontend/src/model/document.ts
epatters: The answer seems to be the [difference between types and interfaces](https://stackoverflow.com/a/71394297) in TypeScript: previously, these type defs were `type`s, now they're `interface`s because that's what tsify generates (see https://github.com/madonoharu/tsify/issues/61). I'm looking for a workaround that doesn't require short-circuiting the TS compiler.

epatters: OK, I added a workaround for this in d8def43.

packages/notebook-types/src/v0/model_judgment.rs
epatters: I did this in a backwards-compatible way in f13b0569d44cbdb9eeb1f3b6c449020bfed14046.

epatters (approved):

PR #479, opened 2025-04-22
Support for Darwin devShell in flake.nix

ydewit:
The existing flake.nix was not working for Darwin.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -537,24 +537,6 @@
         "type": "github"
       }
     },
-    "flake-utils_6": {
-      "inputs": {
-        "systems": "systems_8"
-      },
-      "locked": {
-        "lastModified": 1731533236,
-        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
-        "owner": "numtide",
-        "repo": "flake-utils",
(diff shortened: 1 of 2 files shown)


epatters:
Hi Yuri, and thanks for this improvement! @jmoggr, would you mind giving this a look?

jmoggr (approved):
Looks good! This doesn't break the deploy or my own env.

A note on `flake-utils`:
I don't love the use of `flake-utils`, it's a dependency to replace what otherwise would be a 5 line function. I avoid it in personal projects, but I think it's acceptable for us because it's so widely used.

jmoggr:
> I am not an expert on Nix. Let me see how can I get rid of flake-utils

I think the function would look something like this:

```
    # (system:string -> attrs) -> { system = attrs; … }
    eachDefaultSystem = systems: f:
      builtins.listToAttrs (map (system: {
        name = system;
        value = f system;
      }) systems);

    defaultSystems = [ "x86_64-linux" "aarch64-darwin" ];

    perSystemOutputs = eachDefaultSystem defaultSystems (system: let
```

I'm happy to take this on if you like or otherwise get stuck. Please don't let this eat much of your time!

jmoggr (approved):
Thank you!

I did a quick test with the new lockfile and there were no issues.

epatters:
Thanks both, looks great!

PR #486, opened 2025-04-24
Filter logs using level from RUST_LOG is available

### LINKED ISSUE #484: Library traces are not being printed despite the tracing level being set to INFO
in main.rs we set the tracing level like `tracing_subscriber::fmt().with_max_level(tracing::Level::INFO).init();`

However `debug!` and statements from libraries like `firebase-auth` are not being printed.

jmoggr:
closes #484 

This allows changing the log level in development by running the backend with `RUST_LOG=debug cargo run`, which should make debugging marginally easier in the future. We can also set the `RUST_LOG` variable in the systemd service if we want more robust logs on the servers.


epatters (approved):
Thanks, this should be useful for debugging the backend!

PR #488, opened 2025-04-25
Use nix to install fswatch and the full TeXLive distribution

ydewit:
Added a shell.nix to dev-docs/ folder to install the full TeXLive distribution and fswatch.


epatters (approved):
Thanks Yuri!

PR #494, opened 2025-05-05
Stock-flow visualization: Larger boxes, ortho flows, and blue links

eRedekopp:
Minor changes to Stock & Flow diagram visualizations. Make the flow arrows ortho-style, make the minimum stock size a little bigger, and make link arrows blue. 

**Note**: the change that I made to the CSS assumes that the class name "link" is only ever used for Stock & Flow diagram links, and so all "link"s will be blue. As far as I can tell from looking through the code, that is currently the case. However, I can make a new class for Stock & Flow links specifically if you'd prefer. 

**Before:** 
image

**After:**
image

### DIFF
--- a/packages/frontend/src/stdlib/analyses/graph_visualization.css
+++ b/packages/frontend/src/stdlib/analyses/graph_visualization.css
@@ -1,3 +1,7 @@
 .graph-visualization {
     overflow-x: auto;
 }
+
+g.link path {
+    stroke: blue;
+}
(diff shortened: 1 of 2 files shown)


epatters (approved):
Thanks Eric, this indeed looks much better!

In the future, to prevent CI failures, you should run the formatter and linting before pushing:

```sh
pnpm run format
pnpm run lint
```

I went ahead and did that here.

eRedekopp:
Thanks! Sorry about the linting issue -- I actually did run it, but I got a warning related to a temp file that the linter didn't like (and wasn't uploaded as part of this PR), then it said "no fixes made" or similar. I mistakenly thought that it meant that I had a warning but that it completed successfully otherwise, but apparently it meant I had a warning and it didn't run at all.

Issue #356, opened 2025-01-30
Don't remove permissions button in analysis mode


KevinDCarlson:
See title. Could combine this with possibility of closing out of analysis pane.

KevinDCarlson:
Note that there’s no way to modify permissions for analyses at all right now, so this is a high priority to fix. There are also surprising behaviors with permissioned docs blocking analysis access etc.

tim-at-topos:
yes this is a very frustrating issue (i've just been struggling with it)

epatters:
It is frustrating but should be fixed during or shortly after #562, which puts analysis documents on the same footing as other types of documents.

tim-at-topos:
is this now fixed? if so, I think I have a new bug: when an analysis is open in the right-hand pane, the permissions button seems to actually control permissions for the model (on the left-hand pane). If anybody else confirms this then I'll open an issue (@kasbah ?)

kasbah:
@tim-at-topos That was true but now fixed with #932. I agree the original issue was also fixed along the way somewhere so closing this.

PR #498, opened 2025-05-08
Linked to an analysis instead of a model

### LINKED ISSUE #497: Lotka-Volterra model permissions
When I go to

https://catcolab.org/help/theory/reg-net

and click on the Lotka-Volterra network example, I get a "permissions error" message saying I can't read that page.

KevinDCarlson:
Just a quick fix to close #497. I've tested I can view it while logged out!


epatters (approved):
Thanks Kevin! We can avoid these problems once we have #356.

PR #508, opened 2025-05-17
Use rust-toolchain.toml as source of truth for rust version

jmoggr:
Use the `rust-toolchain.toml` file ([docs](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file)) as the source of truth for the rust version to use in Nix and CI.

This also allows us to get of the `unstable` branch for NixOS. Unfortunately the change in the NixOS version means that a manual deploy is required since the github runner will likely timeout.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,15 +2,17 @@
   description = "configurations for deploying catcolab";
 
   inputs = {
-    # The version of cargo in 24.11 is too old so we need to use unstable until the next relase (25.05)
-    nixpkgs.url = "nixpkgs/nixos-unstable";
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-24.11";
     crate2nix = {
       url = "github:nix-community/crate2nix";
       inputs.nixpkgs.follows = "nixpkgs";
     };
-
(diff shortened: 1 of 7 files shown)


flake.nix
```diff
@@ -32,14 +35,25 @@
         "aarch64-darwin"
       ];
 
+      rustToolchain = fenix.packages.x86_64-linux.fromToolchainFile {
+        file = ./rust-toolchain.toml;
+        sha256 = "sha256-X/4ZBHO3iW0fOenQ3foEvscgAPJYl2abspaBThDOukI=";
+      };
+
+      nixpkgsFor =
```
jmoggr, line 43: note: currently this abstraction is not needed, however it's good to have a single source of truth for the `pkgs` used by all of the outputs. Also there is a decent chance we will need to this in the future.

(I added this to support global overlays to `pkgs`, however I didn't use overlays in the end)

epatters (approved):
Thanks Jason, this is great! It's a nice touch that we can use a standard GitHub action for Rust instead of rolling it ourselves.

PR #510, opened 2025-05-19
Use sqlx_migrator for migrations

jmoggr:
sqlx_migrator is pretty great. I've had no issue with it (85% of the time was spent on rust/nix), it neatly solves the immediate migration coordination problem and it's powerful enough that several future headaches can be avoided with it.

I tried to do this without creating another package and ran into a number of problems:
- The migrator can't be built with/inside the `backend` binary, since this could cause circular dependencies on `sqlx` types (can't compile because the query doesn't match what's in the DB, can't update the DB to match because we can't compile).
- I couldn't figure out how to package a crate with multiple binaries with nix.

Other than that, everything works! This has already been deployed and tested on staging (for developing the nix stuff).

I have not yet updated the docs.


infrastructure/modules/backend.nix
```diff
@@ -33,15 +33,45 @@ let
...
-    # the migrations directory is copied into the output of backend
-    cd ${catcolabPackages.backend}
-    ${lib.getExe pkgs.sqlx-cli} migrate run
+    # Transition migrations from sqlx to sqlx_migrator
+    # Check if we have run sqlx_migrator before, if we have not, mark all migrations as having been applied
+    # If sqlx_migrator has not been run before then the _sqlx_migrations table will exists, but not the
+    # _sqlx_migrator_migrations table.
+    #
+    # We can get rid of this block after all DBs have been updated with sqlx_migrator, new DBs will never
+    # have have had _sqlx_migrations so there will be no need to transition them.
```
epatters, line 42: Makes sense! Just flagging this so we can create an issue that will remind us to remove this shim once both prod and next have had their migrations migrated (!).

epatters (approved):
Thanks Jason, I like this! I'm marking this as approved since the code itself looks great, but can you also:

- Update docs as appropriate (mentioned in your comment)
- Create issue to remove shim (see my comment below)

infrastructure/modules/backend.nix
jmoggr: New issue is here: https://github.com/ToposInstitute/CatColab/issues/514

jmoggr:
I rebased onto `main`, and added a README for the new package. There are almost no references to migrations outside of the Tex developer documentation. That documentation is badly out of date since we have nixified everything and migrations are not really relevant in that context (they're now handled entirely by Nix).

I've created an issue for updating those docs here: https://github.com/ToposInstitute/CatColab/issues/514

epatters (approved):

PR #516, opened 2025-05-21
Add `db-utils` script for database development

jmoggr:
Adds a utility script to make developing against a local database suck less by making it easy to save and restore snapshots of any CatColab DB.

My most recent usecase for developing migrations looked like:
1. `db-utils dump --from staging`
2. `db-utils load --from staging --to local --skip-migrate`
3. develop migrations manually, running above command again to get a clean state
4. `db-utils load --from staging --to local` to test the migrations script

This also makes it trivial to develop/test against the staging or prod DB, and a bunch of other usecases. It also codifies the process for setting up the DB on a new development machine.

It would be understandable if you didn't want to accept this PR: this is a relatively large amount of bash and the rest of the world seems to manage just fine without this script (which I find confusing). Other people have found it useful in the past and I'll keep using it regardless (unless prohibited).

aside: do we have a process for restoring from our DB backups? Has that been tested?

fun fact: I've carried some version of this script with me through every job (12 years!).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,5 +91,8 @@ instrument.mjs
 # direnv cache
 .direnv
 
-# bild build output
+# nix build output
 /infrastructure/result
+
+# Database dumps
+/dumps
--- a/flake.nix
+++ b/flake.nix
@@ -104,6 +104,9 @@
               else
                 ""
             }
+
+            export PATH=$PWD/infrastructure/scripts:$PATH
+
             # Load DATABASE_URL into the environment
             if [ -f packages/backend/.env ]; then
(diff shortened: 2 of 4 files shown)


epatters (approved):
Thanks Jason! This will be helpful.

> It would be understandable if you didn't want to accept this PR: this is a relatively large amount of bash and the rest of the world seems to manage just fine without this script (which I find confusing). 

I'm basically incapable of writing bash but I'm happy to accept this because even if it were to become unmaintainable, removing it would only restore the current status quo of having no script at all.

> fun fact: I've carried some version of this script with me through every job (12 years!).

Ah, that explains how you were able to just pull this out of a hat :)

PR #521, opened 2025-05-22
Linear ODE analysis for CLDs

tim-at-topos:
Adding a new analysis for causal loop diagrams, where we interpret an arrow $A\to^{\pm}B$ as representing the equation $\dot{B}=\pm kA$ for some constant $k$ (so the user is asked to input initial values for $A$ and $B$, and the coefficient $k$); multiple arrows coming in correspond to addition of equations, so we interpret $A\to^{+}B\leftarrow^{-}C$ as $\dot{B}=kA-\ell B$.


tim-at-topos:
Should I add some documentation about this analysis to `help/theory/causal-loop.mdx`? That seems like a sensible place to explain the semantics (and I could also add something for Lotka–Volterra while I'm there too).

packages/catlog/src/simulate/ode/cclfo.rs
```diff
@@ -0,0 +1,33 @@
+//! Constant-coefficient linear first-order (CCLFO) differential equations.
```
epatters: While I appreciate the specificity, the acronym "CCFLO" is non-standard and so I think we should avoid using it as our primary name throughout the API. I will suggest "linear ODE", which doesn't imply that the coefficients are constant but should be clear enough to anyone with applied math background.

packages/catlog/src/simulate/ode/cclfo.rs
```diff
@@ -0,0 +1,33 @@
...
+use super::ODEProblem;
+use super::ODESystem;
+
+/** A CCLFO dynamical system.
+
+A system of linear first-order ODEs with constant coefficient; a semantics for
+enriched causal loop diagrams.
+*/
+#[derive(Clone, Debug, PartialEq)]
+pub struct CCLFOSystem {
```
epatters: So, I would call this struct `LinearODESystem`.

packages/catlog/src/simulate/ode/cclfo.rs
```diff
@@ -0,0 +1,33 @@
...
+    interaction_coeffs: DMatrix<f32>,
```
epatters: Since there's only one type of coefficient/parameter in this case, I'd just call this `coeffs` or `coefficients`.

packages/catlog/src/stdlib/analyses/ode/cclfo.rs
```diff
@@ -0,0 +1,147 @@
...
+#[cfg_attr(
+    feature = "serde-wasm",
+    tsify(into_wasm_abi, from_wasm_abi, hashmap_as_object)
+)]
+pub struct CCLFOProblemData<Id>
+where
+    Id: Eq + Hash,
+{
+    /// Map from morphism IDs to interaction coefficients (nonnegative reals).
+    #[cfg_attr(feature = "serde", serde(rename = "interactionCoefficients"))]
```
epatters: Similar comment: we can just have `rename = "coefficients"` here.

packages/catlog/src/stdlib/analyses/ode/cclfo.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// Duration of simulation.
+    duration: f32,
+}
+
+type Model<Id> = DiscreteDblModel<Id, UstrFpCategory>;
+
+/** CCLFO ODE analysis for models of a double theory.
+
+The main situation we have in mind is ...
```
epatters: Incomplete docstring.

packages/catlog/src/stdlib/analyses/ode/cclfo.rs
```diff
@@ -0,0 +1,147 @@
...
+            .map(|ob| data.initial_values.get(ob).copied().unwrap_or_default());
+        let x0 = DVector::from_iterator(n, initial_values);
+
+        let system = CCLFOSystem::new(A);
+        let problem = ODEProblem::new(system, x0).end_time(data.duration);
+        ODEAnalysis::new(problem, ob_index)
+    }
+}
+
+// #[cfg(test)]
```
epatters: Commented test: ideally we would have a simple test here, but in any case we should avoid commented-out dead code.

packages/frontend/src/stdlib/theories.tsx
```diff
@@ -353,6 +353,9 @@ stdTheories.add(
                         return thSignedCategory.positiveLoops(model, options);
                     },
                 }),
+                analyses.configureCCLFO({
+                    simulate: (model, data) => thSignedCategory.cclfo(model, data),
+                }),
```
epatters, line 361: Might as well add this analysis to reg nets as well, which have an isomorphic theory.

epatters (changes requested):
Thanks Tim, this is a great PR!

- Most of my comments are just bike-shedding the naming
- Something should be done about the commented test code, ideally adding a working test

epatters:
> Should I add some documentation about this analysis to `help/theory/causal-loop.mdx`? That seems like a sensible place to explain the semantics (and I could also add something for Lotka–Volterra while I'm there too).

Yes, this would be a great addition! Cf. my comments in #541

tim-at-topos:
I've written some documentation for this analysis (and Lotka–Volterra) in `causal-loop.mdx`, but in trying to get KaTeX rendering working (following https://mdxjs.com/guides/math/) I seem to have broken something: the help page throws an error (`TypeError: Comp is not a function`) as soon as I actually include any maths.

tim-at-topos:
OK, I've finally updated the help page for the causal loop diagram theory. Please do take a look and see if it's helpful/if my understanding of the Lotka–Volterra semantics are correct! If so, then I'm happy for it to be merged :-)

Thanks again @jmoggr for your help with the KaTeX rendering!

Edit: actually, I still don't have any tests — let me write one "quickly" and then I'll nudge you again

packages/catlog/src/stdlib/analyses/ode/cclfo.rs
tim-at-topos: two tests!

tim-at-topos:
ok @epatters , hopefully the final time I'll tag you here, but I think I've done docs + tests to your standards :-)

epatters (approved):
Thanks Tim, docs and tests look great. Especially happy to get our first long-form docs for analyses.

PR #522, opened 2025-05-22
Add `--clean` argument to `pg_dump` for database backups

jmoggr:
This is stacked on top of https://github.com/ToposInstitute/CatColab/pull/519/

This change allows the DB to be restored by running `psql < dump.sql` as the `catcolab` user, previously the database would have to be dropped and recreated by a postgres admin. The new flags do not break any behaviour that would have been possible previously.

It's still possible to restore from the previously created backups, it's just more of a pain.

### DIFF
--- a/infrastructure/modules/backup.nix
+++ b/infrastructure/modules/backup.nix
@@ -13,7 +13,7 @@ let
 
     cd ~
 
-    "${pkgs.postgresql}/bin/pg_dump" catcolab > $DUMPFILE
+    ${pkgs.postgresql}/bin/pg_dump --clean --if-exists > $DUMPFILE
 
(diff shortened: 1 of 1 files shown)


infrastructure/modules/backup.nix
```diff
@@ -45,7 +45,6 @@ with lib;
         User = "catcolab";
         ExecStart = getExe backupScript;
         Type = "oneshot";
-        EnvironmentFile = config.age.secrets.backendSecretsForCatcolab.path;
```
jmoggr, line 48: Since we're running this as the `catcolab` user we don't need access to any DB credentials, the linux `catcolab` user is already set up to log in as the postgres `catcolab` user.

epatters (approved):
Thanks Jason for exercising this process and making sure it works smoothly!

PR #524, opened 2025-05-22
Increase limit on number of search results returned to 1000

### LINKED ISSUE #523: My Documents cuts off after oldest n documents
The My Documents page shows maybe 100 results at a time, without any ability to go to the next page. It's sorted in what's probably usually the less-useful direction, from oldest to newest, with no way yet to re-sort, so the only way you can get your documents past the cutoff is by search.

jmoggr:
Closes #523


KevinDCarlson:
Thanks @jmoggr. Everything is broken due to sqlx changes though, I think?

KevinDCarlson:
While we're on quick fixed, can we invert the default sort to run newest->oldest?

KevinDCarlson (approved):
This looks correct but I can't actually see the change locally, which is a bit annoying.

PR #525, opened 2025-05-28
Paginate document search results

jmoggr:
Add pagination to document search results.

This PR is meant to be somewhat didactic:
- Type narrowing in JSX blocks is not something that we've really done anywhere yet, but it is very useful and not very intuitive in SolidJS.
- Slightly more idiomatic usage of createResource

Down the road it would be good to replace this with some sort of infinite scroll/incremental loader component.

### DIFF
--- a/packages/frontend/src/user/documents.tsx
+++ b/packages/frontend/src/user/documents.tsx
@@ -123,41 +123,18 @@ function DocumentsSearch() {
                                     </tr>
                                 </Match>
                                 <Match when={rpcResourceOk(pageData)}>
-                                    {(pageData) => (
-                                        <>
-                                            <For each={pageData().content.items}>
-                                                {(stub) => <RefStubRow stub={stub} />}
-                                            </For>
-
-                                            <tr class="pagination-row">
-                                                <td colspan="5" style={{ "text-align": "center" }}>
-                                                    <button
-                                                        disabled={page() === 0}
-                                                        onClick={() => setPage(page() - 1)}
-                                                    >
-                                                        Previous
-                                                    </button>
-
-                                                    <span class="page-info">
-                                                        Page {page() + 1} of{" "}
-                                                        {Math.ceil(
-                                                            pageData().content.total / pageSize,
-                                                        ) || 1}
-                                                    </span>
-
-                                                    <button
-                                                        disabled={
-                                                            (page() + 1) * pageSize >=
-                                                            pageData().content.total
-                                                        }
-                                                        onClick={() => setPage(page() + 1)}
-                                                    >
-                                                        Next
-                                                    </button>
-                                                </td>
-                                            </tr>
-                                        </>
-                                    )}
+                                    {(res) => {
+                                        const { items, total } = res().content;
+                                        return (
+                                            <DocumentRowsPagination
+                                                items={items}
+                                                total={total}
+                                                page={page()}
+                                                setPage={setPage}
+                                                pageSize={pageSize}
+                                            />
+                                        );
+                                    }}
                                 </Match>
                             </Switch>
                         </tbody>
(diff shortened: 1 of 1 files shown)


packages/frontend/src/user/documents.tsx
```diff
@@ -26,32 +26,32 @@ function DocumentsSearch() {
...
+            });
 
-        const result = await api.rpc.search_ref_stubs.query({
-            ownerUsernameQuery: null,
-            refNameQuery: query,
-            includePublicDocuments: false,
-            searcherMinLevel: null,
-        });
-
-        if (latestRequestId() !== requestId) {
```
jmoggr, line 49: createResource automatically cancels the getter when necessary, making this redundant.

packages/frontend/src/user/documents.tsx
```diff
@@ -89,26 +89,73 @@ function DocumentsSearch() {
...
                             >
-                                <Match when={resultOk(refStubs())}>
-                                    {(okRes) => (
-                                        <For each={okRes()}>
-                                            {(stub) => <RefStubRow stub={stub} />}
-                                        </For>
-                                    )}
+                                <Match when={pageData.loading}>
+                                    <tr>
+                                        <td colspan="5">Loading...</td>
```
jmoggr: I know we're not a fan of flashing a loading state, but I think it's warranted here: this is an interactive component and it would be good to provide feedback to the user when they are typing. Otherwise the stale results will be shown until the backend returns, leaving the user unable to know when/if the search completed in cases where the same results could conceivably be returned.

packages/frontend/src/user/documents.tsx
```diff
@@ -89,26 +89,73 @@ function DocumentsSearch() {
...
+                                <Match when={pageData.state === "errored"}>
+                                    <tr>
+                                        <td colspan="5">
+                                            Error caught by fetcher:{" "}
+                                            {JSON.stringify(pageData.error, null, 2)}
+                                        </td>
+                                    </tr>
+                                </Match>
+                                <Match when={rpcResourceOk(pageData)}>
+                                    {(pageData) => (
```
jmoggr: This `pageData` variable is technically shadowing the `pageData` in the components scope, however:
- It has the same type
- Both types are read only and would have the same data, so I don't think shadowing could ever cause a problem
- I couldn't think of a better name, or one that communicates anything useful

packages/frontend/src/user/documents.tsx
KevinDCarlson: There are also intermediate options: gray out the stale results while typing and flash fresh results briefly when they populate in, or just blank out stale results without the "loading..." I don't particularly mind the solution here, although I agree that just leaving stale results with no feedback on when they're refreshed doesn't seem right in this case.

jmoggr: > gray out the stale results while typing

This is non-trivial, it represents more fiddly state management. We can still do it, but I'm just pointing this out.

I added a spinner, which might satisfy this.

packages/frontend/src/user/documents.tsx
```diff
@@ -119,6 +169,7 @@ function DocumentsSearch() {
 }
 
 export function RefStubRow(props: { stub: RefStub }) {
+    console.log(JSON.stringify(props.stub, null, 2));
```
epatters: Lingering log statement.

packages/frontend/src/user/documents.tsx
epatters: That seems reasonable. On a different note, the high level of nesting/frequent line breaking below makes me wonder whether everything below this point in the tree should be factored into its own component. Logically it's unnecessary but it might make the code more pleasant to read.

packages/frontend/src/api/rpc.ts
```diff
@@ -41,16 +42,51 @@ export function createRpcClient(serverUrl: string, firebaseApp?: FirebaseApp) {
...
+//
+// example usage:
+// type Result = { tag: "Ok", content: string } | { tag: "Err", code: number };
+// const [result, setResult] = createSignal<Result>({tag: "Err", code: 0 });
+// ...
+// <Match when={narrow(result, (res) => res.tag === "Ok")}>
+//     {(result) => {
+//         const a = result(); // the type of a is narrowed to { tag: "Ok", content: string }
+// }
+export function narrow<A, B extends A>(accessor: Accessor<A>, guard: (v: A) => v is B): B | null {
```
epatters, line 55: Am I correct that this function is included for pedagogy and not intended to be used directly? If so, while I appreciate the explanation, I don't think we should leave it in the codebase on the general principle that we shouldn't have "dead code."

The docstring is useful though: I'd suggest adding an example usage in this style to `rpcResourceOk` and/or `rpcResourceErr`.

jmoggr: > Am I correct that this function is included for pedagogy and not intended to be used directly?

Not entirely, it can be used directly and I was doing so initially. I created the specialised `rpcResourceOk` and `rpcResourceErr` functions for readability and possible future reuse.

It's not immediately obvious how to write the more general `narrow` function if given only the specialised cases, and it took a non-trivial amount of frustration to figure out what was wrong and to find a solution. 

I propose that the reduction in possible future frustration is worth a couple of dead lines of code. However it's not obvious to me if the same trade off is worth it if it means a breaking a previously well enforced rule.

packages/frontend/src/user/documents.tsx
jmoggr: Stacked PR with the change is here: https://github.com/ToposInstitute/CatColab/compare/search-pagination...documents-pagination-component?expand=1

I'm not really sure which is better.

packages/frontend/src/api/rpc.ts
epatters: Ok, since `narrow` is itself usable/has been used before, I'm happy to leave it in.

epatters (approved):

PR #526, opened 2025-05-28
Add prosemirror KaTex plugin and menu bar for the rich text editor

jmoggr:


packages/frontend/src/components/rich_text_editor/rich_text_editor.tsx
```diff
@@ -0,0 +1,360 @@
...
+        });
+
+        if (props.ref) {
+            props.ref(view);
+        }
+
+        setHeadingLevel(activeHeading(view.state, schema));
+        setMarkStates(activeMarks(view.state, schema));
+
+        setMenuControls({
```
jmoggr, line 199: This method for passing callbacks to the menu is relatively weird, however I've seen a couple examples of it and this may be the least messy way of dealing with the asynchronous loading of the document.

packages/frontend/src/components/rich_text_editor/commands.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        dispatch(tr);
+
+        return true;
+    };
+}
+
+// Currently does not work due to bug in the automerge-prosemirror plugin. (It also might not work in
+// general, but it theoretically should)
+export function insertMathInlineCmd(mathNodeType: NodeType, initialText = ""): Command {
```
epatters: Inline math is at least as important as display math. How do we debug this problem? Is the next step to ask the Automerge devs for help?

Also, if I'm not mistaken, this function comes from here: https://github.com/benrbray/prosemirror-math/blob/master/lib/commands/insert-math-cmd.ts

We should include a link to the upstream definition.

packages/frontend/src/components/rich_text_editor/rich_text_editor.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+function richTextEditorKeymap(schema: CustomSchema, props: RichTextEditorOptions) {
+    const bindings: { [key: string]: Command } = {};
+
+    bindings["Mod-b"] = toggleMark(schema.marks.strong);
+    bindings["Mod-i"] = toggleMark(schema.marks.em);
+    bindings["Mod-m"] = insertMathDisplayCmd(schema.nodes.math_display);
+
+    if (props.deleteBackward) {
+        bindings["Backspace"] = doIfEmpty(props.deleteBackward);
```
epatters: Once we have inline math we probably want to add `mathBackspaceCmd` here: https://github.com/benrbray/prosemirror-math?tab=readme-ov-file#plugins

packages/frontend/src/components/rich_text_editor/rich_text_editor.tsx
```diff
@@ -0,0 +1,361 @@
...
+                    <option value={6}>Heading 6</option>
+                </select>
+            </Show>
+
+            <Show when={props.onBlockQuoteClicked}>
+                <TooltipButton tooltip="Blockquote" onClick={() => props.onBlockQuoteClicked?.()}>
+                    <TextQuote />
+                </TooltipButton>
+            </Show>
+            <Show when={props.onToggleNumberedList}>
```
epatters: Are list supposed to work yet? I didn't see buttons for them in the toolbar.

epatters (changes requested):
Thanks Jason! This is a great start. I'm super excited to have math support.

**Bugs**:

- When I select text and toggle bold, the selection also changes. The selection should remain stable.

**Cosmetic**:

- Toolbar buttons seem a little big
- Toolbar should have some background and border to make it stand out from user content

packages/frontend/src/components/rich_text_editor/rich_text_editor.tsx
jmoggr: added!

packages/frontend/src/components/rich_text_editor/rich_text_editor.tsx
jmoggr: they now work

packages/frontend/src/components/rich_text_editor/commands.ts
jmoggr: Link to the issue for the automerge devs: https://github.com/automerge/automerge-prosemirror/issues/30

I added a link to the upstream definition.

jmoggr:
There are still (at least) two problem with focus moving in and out of math blocks. Neither really breaks anything and I was unable to find a quick fix to any either of them (I spent about an hour on them together).

I'm inclined to create issues for them and move on until we start getting complaints.

KevinDCarlson:
Yes, let's leave the two remaining focus issues for now. Is this all ready as far as you're concerned, Jason, or is there a need to split out more of the inline work or anything else?

epatters (commented):
Thanks Jason, this is looking really good.

I'm aware of a few bugs:

1. The focus issues that you've already noted.
2. This bug that I mentioned above is still present:
   > When I select text and toggle bold, the selection also changes. The selection should remain stable.
   To elaborate, when text is selected and then bold or italicize is pressed, the selection disappears, but continuing to press bold or italicize toggles
3. When I click the button for an unordered list, I get an ordered list. I am not able to create an unordered list.

I'm ready to merge this PR once issues are created to track all of the bugs that remain.

jmoggr:
> 2. This bug that I mentioned above is still present:
>    > When I select text and toggle bold, the selection also changes. The selection should remain stable.

I was only able to repro this on firefox, which means that you're fellow firefox user! (I only use chrome for webdev stuff).

There is a difference in how chrome/firefox process focus changes: firefox does it on mousedown, chrome does it on mouseup. Firefox is technically not following the spec, it has something to do with security concerns.

I refactored a bunch of the menubar code to simplify it, also the menubar tooltips now work.

> When I click the button for an unordered list, I get an ordered list. I am not able to create an unordered list.

Fixed!

jmoggr:
> Yes, let's leave the two remaining focus issues for now. Is this all ready as far as you're concerned, Jason, or is there a need to split out more of the inline work or anything else?

I think this is ready to go. The inline stuff is already separated from this PR (this PR still contains some stubs for it, which I think is fine/useful).

epatters (approved):

packages/frontend/src/components/rich_text_editor/rich_text_editor.tsx
```diff
@@ -0,0 +1,376 @@
...
+                callback={props.onItalicClicked}
+                isActive={props.isEmActive}
+                tooltip="Italics (shortcut: Mod+i)"
+            >
+                <Italic />
+            </TooltipButton>
+            <TooltipButton callback={props.onLinkClicked} tooltip="Add Link">
+                <Link />
+            </TooltipButton>
+            <TooltipButton callback={props.onMathClicked} tooltip="KaTeX block (shortuct: Mod+m)">
```
tim-at-topos, line 310: there's a typo here: `shortuct` instead of `shortcut`

tim-at-topos (commented):
small typo

packages/frontend/src/components/rich_text_editor/rich_text_editor.tsx
tim-at-topos: https://github.com/ToposInstitute/CatColab/pull/566

PR #527, opened 2025-05-30
FIX: postgres collation version mismatch warning

jmoggr:


epatters (commented):
Thanks Jason for fixing this warning. Am I correct that once this migration has run on next and prod, we can/should delete it, because this only happens due to changes to the version of Postgres being deployed?

jmoggr:
This has not been run on next or prod. If we were to run it we should probably NOT delete it. Deleting migrations that have been recorded by the DB is possible, however it is not well supported, is error prone, and IMO should be avoided above most other alternatives.

If you'd like I could just run the SQL directly and we can scrap this migration (which is entirely reasonable).

There are a couple of reasons we might want to keep this as a migration:
- We should not be in the habit of running SQL directly on prod (even though we have been (sorry)).
- This would be relevant if we were ever to restore from backup, and it could be good practice/hygiene to have a log of all operations done to the DB.
- We have not yet had an automatic and error free deployment of a DB migrations across both staging and prod.

At the same time this adds new files/code/legacy for an SQL command that we can functionally just pretend was never run.

I would like to pass this decision to you. Either option is reasonable, this is mostly a cultural issue that I don't yet have enough experience to predict your intentions on.

PR #529, opened 2025-05-30
REFACTOR: Split document search pagination into component for readability

jmoggr:
stacked on top of https://github.com/ToposInstitute/CatColab/pull/525

I dunno which is more readable. I'm struggling to from an opinion given the seemingly low difference in readability and my closeness to this code.

### DIFF
--- a/packages/frontend/src/user/documents.tsx
+++ b/packages/frontend/src/user/documents.tsx
@@ -123,41 +123,18 @@ function DocumentsSearch() {
                                     </tr>
                                 </Match>
                                 <Match when={rpcResourceOk(pageData)}>
-                                    {(pageData) => (
-                                        <>
-                                            <For each={pageData().content.items}>
-                                                {(stub) => <RefStubRow stub={stub} />}
-                                            </For>
-
-                                            <tr class="pagination-row">
-                                                <td colspan="5" style={{ "text-align": "center" }}>
-                                                    <button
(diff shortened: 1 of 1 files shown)


packages/frontend/src/user/documents.tsx
```diff
@@ -168,7 +145,41 @@ function DocumentsSearch() {
...
 
-export function RefStubRow(props: { stub: RefStub }) {
+function DocumentRowsPagination(props: {
+    items: RefStub[];
+    total: number;
+    page: number;
+    setPage: (p: number) => void;
+    pageSize: number;
+}) {
+    const { items, total, page, setPage, pageSize } = props;
```
epatters, line 155: Destructuring props like this is the classic Solid footgun that will break reactivity: https://docs.solidjs.com/concepts/components/props#destructuring-props

jmoggr: Thanks for catching that, it's a habit from React. Interestingly everything still worked as expected, but I have no idea why.

I'm keeping it. Without the component the switch statement code would not easily fit in one screen height, if this was a regular switch statement I would have split it up.

epatters: Such bugs are particularly insidious because they are silent and, in my experience, while things often seem to work fine initially, reactivity failures inevitably appear down the line.

PR #543, opened 2025-06-25
[do not merge] PatchColab

pvh:
A proof-of-concept test of loading CatColab inside of Patchwork. Provides a pair of DataTypes for Models and Analyses, as well as corresponding tools. The two must be manually connected using the Raw editor today by setting the `analysisOf` fields. There are some other broken bits here and there -- need to set some default field names, etc.


packages/frontend/src/App.tsx
```diff
@@ -1,64 +1,37 @@
...
-                return true;
-            }
-
-            return false;
-        },
-        {
-            initialValue: false,
-        },
-    );
+    const api: Api = { repo };
```
pvh, line 34: This was where I "cut in" -- by removing the serverHost and rpc I was able to isolate which parts of the application wouldn't work in Patchwork. Obviously we'll want to discuss if and how to recover those capabilities.

packages/frontend/src/analysis/document.ts
pvh, line 1: Sorry, there's a fair amount of `prettier` reformatting going on here. If I have time before our call I'll clean this up, but this file at least is unchanged aside from an errant debugging console.log.

packages/frontend/src/api/document.ts
```diff
@@ -54,49 +55,37 @@ and upgrade it to the latest version.
...
-    let docHandle: DocHandle<Doc>;
-    if (refDoc.tag === "Live") {
-        const docId = refDoc.docId as DocumentId;
-        docHandle = repo.find(docId) as DocHandle<Doc>;
-    } else {
-        const init = refDoc.content as unknown as Doc;
-        docHandle = localRepo.create(init);
-    }
+    invariant(
+        isValidAutomergeUrl(refId),
```
pvh, line 62: This is the other main change I made, which was making the `refId` the automerge URL directly. We can probably revert this to maintain compatibility between the two builds.

packages/frontend/src/api/document.ts
```diff
@@ -54,49 +55,37 @@ and upgrade it to the latest version.
...
+        () => `Invalid document ref ${refId}`
+    );
+    const { repo } = api;
+
+    let docHandle = await repo.find<Doc>(refId);
```
pvh, line 67: Small note that Automerge Repo 2.x has moved the `await` to repo.find() and .doc() is now synchronous and should never be undefined.

packages/frontend/src/api/document.ts
```diff
@@ -54,49 +55,37 @@ and upgrade it to the latest version.
...
             doc.type === docType,
-            () => `Expected document of type ${docType}, got ${doc.type}`,
+            () => `Expected document of type ${docType}, got ${doc.type}`
         );
     }
 
     const changeDoc = (f: ChangeFn<Doc>) => docHandle.change(f);
 
-    const permissions = refDoc.permissions;
+    const permissions: Permissions = { anyone: "Own", user: "Own", users: [] };
```
pvh, line 79: Faked it here to get things running & satisfy types.

epatters:
Thanks Peter for having gotten this started! I'm going to close this out because the work is continuing elsewhere, but it was helpful to have had this window into it.

PR #546, opened 2025-06-26
Stock and Flow with Auxiliary Variables

quffaro:
This PR introduces an extension of Stock and Flow with _auxiliary variables_, an object which acts as a functions of quantities of stocks connected to it by _variable links_. The auxiliary variable acts on flows by a _flow link_. This effort implements the theory and frontend partially completed in #537, whereas the simulation work will be punted to a new PR.

### DIFF
--- a/packages/catlog/src/stdlib/theories.rs
+++ b/packages/catlog/src/stdlib/theories.rs
@@ -3,7 +3,7 @@
 use ustr::ustr;
 
 use crate::dbl::theory::*;
-use crate::one::{fp_category::UstrFpCategory, Path};
+use crate::one::{Path, fp_category::UstrFpCategory};
 
 /** The empty theory, which has a single model, the empty model.
 


packages/catlog-wasm/src/theories.rs
```diff
@@ -279,6 +279,42 @@ impl ThCategoryLinks {
...
+    }
+
+    #[wasm_bindgen]
+    pub fn theory(&self) -> DblTheory {
+        DblTheory(self.0.clone().into())
+    }
+
+    /// Simulates the mass-action system derived from a model.
+    #[wasm_bindgen(js_name = "mass_action")]
+    pub fn mass_action(
```
epatters: Remove for now until we have an idea about semantics for system structure diagrams with auxiliary variables: the existing mass action semantics will not use the auxiliary variables at all.

packages/catlog/src/stdlib/theories.rs
```diff
@@ -136,6 +136,42 @@ pub fn th_category_links() -> UstrDiscreteTabTheory {
...
+    th.add_mor_type(
+        ustr("Inflow"),
+        TabObType::Basic(x),
+        th.tabulator(th.hom_type(TabObType::Basic(x))),
+    );
+    th.add_mor_type(
+        ustr("Outflow"),
+        TabObType::Basic(x),
+        th.tabulator(th.hom_type(TabObType::Basic(x))),
+    );
```
epatters: As we discussed today, these "inflows" and "outflows" are actually new kinds of links. We decided to punt on inflows/outflows until our theories can be souped up to handle terminal objects.

epatters:
Closing due to staleness. I still think this would be a nice logic to have, even if we're not sure about the quantitative semantics for system structure diagrams with auxiliary variables.

PR #547, opened 2025-06-26
Configure vite to support KaTeX in MDX docs

jmoggr:


packages/frontend/vite.config.ts
```diff
@@ -1,33 +1,21 @@
...
-            jsxImportSource: "solid-js",
-            providerImportSource: "solid-mdx",
-            rehypePlugins: [
-                [rehypeRaw, { passThrough: nodeTypes }],
-                [rehypeKatex, { throwOnError: false }],
-            ],
-            remarkPlugins: [[remarkMath, { singleDollarTextMath: true }]],
-        }),
-        solid({
-            extensions: [".mdx", ".md"],
```
jmoggr, line 29: @epatters 

Removed because it is superfluous:
- we don't have any `.md` file and I'm pretty sure they wouldn't work correctly anyway.
- `.mdx` files are handled by `@mdx-js/rollup`. The docs for `vite-plugin-solid` mention adding `.mdx`, but I can't figure out what the usecase for that actually is.

packages/frontend/vite.config.ts
```diff
@@ -1,33 +1,21 @@
...
-
 export default defineConfig({
     plugins: [
         wasm(),
-        mdx.withImports({})({
-            jsx: true,
-            jsxImportSource: "solid-js",
-            providerImportSource: "solid-mdx",
-            rehypePlugins: [
-                [rehypeRaw, { passThrough: nodeTypes }],
```
jmoggr, line 23: @tim-at-topos Was there any reason for including `rehypeRaw`? I didn't look to me like it did anything.

packages/frontend/vite.config.ts
```diff
@@ -1,33 +1,21 @@
...
-                [rehypeKatex, { throwOnError: false }],
```
jmoggr, line 24: @tim-at-topos

I removed `{ throwOnError: false }` since it would cause KaTeX errors to fail silently. I imagine this was included for debugging.

packages/frontend/vite.config.ts
```diff
@@ -1,33 +1,21 @@
...
-            ],
-            remarkPlugins: [[remarkMath, { singleDollarTextMath: true }]],
```
jmoggr, line 26: @tim-at-topos

`{ singleDollarTextMath: true }` was removed because it's the default value.

packages/frontend/vite.config.ts
```diff
@@ -1,33 +1,21 @@
...
-import rehypeRaw from "rehype-raw";
 import remarkMath from "remark-math";
 import { defineConfig } from "vite";
 import solid from "vite-plugin-solid";
 import wasm from "vite-plugin-wasm";
 
-// @ts-expect-error Types are missing.
-// *Also*, this plugin causes Vite 5 to complain about CJS.
-// https://github.com/nksaraf/vinxi/issues/289
-import pkg from "@vinxi/plugin-mdx";
```
jmoggr, line 12: @epatters 

This package was the problem.

It's API suggests that it supports the same options as `@mdx-js/mdx`, but that didn't work in practice. Vinxi looks like it's (poorly) rolling it's own bundler, so it doesn't seem to work with other plugins that interact with the bundling process (like `rehype` and `remark`).

Was there a specific reason why this was chosen which might now cause issues?

epatters: No specific reason. I remember being super confused about how to make mdx work with Solid/Vite and through trial-and-error stumbled onto something that seemed to work. I can't say I'm surprised that it doesn't actually work properly.

packages/frontend/vite.config.ts
epatters: Sounds good. As I noted below, I stumbled my way through this without understanding how all the pieces fit together.

packages/frontend/vite.config.ts
tim-at-topos: this was already there — I just left it in!

packages/frontend/vite.config.ts
tim-at-topos: oh that's good to know, I was just making double sure that the errors weren't due to some `$$` vs `$` mistake

tim-at-topos:
thank you so much @jmoggr !

PR #549, opened 2025-06-27
Default name for arrows in analyses

tim-at-topos:
I think it would be nice if arrows still have a user-friendly default name if one is not set, essentially for the purpose of improving the usability of certain analyses (such as knowing which interaction coefficient corresponds to which arrow in Lotka–Volterra analyses for causal loop diagrams), but I can imagine imagining other future uses too.

My suggestion for a "sensible" user-friendly name would be something simple like "A→B", where A and B are the domain and codomain (resp.). Maybe we need to worry about multiarrows or other fancy things at some point, but I wouldn't want that to bikeshed this if we can avoid it.

For now, this PR is a proof of concept that just does this (very horribly, since it was written last thing on a Friday evening) in the frontend, at the point where the Lotka–Volterra analysis panel looks for the name of the morphism. Here are some questions/desiderata:

1. Is this something that should live further back in the codebase, somewhere in the backend? Maybe some function that says "oh the name of a morphism is the concatenation of the name of its domain, an arrow, and the name of its codomain" that the frontend can then use.
2. Something I couldn't figure out was how to (imo) improve the UI even more by including the *sign* of the arrow (in the case of causal loop diagrams) in front of the user-input, so that it's clear that e.g. they are indeed entering a negative value into the simulation by entering a positive one into the textbox.

---

Answers to the questions above:

1. This should be purely frontend, but somewhere more general than Lotka–Volterra
2. Bump this to later down the road

### DIFF
--- a/packages/frontend/src/stdlib/analyses/linear_ode.tsx
+++ b/packages/frontend/src/stdlib/analyses/linear_ode.tsx
@@ -8,8 +8,7 @@ import {
     Foldable,
     createNumericalColumn,
 } from "../../components";
-import type { MorphismDecl, ObjectDecl } from "../../model";
-import { morNameOrDefault } from "../../model/types";
+import { type MorphismDecl, type ObjectDecl, morNameOrDefault } from "../../model";
 import type { ModelAnalysisMeta } from "../../theory";
 import { ODEResultPlot } from "../../visualization";
 import { createModelODEPlot } from "./simulation";
--- a/packages/frontend/src/stdlib/analyses/lotka_volterra.tsx
+++ b/packages/frontend/src/stdlib/analyses/lotka_volterra.tsx
@@ -13,8 +13,7 @@ import {
     Foldable,
     createNumericalColumn,
 } from "../../components";
-import type { MorphismDecl, ObjectDecl } from "../../model";
-import { morNameOrDefault } from "../../model/types";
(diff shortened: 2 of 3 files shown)


tim-at-topos:
to-do: rewrite this using the nice TypeScript `?`

packages/frontend/src/util/default_names.ts
epatters, line 1: The `util` folder is intended for typically quite general stuff that doesn't fit elsewhere. I would put this function in `model/types.ts`.

(I've heard it said one should never create a module called `util` at all, since anything that one might be tempted to put there, ought instead go somewhere more meaningfully named, but alas, here I am :)

packages/frontend/src/util/default_names.ts
```diff
@@ -0,0 +1,21 @@
+import type { MorDecl } from "catlog-wasm";
+import type { ModelAnalysisProps } from "../analysis/types";
+
+export function morNameOrDefault<T>(mor: MorDecl, props: ModelAnalysisProps<T>) {
```
epatters: It would be cleaner, I think, to pass only what you need. Also nice to explicitly state the return type, especially when it's short. So:

```typescript
function morNameOrDefault(mor: MorDecl, objectNameMap: Map<Uuid, string>): string {
    ...
}
```

PR #553, opened 2025-06-30
Rework logic help pages and add analysis help pages

tim-at-topos:
- [x] Shorter `/help/` homepage
- [x] Created `/help/concepts/`
- [x] Renamed `/theories/` to `/logics/`
- [x] Renamed `/theory/...` to `/logics/` (note the singular to plural)
- [x] Better navigation (e.g. breadcrumbs, or navbar)
- [ ] ~~Underline current page in navbar~~
- [x] Instances documentation on each individual logic page (when relevant)
- [x] Analyses documentation on each individual logic page
- [ ] ~~(Bonus) Lots more "Further reading" links~~
- [x] General styling for the individual logic pages
- [x] Rewrite `/help/quick-intro/` (which is now `/help/example-models/`)
- [ ] ~~Content (and formatting) for `/guides/`~~ **[future PR]**
- [x] Fix routing for `/guides/...` including 404s
- [x] "Permalink" tree titles in the dev docs forest
- [x] Fixed 404 page formatting
- [ ] ~~Guide to the fundamentals of CatColab (`/guides/fundamentals/`) (following e.g. [Evan's slides](https://www.epatters.org/assets/slides/2025-act/slides.html))~~ **[future PR]**
- [x] (Bonus) Move the analysis documentation to individual files (just as for the logics) and automatically include them on each logic page
- [x] Address all `TO-DO`


tim-at-topos:
The user docs all speak of **logics**, but the URLs (and directory structure) speak instead of **theories**. I think it would be better to be more consistent here: either update all the docs to say "logics (also known as theories), or change all the occurrences of "theories" in URLs and links to "logics"; I think the latter sounds nicer.

Thoughts?

Edit: It seems like there's general agreement that we should do the latter, so I'll do that.

tim-at-topos:
@epatters I think the credits page (/help/credits/) probably needs updating (in particular for recognising ARIA for funding)

tim-at-topos:
My understanding of (S)CSS is that the ampersands in e.g. `packages/frontend/src/help/help_container.css` are unneccesary (and I also can't see any difference anywhere, even outside of the help page, when I remove them). @jmoggr could you confirm/deny this before I make the move to remove them for clarity?

epatters:
> @epatters I think the credits page (/help/credits/) probably needs updating (in particular for recognising ARIA for funding)

Thanks for pointing this out. The credits page has become slightly stale, as it tends to do. In addition to mentioning ARIA, we should mention you!

I'll make these updates in a separate PR. *Update*: PR https://github.com/ToposInstitute/CatColab/pull/557

epatters:
> My understanding of (S)CSS is that the ampersands in e.g. `packages/frontend/src/help/help_container.css` are unneccesary

I'm not an expert, but reading the [docs at MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector), it seems that the ampersand is necessary in certain cases, e.g., for pseudo-classes, but not for the common case of (ordinary) classes. I'm just been using them everywhere. Maybe that's unidiomatic. Anyway, I'll let @jmoggr chime in with his opinion.

tim-at-topos:
@epatters I was thinking of writing documentation for analyses in individual files (just as is done for each logic), and then dynamically including these into the help page for any given logic by just iterating over the analyses available for that logic. Do you think this seems sensible, or do you think it's likely that this would lead to an issue in the future (or even now, given that I don't know all the analyses) in a scenario where two logics allow the same analysis but the description of e.g. the semantics (as I've written in `packages/frontend/src/help/logic/causal-loop.mdx` for exarmple) would need to be different for each logic?

(If the above makes sense, then I might consider actually making a help page for just the analyses, and ensuring that everything is forward/backward linked — analyses to the logics that allow them, and logics to the analyses that they allow)

It would be nice to know that all the analyses available to any logic are automatically documented, and that we only need to write that documentation once in one place, but I don't want to build a super rigid over-designed system too rapidly!

epatters:
Good question. As you note, the situation is complicated by the fact that the isomorphic theories may use different domain-specific jargon, e.g., what is called a "variable" in a causal loop diagram is called a "species" in a regulatory network. A similar example coming in the future is "places" in a Petri net and "species" in a reaction network.

That said, an analysis applying to several theories can likely be explained once in a way that will be intelligible in all cases. Or at least, we might proceed for now under that hope/assumption until experience proves otherwise. Having several mostly-but-not-exactly identical copies of the same text creates lots of opportunities for things to get out of sync or outdated in annoying ways.

jmoggr:
Hey Tim,

I suspect that the `HelpAnalysisById` component is more complicated than it needs to be. It looks like it was made with the future use in mind, but I currently don't feel comfortable making suggestions because I don't understand what that future looks like.

I have some questions that should help solidify my understanding of what the docs are meant to do:

### 1

in `help/logic/causal-loop.mdx` `<HelpAnalysisById theory={props.theory} analysisId="..."/>` will `props.theory` always be the `causal-loop` theory (both now and in the future)?

(my guess is yes)

### 2

from `help/logic/causal-loop.mdx`, will the following text ever need to appear in the docs again?
```
  Negative feedback loops, or *balancing loops*, are composites of path within the diagram that start and finish at the same point, and whose overall sign (given by multiplying all the signs of each constituent path) is positive.
  Reinforcing loops tend to be stabilising.

  CatColab supports this analysis by searching for morphisms from the causal loop diagram given by a free loop.
```

(My guess is yes)

### 3

Will there ever need to be theory specific documentation for an analysis, or will the documentation for an analysis be automatically applicable (and sufficiently useful) for every theory that supports an analysis?

(My guess is yes)

### 4

Are there two types of analysis, `negative-loops` and `positive-loops` that both happen to share the same "loops" interface (what is described in `help/analysis/loops.mdx`)? Do we expect many analysis to share the same interface?

(My guess is yes)

tim-at-topos:
Great questions, thanks for getting to the details! Here are some quick-ish answers — happy to chat more about them one-to-one if that's easier :-)

1. Yes! I just couldn't figure out how to get this to work
2. Surprisingly, no, not really. I've since put the last sentence into `help/analysis/loops.mdx` so it will appear in all analysis help panels that use it, but the first paragraph really is specific to causal loop diagrams
3. Yes, I'm finding that a majority of the analysis documentation is applicable to all theories (e.g. for documenting which user inputs the analysis offers, as in `help/analysis/linear-ode.mdx`), but there is a non-trivial amount that is very much theory specific. That's why I've opted for this approach of having some default text in the file in `help/analysis/FOO.mdx` and being able to add to it in the `HelpAnalysisById` element in each specific `help/logic/BAR.mdx` page.
4. Yeah, both of those are really using the same analysis tool, just with different configs (distinct enough that we want users to see them as two separate analyses, rather than one analyses with the config being toggled within). As for the future, I'm unsure about this one actually. My guess is it will be 50/50 in terms of analysis sharing interfaces versus being completely standalone, but I could be off by 49% in either direction.

jmoggr:
Tim and I had a discussion about the problems of code organisation for logic documentation. I've come up with 2 possible alternative methods. Both would recreate the `Analyses` section from [causal-loop.mdx](https://github.com/ToposInstitute/CatColab/blob/a64f23a3028e9d0d3f0973e60746c6ab0ed4bbf9/packages/frontend/src/help/logic/causal-loop.mdx) (and override a pretend analysis as an example)

Both metheds provide:
- providing default analysis documentation for all analysis in a theory
- the ability to extend the default documetnation for an analysis
- the ability to completely override/replace the default documentation for an analysis


## Method 1: 

`help/logics/causal-loop.mdx`:
```
import CausalLoopsNegtiveLoops from ...
import CausalLoopsPositiveLoops from ...

import ConfigureSubmodulesAnalysis from ...

theory documentation

<Analyses
  # this would be the content currently inside of the `HelpAnalysisById` components
  extensions={{
    "negative-loops": CausalLoopsNegtiveLoops
    "positive-loops": CausalLoopsPositiveLoops
  }}

  # this would replace the entire `HelpAnalysisById` component for the "example-loops" analysis
  overrides={{
    "example-loops": ConfigureSubmodulesAnalysis
  }}
/>

more theory documentation
```

## Method 2:

`help/logics/causal-loop.mdx`:
```
theory documentation

<Analyses />

more theory documentation
```

`help/logics-help-mapping.ts`
```
import CausalLoopsNegtiveLoops from ...
import CausalLoopsPositiveLoops from ...
import ConfigureSubmodulesAnalysis from ...

# we could could import this implicity from the theoryId, but it might be better to be explicit
import CausalLoopHelp from "./logics/causal-loop.mdx"

const helpMapping = [
  {
    theory: "causal-loop",
    component: CausalLoopHelp,
    extensions: {
      "negative-loops": CausalLoopsNegtiveLoops,
      "positive-loops": CausalLoopsPositiveLoops
    }
    overrides={
      "example-loops": ConfigureSubmodulesAnalysis
    }
  },
  ...
];
```

I think I'm leaning towards Method 2. It effectively creates a declarative layer for composing the documentation , which should be easier to extend/manage/check in the future.

However method 1 might be nicer for documentation writers, since all the documentation for a logic is defined in one place.

I think either would be fine. They are similar in difficulty (easy) to implement.

Both would require creating more `.mdx` components. For reasons that I'm unable to succinctly explain, unless we put more content in different components we can't improve much on what is currently being done. The requirement for more components might be a deal breaker for either of the proposed methods, depending on how documentation writers feel about it.

epatters:
Thanks Tim and Jason for thinking hard about how to balance the tradeoffs involved here. Of Jason's two proposals, I think I slightly favor the second, but I'm happy to defer to your collective judgment about what will be the right balance between convenience and extensibility.

tim-at-topos:
Thanks for the proposals Jason! I think I also like the second one, but I'm a bit confused as to whether or not it will support the structure that I'd like to see.

Here is a screenshot of one of the current analysis help panels:
Screenshot 2025-07-10 at 10 28 30
and here's the code that generates it, from `causal-loop.mdx`:

```html
<HelpAnalysisById theory={props.theory} analysisId="linear-ode">
  An arrow $A\xrightarrow{\pm}B$ is interpreted as the equation $\dot{B}=\pm k A$
  for some non-negative constant $k$.
  Multiple incoming arrows are interpreted additively, so e.g.
  $A\xrightarrow{+}B\xleftarrow{-}C$ is interpreted as $\dot{B}=kA-\ell C$.
</HelpAnalysisById>
```

- The part in purple (1) is taken from the name and description of the analysis itself, from `theories.tsx`
- The part in green (2) is what I've entered into the `<HelpAnalysisById>` container into `causal-loops.mdx` directly
- The part in orange (3) is taken from the "generic" documentation for this analysis in `linear-ode.mdx`

I think something that's important is that the stuff (2) is specific to this analysis *in the context of this logic*, so it really should be written in `causal-loop.mdx`. Does your proposal allow for that? If I understand correctly (which I doubt I do!) it looks like I would have to make a separate `causal-loop-linear-ode.mdx` and set this as an override in `logics-help-mapping.tsx`, right?

Essentially, as long as the structure supports (1), (2), and (3) I'm happy with whatever makes the most sense to you!

jmoggr:
Thanks for the visual Tim.

It will support the sturcture of (1), (2), and (3), but it will not keep all of the content in `causal-loop.mdx`. 

The part in green (2) Would come from components imported like this:
```
    extensions: {
      "negative-loops": CausalLoopsNegtiveLoops,
      "positive-loops": CausalLoopsPositiveLoops
    }
 ```
To use Method 2 these components would have to be imported from a separate file. With Method 1 they could be defined inline, but they would have to be written with JSX.

(1) would still come from `theories.tsx`, and (3) would still come from `help/analysis`

Would you still want to go through with this change? I'm unable to weigh the benefits of automation + checks against the PITA of splitting docs up across more files.

**Comment by jmoggr** [comment]
The only way to while automatically generate default documentation while keeping everything in one file is to inspect and manipulate the HTML produced from MDX of something like `causal-loop.mdx`. This is non-trivial, it would probably be easiest/most appropriate to do that in a plugin for MDX and that will take more time than I think I can reasonably allocate to this right now.

**Comment by jmoggr** [comment]
An example of using inline JSX with Method 1:

<Analyses
  extensions={{
    "negative-loops": (
      <p> <Math expression="A\xrightarrow{\pm}B" /> is interpreted as the equation <Math 
        expression="\dot{B}=\pm k A" /> ...</p>
     ),
    "positive-loops": CausalLoopsPositiveLoops
  }}
/>

It's not obvious that this is better than importing a component from a different file.

**Comment by tim-at-topos** [comment]
Inline JSX seems not ideal, especially given that some help documentation would contain KaTeX-rendered maths chunks, which are likely not fun to type directly as compenents.

I think I'm happy with your suggestion number 2, but could you just explain what the directory structure would look like? My guess would be that there's a directory for each logic, and within that directory is (a) the .mdx file for the parent help page, and also (b) any supplementary .mdx files for any logic-specific analysis help pages.

**Comment by tim-at-topos** [comment]
Also, I'm now happy with the content of this PR (I'm going to save writing the guides for a separate PR, because I would like more content-specific feedback for those), so once we've figured out the structure I think I'm ready to mark this for review!

**Comment by epatters** [comment]
Thanks for writing all this great content, Tim!

I don't know whether you want to do it now or later, but this PR will need to be rebased before review because it's showing an enormous diff with content from other PRs.

**Comment by tim-at-topos** [comment]
OK, I rebased, so hopefully the diffs shouldn't be so bad, but let me know if it's diverged too much again though.

I think it would be nice to get this content merged before August, but there's no real urgency. I'm happy to let @jmoggr move ahead with his suggestion (2) before asking to merge this, but if you'd rather not have any time pressure Jason then we could also merge and then redesign at a later point. Up to you!

File: packages/frontend/src/help/analysis/lotka-volterra.mdx [hunk 1]
```diff
@@ -0,0 +1,16 @@
+import { A } from "@solidjs/router";
+
+We might want to think about Lotka–Volterra dynamics whenever influence between variables is *bidirectional*, e.g. for predator-prey models, where higher population of predators means lower population of prey means lower population of predators means ... and so on.
```
epatters: Delete the " ... and so on" since the "e.g." is already do this work? Or at least avoid a space before the ellipsis.

packages/frontend/src/help/concepts.mdx
```diff
@@ -1,40 +1,20 @@
...
 While CatColab has a notebook-style interface inspired by computational
 notebooks like JupyterLab and structured document editors like Notion, its
 conceptual underpinnings are quite different from both kinds of tools. Here is
-an overview of the concepts that you'll encounter in CatColab today.
+an overview of the concepts that you'll encounter in CatColab today:
 
 <dl>
 
-<dt>Logic</dt>
+<dt>Logic (or theory)</dt>
```
epatters: In light of our previous conversation about terminology, perhaps we don't want to immediately call attention to "theory" as an alternative word. The parenthetical in the text below seems like the right level of nod to this.

packages/frontend/src/help/logics/causal-loop-delays.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+
+## Instances and analyses
+
+### Instances
+
+<Question>
+As with <A href="../causal-loop">causal loop diagrams</A>, instances are not currently supported, though we'd be curious to learn more about how they might be used in practice.
+If you have any ideas then please do get in touch.
+</Question>
```
epatters: While I like the spirit of inquiry behind this, on balance I feel like we shouldn't pose what is closer to a research question in the middle of the docs about how to use CLDs. It also seems undesirable to copy this boilerplate into the docs for every logic for which we have not enabled instances.

packages/frontend/src/help/logics/logic-not-found.mdx
```diff
@@ -0,0 +1 @@
+### not found
```
epatters: Is this a stub? Would be nice to say a little more and provide a link back to the main help page. At minimum, "not found" should be capitalized.

packages/frontend/src/help/logics/schema.mdx
```diff
@@ -0,0 +1,85 @@
...
+- **Attributes**, from entities to attribute types
+- **Operations**, from attribute types to attribute types.
+
+What this data actually "means" can be explained in terms of what it corresponds to in an *instance* of the schema (see below, under *Instances and analyses*).
+
+<Warning>
+CatColab currently only supports **single-input** operations, so many familiar examples (such as sums and averages) are not directly representable.
+
+Note that consistency constraints on the schema are also not yet supported, though they can be added informally in rich text cells.
+</Warning>
```
epatters: Not opposed to being transparent about current limitations of our system, but flagging that someone will need to update this caveat when we *do* implement those features and I'm not currently sure how that information will flow. Nothing actionable here, most likely. We can discuss more later.

packages/frontend/src/help/logics/stock-flow.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+## Description
+
+**Stock and flow** (or simply **stock-flow**) diagrams are fundamental to systems dynamics.
+They provide a rich language, allowing the scientist to model a system of **stocks** of various quantities and flows between them, with the rate of the flows depending in complex ways on various stocks via **links**.
+
+<Warning>
+In CatColab we implement what we call **primitive** stock-flow diagrams, which do not allow for the auxiliary variables permitting flows to be arbitrary functions of the stocks (which is characteristic in systems dynamics), but instead simply products of the stocks with a constant rate.
+We expect to implement the richer semantics in the future.
+</Warning>
```
epatters: I think this can be a Note/Info rather than a Warning: there are different flavors of stock-flow diagrams, and it's useful to have both simple and complicated ones.

The bigger problem is that what I'm currently calling a "stock-flow diagram" should, in official system dynamics jargon, be called a "system structure diagram," the difference between the stock-flow diagram is standardly construed to include quantitative flow functions, whereas a system structure diagram (as the name suggests) describes only the structure of the system. But let's address that in a future PR.

packages/frontend/src/index.css
```diff
@@ -1,15 +1,18 @@
...
-    --h1-font-size: 3em;
-    --h2-font-size: 2em;
-    --h3-font-size: 1.5em;
-    --h4-font-size: 1.3em;
+    --h1-font-size: 2.25em;
+    --h2-font-size: 1.75em;
+    --h3-font-size: 1.4em;
+    --h4-font-size: 1.2em;
+
+    --body-font-size: 1.05rem;
```
epatters: If I'm not mistaken these changes to the font sizes will have global effects throughout the app, not confined to the help pages. Two comments:

1. If these changes were made to achieve a particular effect on the help pages, they should be localized to the help page CSS.
2. If the aim is to affect the overall app, I'd rather handle that through a separate follow-on PR. No question that our styling could be dramatically improved throughout, but I'd like to give that proper attention rather than having it come in through a large, mostly unrelated PR.

packages/frontend/src/model/model_editor.css
```diff
@@ -6,10 +6,18 @@
 
     & .title {
         font-size: var(--h1-font-size);
+        flex: 1;
     }
 
     & .theory-selector-trigger {
         font-size: var(--h4-font-size);
-        max-width: 25%;
+        /*max-width: 25%;*/
```
epatters: Better to delete stale/replaced CSS than to leave it in but commented-out.

packages/frontend/src/model/theory_selector.tsx
```diff
@@ -22,15 +23,17 @@ export function TheorySelectorDialog(props: TheorySelectorProps) {
...
+            <TheoryHelpButton theoryMeta={props.theoryMeta} />
             <Dialog.Portal>
                 <Dialog.Overlay class="overlay" />
                 <Dialog.Content class="popup">
                     <TheorySelector
-                        theory={props.theory}
+                        theoryMeta={props.theoryMeta}
                         setTheory={(id) => {
                             props.setTheory(id);
+                            console.log(props.theoryMeta.id);
```
epatters: Lingering `console.log`.

packages/frontend/src/page/toolbar.tsx
```diff
@@ -49,30 +49,40 @@ const Brand = () => (
 
 /** Button that navigates to the help page for a theory.
 
-If no theory is set, it naviagtes instead to the list of all theories.
+If no theory is set, it navigates instead to the list of all theories.
  */
 export function TheoryHelpButton(props: {
     theory?: Theory;
+    theoryMeta?: TheoryMeta;
```
epatters, line 56: Hmm, seems like this component should take either a `Theory` or a `TheoryMeta`, but not potentially both while arbitrarily preferring one over the other. I'd have to look closer to diagnose why this was done, but it should be avoidable by a suitable refactoring.

packages/frontend/src/stdlib/types.ts
```diff
@@ -16,16 +16,15 @@ export type TheoryMeta = {
...
      */
     isDefault?: boolean;
 
     /** Group to which the theory belongs. */
     group?: string;
 
-    /* Name of help page for the theory. */
-    help?: string;
+    /** Does a help page for this theory exist? */
+    help?: boolean;
```
epatters, line 27: It's too bad we have to manually supply this flag. Am I correct that this should be `true` if and only if there exists an appropriately named MDX file in `help/logics`? The problem is that I don't know whether it's possible to obtain this information at runtime. Doesn't have to be addressed right now, but flagging for future consideration.

epatters (changes requested):
Thanks Tim, this is excellent! A number of mostly small comments below.

As a general point, I think that Warnings are used a little too liberally in the docs. I noted one particular case of this but overall I think that these really draw the eye and should therefore be used sparingly, namely when something could meaningfully go wrong by ignoring it, not just that some desirable feature is missing.

packages/frontend/src/help/logics/schema.mdx
tim-at-topos: Yeah, I agree that this is part of something we need to figure out a good solution for in general: how can we be sure that need for documentation updates are flagged whenever relevant code is changed. I'm not sure whether our solution to this will be primarily cultural or technological or somewhere in between.

packages/frontend/src/page/toolbar.tsx
tim-at-topos: I ran into this issue because I couldn't find access to the *theory* but only the metadata, so I hacked this in (with the help (and thus blame) of @olynch). In general though, I think it makes sense that the help button only needs the theory meta? Not only this, but it seems possibly reasonable that we should have some helper function around that recovers the corresponding Theory from the the `id` of a TheoryMeta (this is precisely what I do in this case anyway).

Would you like to fix this now, or merge as-is and then open an issue and assign it to me?

packages/frontend/src/stdlib/types.ts
tim-at-topos: Correct, and I think this is something that we could merge now and immediately open an issue for, but I'm happy to postpone and try to fix this in this PR if you prefer.

tim-at-topos: Actually, a small correction here: the current behaviour is that this flag determines whether or not the logic actually shows up on the "list of logics" help page (`/help/logics`) regardless of whether or not a corresponding MDX file exists; if the corresponding MDX file does *not* exist then the help page will just contain the automatically generated information of `modelTypes` and `modelAnalyses` with their (short) descriptions

packages/frontend/src/page/toolbar.tsx
epatters: Whichever you prefer. I don't mind merging as-is as long as an issue is created.

epatters (approved):

packages/frontend/src/page/toolbar.tsx
tim-at-topos: https://github.com/ToposInstitute/CatColab/issues/611

PR #560, opened 2025-07-05
Run NixOS integration tests in CI

### LINKED ISSUE #531: Make Nix definitions in a VM
Jason Moggridge [said](https://catcolab.zulipchat.com/#narrow/dm/784783,784933,890432-group/near/521339254):

Sometime soonish it might be worthwhile to spend 1-3hr to try and make the Nix definitions runnable in a VM. It would greatly reduce the number of iterations and time taken to debug issues like these.

jmoggr:
Closes #531

Summary of changes

#### 2: Use `deploy-rs` to deploy from CI
  - this should speed up deployment times, building the nix profile in CI is much faster than on AWS. Also we get the failsafe/automatic rollback through `deploy-rs`, which is nice.
  - deploy using the `catcolab` user instead of `root`
    - this required disabling the password prompt for the `sudo` command
      -  This has a side effect of being a quality of life improvement: since there was no password for the `catcolab` user, the `sudo` command was effectively useless. I basically always had to log in as `root`.
      - After the next deploy to prod we could disable `root` logins through ssh, leaving only the `catcolab` as a single entry point.
        - I'm not sure if we actually want to do this. It is a best practice, however on my personal machines I root access open in case of a major screw up.
    - this change is not strictly needed, it was just the easiest way to remove the `deployuser` user since `deploy-rs` was working (`deployuser` only existed to work around `deploy-rs`, and I'm not certain that it was ever actually needed, it just sort of came about organically)
    - removes the `deployuser` user, but repurposes the `deployuser` ssh key as a key for the `catcolab` user.
      - (we effectively just merged `deployuser` into `catcolab`)

#### 3: Add `catcolab-vm` configuration
- used for running tests and testing service changes locally
- it does not have access to secrets

#### 4: refactor nix module definitions
- Fully separate the host specific config from the services config
   - This was required because the VM host is somewhat different and forces us to do a better job separating concerns.
- The new `catcolab` module is more idiomatic, it would not look out of place in `nixpkgs`
  - Other people wanting to host an instance of Catcolab would just need to import the `catcolab` module and configure it.
- Most of the diff is from moving files around and changing imports, git didn't do a great job at capturing that. If that's a major issue in reviewing this I can break it down into easier to track changes.

#### 5: Add an `integrationTests` NixOS check
- this is a python script that runs in a custom environment which has access to the `catcolab-vm`
- it checks to make sure that all the services start
  - it does not yet run the integration tests that we have for our frontend. It could/should, but I felt that it was out of scope.
- this greatly reduces the risk of having the services fail to start after a deployment.

#### 6: Unified `backendSecretsForCatcolab` and `backendpSecretsForPostgres` into `catcolabSecrets`
- Secrets are now always loaded into the env via the systemd service, removing a lot of fiddly permissions management.


infrastructure/hosts/catcolab-next/default.nix
```diff
@@ -10,47 +12,49 @@ let
 in
 {
   imports = [
-    ../../modules/backend.nix
-    ../../modules/host.nix
-    ../../modules/backup.nix
-    "${inputs.nixpkgs}/nixos/modules/virtualisation/amazon-image.nix"
```
jmoggr, line 16: Referencing `inputs` like this caused an infinite recursion error for very nixy reasons. `modulesPath` is the more "correct" method for this.

epatters:
> * This has a side effect of being a quality of life improvement: since there was no password for the `catcolab` user, the `sudo` command was effectively useless. I basically always had to log in as `root`.

Yep, that was annoying! Thanks for fixing it.

> * we effectively just merged `deployuser` into `catcolab`

Also a nice simplification.

> * it does not yet run the integration tests that we have for our frontend. It could/should, but I felt that it was out of scope.

Sounds good, I created an issue to track this (#593).

README.md
```diff
@@ -112,3 +125,7 @@ over time, is inspired by a wide body of research in applied category theory and
 beyond. Incomplete bibliographies are in the [dev
 docs](https://next.catcolab.org/dev/bib-0001.xml) and the [core
 docs](https://next.catcolab.org/dev/rust/catlog/refs).
+
+
+## TODO: move to dev docs
```
epatters: I think it's fine to leave the info above in the README, but this TODO should be removed from the README.

epatters (approved):

PR #562, opened 2025-07-07
Notion-like navigation UI

### LINKED ISSUE #447: Notion-like UI for navigation

jmoggr:
Closes #447

### DIFF
--- a/packages/frontend/src/page/document_sidebar.tsx
+++ b/packages/frontend/src/page/document_sidebar.tsx
@@ -21,7 +21,7 @@ export function DocumentSidebar(props: {
         <>
             <AppMenu />
             <RelatedDocuments liveDocument={props.liveDocument} />
-            <NewModelItem2 />
+            <NewModelItem />
         </>
     );
 }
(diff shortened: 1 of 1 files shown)


epatters:
Superseded by #749.

PR #563, opened 2025-07-07
REFACTOR: split up get_ref_stubs query to make reusable parts

jmoggr:


jmoggr:
This change was included in #749

PR #565, opened 2025-07-07
Customize Colors for Stock-Flow plots

quffaro:
This is proof of concept for custom visualization for stock-flow analyses.


epatters:
Per our discussion in this week's dev meeting, I'm closing out this PR in favor of:

- fixing the bug that color assignments are unstable, likely by just sorting the UUIDs before feeding them into e-charts
- investigating whether e-charts has support for color palettes

PR #574, opened 2025-07-09
Adding DefaultToolbar component to ErrorBoundaryDialog

quffaro:


quffaro:
@epatters I merged in a fix from @jmoggr which tosses out the corvu dialog component for a new error page component, allowing us to keep the menubar intact. His full explanation is given his PR #577, but my opinion on his comment

> This is a bit hacky, ideally we would have a layout with a toolbar and some content element, then just wrap that content element in a error boundary, instead of having to recreate part of the page in an error component. This might be addressed by the notion style UI, and if it's not then we should address this hierarchy problem after it's merged.

is that I agree we should wrap the page content element in the error boundary, but that we should consider that refactor alongside whichever changes are happening in the Notion style UI refactor.

epatters:
@jmoggr, how do you want to handle this one? I'd expect anything here to conflict with your work on #447.

epatters:
I'm going to close this PR as it's now quite stale. @quffaro, feel free to open a new PR along these lines if you're interested. I believe the issue that it was solving still stands.

PR #575, opened 2025-07-10
Remove duplicate exports for ObType and MorType

### LINKED ISSUE #567: Duplicate `ObType` and `MorType` exported in catlog_wasm.d.ts
When I run `pnpm run build` and look at `packages/catlog-wasm/pkg/catlog_wasm.d.ts`, there are two exports for `ObType` (lines 31 and 71), which are identical:

```{typescript}
/**
 * Object type in a double theory.
 */
export type ObType = { tag: "Basic"; content: Ustr } | { tag: "Tabulator"; content: MorType };
```

However, there are two *non-identical* exports for `MorType` (lines 36 and 66), with the latter not including the `Composite` tag type:

```{typescript}
/**
 * Morphism type in a double theory.
 */
export type MorType = { tag: "Basic"; content: Ustr } | { tag: "Composite"; content: MorType[] } | { tag: "Hom"; content: ObType };

...

/**
 * Morphism type in a double theory.
 */
export type MorType = { tag: "Basic"; content: Ustr } | { tag: "Hom"; content: ObType };
```

This is causing issues in the ECLD branch, because we use `Composite`, and it seems like the latter declaration overwrites the former, which results in the following browser console error when trying to actually create one of these composite morphisms:

```
Error: Error: unknown variant `Composite`, expected `Basic` or `Hom`
```

---

I *think* the problem comes from the fact that both `packages/catlog-wasm/src/theory.rs` and `packages/notebook-types/src/v0/theory.rs` define a `MorType`, and the latter only includes `Basic` and `Hom`.

tim-at-topos:
closes #567

### DIFF
--- a/packages/catlog-wasm/src/theory.rs
+++ b/packages/catlog-wasm/src/theory.rs
@@ -16,33 +16,6 @@ use catlog::dbl::theory;
 use catlog::dbl::theory::{DblTheory as _, TabMorType, TabObType};
 use catlog::one::Path;
 
-/// Object type in a double theory.
-#[derive(Clone, PartialEq, Eq, Hash, Debug, Serialize, Deserialize, Tsify)]
-#[serde(tag = "tag", content = "content")]
-#[tsify(into_wasm_abi, from_wasm_abi)]
-pub enum ObType {
-    /// Basic or generating object type.
-    Basic(Ustr),
-
-    /// Tabulator of a morphism type.
-    Tabulator(Box<MorType>),
-}
-
-/// Morphism type in a double theory.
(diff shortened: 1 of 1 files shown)


tim-at-topos:
@epatters ok, looking at all the failing build errors, it was definitely not as simple as just removing those lines completely... maybe I misunderstood what you meant

epatters:
Most of the errors are just imports that need to be fixed, but there will be a couple lines of substantive changes that I didn't think about.

I'll take care of this tonight.

epatters:
Replaced by #578.

PR #576, opened 2025-07-10
Fix URLs for dev docs sections

tim-at-topos:
I think this should mean that the following three URLs now work:

- https://next.catcolab.org/dev/api.xml
- https://next.catcolab.org/dev/mathematics.xml
- https://next.catcolab.org/dev/technologyxml

which would be helpful for making nice links to things (especially in https://github.com/ToposInstitute/CatColab/pull/553)

@olynch is this indeed the right way to make this happen?


olynch:
Yes, that is right!

PR #577, opened 2025-07-10
FIX: Menu not working in some error states

jmoggr:
The problem was that the menu relies on the `PageActionsContext`, which is not provided at the root level of the app. We can put an `ErrorBoundary` in `PageContainer`, however that won't show the menu since the menu is part of the "page" that is being contained. We can overcome this by creating another fallback component `ErrorBoundaryPage` which assumes that it will take up the entire page and add it's own `DefaultToolbar`.

This is a bit hacky, ideally we would have a layout with a toolbar and some content element, then just wrap that content element in a error boundary, instead of having to recreate part of the page in an error component. This might be addressed by the notion style UI, and if it's not then we should address this hierarchy problem after it's merged.

Also there's no good reason for the error message to be a dialogue, I had thought that there might still be something visible in the error boundary that we would put a dialogue on top of (iirc this is what React does), but instead the area in the ErrorBoundary is just not rendered at all so there's no reason not to replace it.

Feel free to edit any of the error stuff I'm adding here.

### DIFF
--- a/packages/frontend/src/model/model_editor.tsx
+++ b/packages/frontend/src/model/model_editor.tsx
@@ -38,6 +38,7 @@ import {
 import "./model_editor.css";
 
 export default function ModelPage() {
+    throw "an error";
     const api = useApi();
     const theories = useContext(TheoryLibraryContext);
     invariant(theories, "Must provide theory library as context to model page");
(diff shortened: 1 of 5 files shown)


packages/frontend/src/model/model_editor.tsx
```diff
@@ -38,6 +38,7 @@ import {
 import "./model_editor.css";
 
 export default function ModelPage() {
+    throw "an error";
```
jmoggr, line 41: I left this in here to demo the fix, because it matters where the error comes from. Errors that happen outside of `PageContainer` will not have a working menubar.

PR #578, opened 2025-07-11
Remove redundant type defs and complete notebook-types refactor

### LINKED ISSUE #567: Duplicate `ObType` and `MorType` exported in catlog_wasm.d.ts

epatters:
Closes #567. This was more work than I expected since it turns out that #473 hadn't finished refactoring `ObType` and `MorType` to use the quoting and elaboration traits.

### DIFF
--- a/packages/catlog-wasm/src/model.rs
+++ b/packages/catlog-wasm/src/model.rs
@@ -17,6 +17,7 @@ use catlog::one::{Category as _, FgCategory, Path, fp_category::UstrFpCategory};
 use catlog::validate::Validate;
 use notebook_types::current::{path as notebook_path, *};
 
+use super::notation::*;
 use super::result::JsResult;
 use super::theory::{DblTheory, DblTheoryBox};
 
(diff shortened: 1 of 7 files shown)


packages/catlog-wasm/src/model.rs
```diff
@@ -235,25 +193,6 @@ impl CanQuote<Path<TabOb<Uuid, Uuid>, TabEdge<Uuid, Uuid>>, Mor> for Quoter {
...
-            Mor::Composite(Box::new(demote_path(path.clone().map(Ob::Basic, Mor::Basic))))
-        }
-    }
-}
-
-impl CanQuote<Path<Ustr, Ustr>, MorType> for Quoter {
-    fn quote(&self, path: &Path<Ustr, Ustr>) -> MorType {
-        match path {
-            Path::Id(u) => MorType::Hom(Box::new(ObType::Basic(*u))),
-            Path::Seq(non_empty) => MorType::Basic(non_empty[0]),
```
epatters, line 252: This was also a bug.

PR #584, opened 2025-07-12
Completions match within string

quffaro:
This small PR replaces `startsWith` with `includes`


epatters (approved):
Thanks Matt, works great.

PR #585, opened 2025-07-12
Button in permissions dialog to copy document link to clipboard

quffaro:


packages/frontend/src/user/permissions.tsx
```diff
@@ -211,7 +234,7 @@ function AnonPermissionsButton() {
         await signOut(getAuth(firebaseApp));
         setOpen(false);
     };
-
+    1;
```
jmoggr: typo?

packages/frontend/src/user/permissions.css
```diff
@@ -13,3 +13,16 @@
     align-items: center;
     gap: 5em;
 }
+
+.button-container {
```
epatters: This style isn't in a CSS module, so it applies globally across the app. To avoid potential conflicts, it should have a more specific name, like `permissions-button-container`.

packages/frontend/src/user/permissions.css
```diff
@@ -13,3 +13,16 @@
...
+    display: flex;
+}
+
+.utility {
+    color: var(--main-color);
```
epatters: Seems odd to me to have this here. Put this in `form button.utility` above?

packages/frontend/src/user/permissions.css
```diff
@@ -13,3 +13,16 @@
...
+}
+
+.spacer {
```
epatters: Ditto about `spacer`, too generic for use without namespacing.

packages/frontend/src/user/permissions.tsx
```diff
@@ -94,6 +96,18 @@ export function PermissionsForm(props: {
         props.onComplete?.();
     };
 
+    const doclink = createMemo(() => {
+        return window.location.href;
+    });
```
epatters: A memo here is overkill: no need to memo-ize a property that costs nothing to compute. Delete this and replace `doclink()` below with `window.location.href`.

packages/frontend/src/user/permissions.tsx
```diff
@@ -94,6 +96,18 @@ export function PermissionsForm(props: {
...
+
+    const copyToClipboard = async () => {
+        if (navigator.clipboard) {
+            await navigator.clipboard.writeText(doclink());
+        } else {
+            Error("Could not be copied.");
```
epatters: If you want to throw an error, should be `throw new Error(...)`. I'd also make the error more description, something like "Link to document could not be copied."

packages/frontend/src/user/permissions.tsx
quffaro: thanks for spotting that, @jmoggr

epatters (approved):

PR #588, opened 2025-07-14
Drop indicator for dragging cells

quffaro:


packages/frontend/src/model/model_editor.tsx
```diff
@@ -84,7 +84,6 @@ export function ModelPane(props: {
     const liveDoc = () => props.liveModel.liveDoc;
 
     const selectableTheories = () => {
-        console.log(props.liveModel.theory().inclusions);
```
epatters, line 87: Thanks for noticing and fixing this! I take responsibility for this one :)

packages/frontend/src/notebook/notebook_cell.tsx
```diff
@@ -134,11 +138,17 @@ export function NotebookCell(props: {
...
+    const [dropTarget, setDropTarget] = createSignal(false);
+
     createEffect(() => {
         const cleanup = combine(
             draggable({
                 element: handleRef,
-                getInitialData: () => createCellDragData(props.cellId),
+                getInitialData: () => ({
+                    ...createCellDragData(props.cellId, props.index),
+                    index: props.index,
```
epatters: If I'm not mistaken, this is redundant: your modified `createCellDragData` already adds the `index` field to the returned data. So I don't think you need to make any change here at all.

packages/frontend/src/notebook/notebook_cell.tsx
```diff
@@ -147,13 +157,27 @@ export function NotebookCell(props: {
...
                 getData({ input }) {
-                    const data = createCellDragData(props.cellId);
+                    const data = createCellDragData(props.cellId, props.index);
                     return attachClosestEdge(data, {
                         element: rootRef,
                         input,
                         allowedEdges: ["top", "bottom"],
                     });
                 },
+                onDragEnter: (args) => {
```
epatters: Not actually important, but there's syntactic sugar which would let you write this first line as `onDragEnter(args) {`

epatters (changes requested):
Thanks Matt, this will be a big quality of life improvement!

epatters (approved):

PR #594, opened 2025-07-17
Changed line height to avoid descender clipping

quffaro:
before:
image

after:
image

This small PR increases the line height of boxes in order to avoid clipping of descenders.


epatters (approved):
Thanks Matt, this is nice simple fix.

PR #601, opened 2025-07-21
Mass-action ODE semantics for Petri nets

epatters:
Sequel to #597 adding the ODE semantics for Petri nets based on the law of mass action.

I also took the opportunity to refactor the linear ODE/Lotka-Volterra analyses to reuse the code that constructs the coefficient matrix (cc @tim-at-topos).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1871,9 +1871,9 @@ dependencies = [
 
 [[package]]
 name = "nonempty"
-version = "0.10.0"
+version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


tim-at-topos:
Not a proper review, but just to say that this makes sense and is much tidier! I think that we will end up modifying `packages/catlog/src/stdlib/analyses/ode/signed_coefficients.rs` for some of the ECLD analyses, but it shouldn't change the way you've separated things out here.

epatters:
Thanks for taking a look! Yes, I'm happy to further evolve this design, but hopefully what's here is already an improvement on the current situation.

PR #603, opened 2025-07-21
Migrate notebook format to support cell reordering

jmoggr:
This implements the method for migrating notebooks/documents described by Owen here: https://github.com/ToposInstitute/CatColab/blob/main/packages/notebook-types/README.md

It deviates on this point:
> We make a sum type over all versions with a .to_current() method which upgrades from any previous version to the current version.

Because it would require each version to know how to upgrade to versions more recent than itself. I think In theory you could do this with clever interfaces and a recursive call, however I think that from a design perspective the logic for going from one version to some arbitrary current version should live outside the context of individual versions. (Also my rust is not good enough to do what Owen described, I tried and failed)

The equivalent `.to_current` is now is a member of the sum type.


flake.nix
```diff
@@ -2,7 +2,7 @@
   description = "configurations for deploying catcolab";
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs/nixos-24.11";
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-25.05";
```
jmoggr, line 5: Support for 24.11 ended on June 30. The next version, 25.11, will be released on November 30, and 25.05 will become unsupported on on December 31st.

It might be good to set up one or two calendar notifications for upgrading the NixOS version? The other upgrade window is May 31 to June 30.

Crane needed 25.05, so the upgrade got lumped in. The only change that the upgrade required was replacing nodejs_23 with nodejs_24.

flake.nix
```diff
@@ -39,25 +49,83 @@
...
           config.allowUnfree = true;
         };
 
+      rustToolchainFor =
+        system:
+        inputs.fenix.packages.${system}.fromToolchainFile {
+          file = ./rust-toolchain.toml;
+          sha256 = "sha256-Qxt8XAuaUR2OMdKbN4u8dBJOhSHxS+uS06Wl9+flVEk=";
+        };
+
```
jmoggr, line 57: This fixes a previous minor oversight: the rust toolchain can differ by system in the same way that nixpkgs can.

packages/frontend/src/diagram/document.ts
```diff
@@ -69,9 +70,13 @@ function enlivenDiagramDocument(
     const { doc } = liveDoc;
 
     const formalJudgments = createMemo<Array<DiagramJudgment>>(() => {
-        return doc.notebook.cells
-            .filter((cell) => cell.tag === "formal")
-            .map((cell) => cell.content);
+        return (
+            doc.notebook.cellOrder
+                // biome-ignore lint/style/noNonNullAssertion:
```
jmoggr: Unfortunately I'm adding a lot of these since we now have to do a lookup on an object.

The only thing to do to avoid these assertions is to have an explicit check and throw an error if it fails, however that is a lot more verbose than just letting it fail from being null, especially since we are not expecting this to fail.

The circumstances are different enough each time that there are never more than two examples of the same access pattern, so I didn't think it warranted new helper functions.

It feels messy, but it's the best I could come up with. I'm open to other ideas.

packages/notebook-types/src/lib.rs
```diff
@@ -1,7 +1,91 @@
...
+}
+
+pub static CURRENT_VERSION: u32 = 1;
+
+#[wasm_bindgen(js_name = "currentVersion")]
+pub fn current_version() -> u32 {
+    CURRENT_VERSION
+}
+
+impl<'de> Deserialize<'de> for VersionedDocument {
```
jmoggr, line 32: Custom deserializer for versioned documents: there were some confounding factors that mandated this over "regular" serde decordators:
- v0 does not have a version number
- Serde expects a tag field to be a string, and the version field is a number
- Serde expects the content of a tagged enum to be in a single property.

I think this is more intuitive than the serde decorators would be (if I could get them to work...), with the downside that we need to update this function on every new version.

flake.nix
```diff
@@ -14,9 +14,18 @@
       inputs.nixpkgs.follows = "nixpkgs";
     };
 
+    crane = {
```
jmoggr, line 17: Yup, that's right. I'm adding a second method for building rust packages alongside `crate2nix`...

There is a mind numbing bug in `crate2nix` that makes it difficult to support the WASM target: [link](https://github.com/cargo2nix/cargo2nix/issues/286). "Regular"/vanilla methods for building a WASM target also failed for dense reasons.

In retrospect Crane is a MUCH better library than `crate2nix`, I think I was somewhat suckered by the `nix-community` label on `crate2nix`. It's a lot more intuitive, easier to customise, and does not require the weird `Cargo.nix` file. I would like to use it for everything, HOWEVER currently I see no way of using it without it rebuilding all of our rust packages on any source code change. I suspect this is a bug/oversight and will open an issue with them. In the meantime `catlog-wasm` is build any ANY source code change, but it builds fast enough (adding ~1min to CI times) that I think we can live with it for now.

packages/automerge-doc-server/package.json
```diff
@@ -13,8 +14,10 @@
     "dependencies": {
         "@automerge/automerge-repo": "^1.2.1",
         "@automerge/automerge-repo-network-websocket": "^1.2.1",
+        "catlog-wasm": "link:../catlog-wasm/dist/pkg-node",
```
epatters: Ideally, the Automerge sync server would depend only on `notebook-types` (which we ought to eventually rename to `document-types`), rather than all of `catlog-wasm`. The latter pulls in various dependencies to do with ODE simulation that are quite irrelevant to the sync. I assume you're depending on `catlog-wasm` just because the Wasm compilation target was already set up?

packages/automerge-doc-server/src/server.ts
```diff
@@ -119,6 +122,21 @@ export class AutomergeServer implements SocketIOHandlers {
...
+        // In order to register changes from the migrations with Automerge we diff the original document (which
+        // is the proxy object) with the output of the migrations (which is a plain JSON object) and apply the
+        // diff to the original object. The application of the diff happens entirely is JS land, so the changes
+        // are captured by Automerge.
+        const docBefore = await handle.doc();
+        const docAfter = catlog.migrateDocument(docBefore);
+        const patches = jsonpatch.compare(docBefore as any, docAfter);
+        handle.change((doc: any) => {
+            jsonpatch.applyPatch(doc, patches);
+        });
```
epatters, line 138: Slick use of patching. I wasn't previously aware of JSON-patch.

packages/notebook-types/src/v1/notebook.rs
```diff
@@ -0,0 +1,41 @@
...
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use tsify::Tsify;
+use uuid::Uuid;
+
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct Notebook<T> {
+    #[serde(rename = "cellContents")]
+    #[tsify(type = "Record<string, NotebookCell<T>>")]
```
epatters: You should be able to avoid this manual type def by replacing the top-level directive with

```rust
tsify(into_wasm_abi, from_wasm_abi, hashmap_as_object)
```

packages/catlog-wasm/src/model.rs
```diff
@@ -434,7 +434,9 @@ pub fn collect_product(ob: Ob) -> Result<Vec<Ob>, String> {
 #[wasm_bindgen(js_name = "elaborateModel")]
 pub fn elaborate_model(doc: &ModelDocumentContent, theory: &DblTheory) -> DblModel {
     let mut model = DblModel::new(theory);
-    for cell in doc.notebook.cells.iter() {
+
+    for cell_id in &doc.notebook.cell_order {
+        let cell = &doc.notebook.cell_contents[cell_id];
```
epatters: Not sure whether we want to treat the structs in `notebook-types` as "pure data" or genuine Rust types, but we might considering adding a helper method `cells(&self) -> impl Iterator<Item = &Cell>` to the `Notebook` struct, so that this change and the one below are as simple as replacing `doc.notebook.cells.iter()` with `doc.notebook.cells()`.

packages/frontend/src/model/document.ts
```diff
@@ -72,9 +74,14 @@ function enlivenModelDocument(
     // Memo-ize the *formal* content of the notebook, since most derived objects
     // will not depend on the informal (rich-text) content in notebook.
     const formalJudgments = createMemo<Array<ModelJudgment>>(() => {
-        return doc.notebook.cells
+        console.log(JSON.parse(JSON.stringify(doc.notebook)));
```
epatters: Lingering `console.log` statement.

packages/frontend/src/model/document.ts
```diff
@@ -72,9 +74,14 @@ function enlivenModelDocument(
...
+        const a = doc.notebook.cellOrder
+            // biome-ignore lint/style/noNonNullAssertion:
+            .map((cellId) => doc.notebook.cellContents[cellId]!)
             .filter((cell) => cell.tag === "formal")
             .map((cell) => cell.content);
+        // console.log("after");
+        return a;
```
epatters: More cruft from debugging. I assume you want to just return the above instead of binding it to `a`.

packages/frontend/src/notebook/notebook_cell.tsx
```diff
@@ -63,9 +62,6 @@ export type CellActions = {
     /** Delete this cell in the forward/downward direction. */
     deleteForward: () => void;
 
-    /** Duplicate this cell, adding the new cell below this one. */
-    duplicate: () => void;
-
```
epatters, line 68: I don't think we should just remove the duplicate feature entirely. Instead:

- The `duplicate` field should become optional:

```typescript
duplicate?: () => void
```

- The menu item should be visible only when `duplicate` is defined
- It should be defined as before for all cell types *except* rich text

packages/frontend/src/diagram/document.ts
epatters: Yeah, the extra indirection is inherently a little annoying, but in this case and the one below, couldn't you avoid the lint override by writing:

```typescript
doc.notebook.cellOrder
                .map((cellId) => doc.notebook.cellContents[cellId])
                .filter((cell) => cell?.tag === "formal")
```

packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -249,60 +247,74 @@ export function NotebookEditor<T>(props: {
...
             </Show>
             <ul class="notebook-cells">
-                <For each={props.notebook.cells}>
-                    {(cell, i) => {
+                <For each={props.notebook.cellOrder}>
+                    {(cellId, i) => {
+                        const cell = props.notebook.cellContents[cellId];
+                        if (!cell) {
+                            throw `Failed to find contents for cell ${cellId}`;
+                        }
```
epatters: I agree that in this key spot, it's better to throw an error than silently ignore the missing cell, because this is an important invariant that should never fail. In fact, using `tiny-invariant`, I like to write these checks as

```typescript
invariant(cell, () => `Failed to find contents for cell ${cellId}`);
```

which saves a line or two of boilerplate.

packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -249,60 +247,74 @@ export function NotebookEditor<T>(props: {
...
                             },
                             createBelow() {
+                                console.log("Creating below");
                                 props.changeNotebook((nb) => {
-                                    nb.cells.splice(i() + 1, 0, newStemCell());
+                                    const newCell = newStemCell();
+                                    nb.cellOrder.splice(i() + 1, 0, newCell.id);
+                                    nb.cellContents[newCell.id] = newCell;
+                                    // @ts-ignore
+                                    nb.cellStuff = { test: "huh" };
```
epatters: Debugging cruft.

packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -249,60 +247,74 @@ export function NotebookEditor<T>(props: {
...
-                                const n = props.notebook.cells.length;
+                                const n = props.notebook.cellOrder.length;
                                 i() < n - 1 && setActiveCell(i() + 1);
                             },
                             createAbove() {
                                 props.changeNotebook((nb) => {
-                                    nb.cells.splice(i(), 0, newStemCell());
+                                    const newCell = newStemCell();
+                                    nb.cellOrder.splice(i(), 0, newCell.id);
+                                    nb.cellContents[newCell.id] = newCell;
```
epatters: Though I know it's sort of trivial, this pair of lines is appearing so frequently that it might be worthwhile to have a helper function `insertCell(notebook, i, cell)` in `src/notebook/types.ts`.

packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -249,60 +247,74 @@ export function NotebookEditor<T>(props: {
...
                                     setActiveCell(i());
                                 });
                             },
                             createBelow() {
+                                console.log("Creating below");
```
epatters: Lingering log statement.

packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -249,60 +247,74 @@ export function NotebookEditor<T>(props: {
...
                                     setActiveCell(i() + 1);
                                 });
                             },
                             deleteBackward() {
                                 props.changeNotebook((nb) => {
-                                    nb.cells.splice(i(), 1);
+                                    const cellId = nb.cellOrder[i()];
+                                    // biome-ignore lint/style/noNonNullAssertion:
+                                    delete nb.cellContents[cellId!];
+                                    nb.cellOrder.splice(i(), 1);
```
epatters: And perhaps also a helper function `deleteCell(notebook, i)`! Then you can actually check the invariant instead of having to put a non-null assertion.

packages/notebook-types/README.md
```diff
@@ -7,3 +7,24 @@ We start with `v0`, which is meant to be fully compatible with the JSON types th
...
+  use serde::{Deserialize, Serialize};
+  use tsify::Tsify;
+
+- #[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Tsify)]
+- #[tsify(into_wasm_abi, from_wasm_abi)]
++ #[derive(PartialEq, Eq, Debug, Serialize, Deserialize)]
+  pub struct Notebook<T> {
+      pub cells: Vec<NotebookCell<T>>,
+  }
+```
```
epatters, line 30: Thanks for extending these docs!

packages/notebook-types/src/lib.rs
epatters: Seems reasonable. My only comment is if it would be more convenient to make the version field a string, now is the time to do it!

epatters (changes requested):
Thanks Jason, this is excellent work! I've left some comments, mostly of a very tactical nature.

olynch:
> It deviates on this point:
> 
> > We make a sum type over all versions with a .to_current() method which upgrades from any previous version to the current version.


Actually, I think that your design here is more or less what I had in mind: putting the `.to_current()` on the sum type not the individual types. Each individual version should only know how to be converted into from the previous version, and then `.to_current()` should chain these upgrades until the right version is reached.

olynch:
I've not read in depth the whole PR, but I looked at the parts which dealt with notebook-types, and they look good to me! Happy to see that my design is useful :)

packages/notebook-types/src/v1/notebook.rs
jmoggr: neat, thanks!

packages/automerge-doc-server/package.json
jmoggr: kinda/close enough. It was just development scar from a particularly painful problem. PR with the change is here: https://github.com/ToposInstitute/CatColab/pull/640

jmoggr: I've closed that PR. I'm currently having issues with something somewhere in wasm land and am not sure what the source issue is.

jmoggr: I had forgotten to rebuild catlog-wasm for the client, so the issue had nothing to do with my changes. I've just pushed the notebook-types wasm changes to this branch (currently it looks like there is more risk in juggling branches than in CI).

packages/automerge-doc-server/src/server.ts
jmoggr: Thank you!

packages/notebook-types/src/lib.rs
jmoggr: I changed the version field to a string. It seems that most system use strings for their versions and I think it's reasonable to make it more flexible in the future.

packages/catlog-wasm/src/model.rs
jmoggr: I feel conflicted about it. 

If we were to have a lot of access/mutation patterns for notebooks, especially if they were needed in both rust and TS, then I think it would make sense to keep it as "pure data" and have a cross platform library that we would use (something like NotebookUtils TS namespace but in rust).

If that is not the case, then I don't think it matters that much.

I'm not sure what the future looks like for these data structures, but it will be easy to change our strategy either way for the foreseeable future.

I implemented the helper method cause it was easy enough to do.

epatters: Agreed, we shouldn't overthink it at this point!

jmoggr:
I realised that I had an English parsing error on Owen's comment. I read:

> (We make a sum type) (over all versions with a .to_current() method) 

and he meant:
> (We make a sum type over all versions) (with a .to_current())

I updated the phrasing slightly to remove possible ambiguity (and learned something about restrictive and appositive phrases!)

Issue #184, opened 2024-09-25
Popup menu to change object/morphism type 


epatters:
It should be possible to change the type of an object or morphism without deleting it and recreating it, e.g., changing a link from positive to negative or changing an object from an entity to attribute type.

A nice interaction would be to click the object/morphism type on the right margin of the cell and get a popup menu with types that it can be changed to.

PR #604, opened 2025-07-21
Switching cell type from frontend

quffaro:
This feature lets users switch the cell type on a notebook.


epatters:
Closing this PR as stale. I would still love to have this feature though!

tim-at-topos:
should we open this as an issue? I think we can make it specific enough

epatters:
This feature is being tracked by #184.

Matt has also expressed interest in getting back to this one eventually.

PR #609, opened 2025-07-23
Modal Double Theory with State and Auxiliary Variables

quffaro:


epatters:
Just doing some house cleaning: closing due to staleness.

PR #615, opened 2025-07-24
New guides for the help pages

tim-at-topos:
- [ ] Fundamentals
- [ ] Predator-prey (CLDs)
- [ ] Employee structure (schemas)
- [ ] SEIRV (stock-flow diagrams)
- [ ] ??? (Petri nets)


epatters:
Hi @tim-at-topos, what's the status of this PR? For some reason there is a massive diff, so I can't easily tell what's going on.

tim-at-topos:
Ah the status is just that this got bumped down my to-do list in terms of priority, so it's currently stalled. There is essentially no content in this PR, I'd just set up the help page to have links in the correct places. We should discuss how much we think this is a good thing to be working on, and then either close this PR or I'll do a rebase and fix it.

Issue #589, opened 2025-07-15
Split theories in frontend into own modules


epatters:
The standard library of theories in the frontend (`src/stdlib/theories.tsx`) is getting long and it's only going to get longer. We should keep the theory metadata (`TheoryMeta`) in that file but move each theory definition into its own file. These can then be dynamically imported when needed. This will be an easy refactor because the theory content is already being created lazily.

To avoid painful merge conflicts, wait until at least #553 is merged.

tim-at-topos:
This sounds like fun! I'll have a look at it after #553

PR #616, opened 2025-07-24
Split frontend theories into individual files

tim-at-topos:
fix https://github.com/ToposInstitute/CatColab/issues/589


packages/frontend/src/stdlib/theories.tsx
```diff
@@ -20,13 +21,7 @@ stdTheories.add(
...
         help: true,
     },
-    (meta) => {
-        const thEmpty = new catlog.ThEmpty();
-        return new Theory({
-            ...meta,
-            theory: thEmpty.theory(),
-        });
-    },
+    createEmptyTheory,
```
epatters: To reduce boilerplate, we should `export default` from the individual theory modules and then dynamic import here:

```typescript
() => import("./theories/empty")
```

tim-at-topos: I *knew* there was a better way of doing this, so thank you for saving me 20 minutes of randomly paging through docs

tim-at-topos: hm, I am not familiar enough with dynamic imports to make this work: do you literally mean replacing `createEmptyTheory` with `() => import("./theories/empty")` (after adding `default` to the export)? this gives me an LSP error since the import is just a Promise

epatters: Ah dang, that was what I meant, but I forgot that dynamic imports are async. Come to think of it, it perhaps *ought* to be the case that fetching a theory from its ID is assumed to be async, because that opens the door to an eventual plugin architecture where theories can be loaded from sources besides the stdlib. However, async functions in JavaScript are viral in the sense that once a function becomes async, any consumers of that function must themselves become async.

So making this change will require changing other parts of the codebase. I *hope* this would be confined to changing [this `createMemo`](https://github.com/ToposInstitute/CatColab/blob/fce04080cd36c02867c254fcedf50384a5398c16/packages/frontend/src/model/document.ts#L100) to a [`createResource`](https://docs.solidjs.com/reference/basic-reactivity/create-resource), but without digging into it I couldn't be sure.

You could attempt this or, if it starts spiraling out of control, I could look at it more closely myself.

tim-at-topos: I'll see how it goes and let you know if (when) I get stuck

epatters:
@tim-at-topos, the commit above shows how to make loading theories async.

To fix the remaining errors, the frontend code will need to be updated to reflect that `liveModel.theory()` can now return `undefined`. That should be straightforward, mostly a matter of sprinkling `?` operators around, but I'll wait until after #603 lands to avoid merge conflicts (unlikely, I think, but possible).

tim-at-topos:
ok, sounds like you know what you're doing, so I'll assign this over to you — feel free to switch back to me when #603 lands if you like!

epatters (approved):
I pushed this a bit further than strictly necessary to achieve the immediate goal: the object/morphism cell editors for models/diagrams now receive the `Theory` as a prop, hence are only rendered when the theory is available. Not only does this save on some annoying `?` operators, it is ultimately more correct. If we don't know what theory we're in, we shouldn't just guess/fallback to defaults about how the objects/morphisms are supposed to appear. Moreover, in the future, the theory will even be able to select different widgets for the objects/morphisms.

PR #618, opened 2025-07-24
Reorganize dev and math docs

### LINKED ISSUE #605: Dev docs navigation and structure
The [dev docs](https://next.catcolab.org/dev/) currently have three sections:

1. API documentation
2. Mathematics
3. Technology

The first one is just a link to two other sites: `next.catcolab.org/dev/rust/catlog/` and `next.catcolab.org/dev/frontend/`. (Note that it's missing e.g. `catlog_wasm`). I think that documentation for software devs should live in places/formats that software devs generally tend to see documentation, and the forest is not such a place. I also worry that this will lead to some confusing mix where documentation that *should* be going in e.g. `next.catcolab.org/dev/rust/catlog` actually ends up in some tree in the forest, and then becomes split off from the auto-generated docs.

(The second one is nice albeit a bit confusing to navigate).

The third one is some notes that seem like they should actually be internal, simply because they talk about our current AWS setup and decisions that guided the initial selection of the tech stack, and neither of these are things that seem like they would be relevant to any external developer.

Things are further slightly confused by there also being developer documentation in the main README, rather than just having links. (And e.g. the frontend link there points to just the frontend README file instead of the autogenerated docs).

---

My actual suggestion is "we should have a proper discussion about this", but below are some simple suggestions to get a conversation started :-)

**Suggestions:**

- Have `next.catcolab.org/dev` point to some resource that links to the auto-generated package documentation/README pages (but ideally not both!) for all the packages
- Have `next.catcolab.org/maths` point to the forest, which would then contain just the "Mathematics" tree (and set up a redirect from `next.catcolab.org/math` 😉 )
- Move the "Technology" tree in the forest to somewhere else (perhaps the Topos forest?)

tim-at-topos:
fixes #605

- [x] create `CONTRIBUTING.md` and move all the stuff under the "API" tree in the forest there
- [x] render `CONTRIBUTING.md` to a nice webpage, and route `/dev` to it
- [x] make `/dev/rust` work as a link
- [ ] ~link back to `/dev/` from all rust packages without summaries~
- [x] move the "Technology > Infrastructure" content from the dev-docs forest into `/infrastructure/README.md`
- [x] remove everything else in the "Technology" tree
- [x] move the "Mathematics" content in the forest (which should now be the only thing left) to the top level
- [ ] ~add some pointers to make the mathematics forest easier to navigate~
- [ ] ~update the forester version (for pretty URLs)~
- [x] route `/math/` to the forest
- [x] remove `catcolab.org/dev` entirely (just have `next.catcolab.org/dev`)
- [x] remove `catcolab.org/math` entirely (just have `next.catcolab.org/math`)
- [x] update all the links in the help pages
- [x] update all the links in the main `README.md`
- [x] find any other places with links (welcome overlay? our website?) and update them


tim-at-topos:
Given that we now have three types of documentation for three types of user, namely:

- general user: `/help`
- developer: `/dev`
- mathematician: `/math`

I think it would be good to decide on who should be pointed to the stable release and who should be pointed to the staging release, so that we can fix all the links. My suggestion would be

- general user: `catcolab.org/help` or `next.catcolab.org/help`, depending on what they want
- developer: `next.catcolab.org/dev`
- mathematician: `next.catcolab.org/math`

and I think I would make this suggestion strict, in the sense that `/dev` and `/math` both redirect to the corresponding page on `next`. This way we just avoid people having possibly conflicting versions of documentation, and it seems reasonable to expect that (a) developers will be happy to read documentation on `next`, and (b) mathematicians will want not care much about the difference between mathematics documentation on the stable or staging branches.

Thoughts @epatters ?

epatters:
Completely agreed. Developers will want the latest API docs, and the distinction b/w prod and staging is irrelevant to mathematicians.

tim-at-topos:
I'm going to save any forester fiddling for a separate PR, because I think it should be part of a larger discussion about the structure/purpose/maintainance etc. of the _mathematical_ docs. However, this PR still changes the URL to `/math` instead of `/dev`, so it will at least semi-stabilise any links irrelevant of future choices about what we actually host at `/math`

tim-at-topos:
OK, ready for review @epatters ! This is largely just fixing some URLs, and then the rest is moving content around to avoid duplication or add context

| Staging vs Production | URL | description | 
| :-- | :-- | :-- |
| `catcolab.org` | `/help` | just as before: user help docs (from release) |
| | `/dev` | **does not exist: no longer created in `deploy.yml`** |
| | `/math` | **does not exist: no longer created in `deploy.yml`** |
| `next.catcolab.org` | `/help` | just as before: user help docs (from `main`) |
| | `/dev` | **now auto-generated from `CONTRIBUTING.md`, links to all other dev docs** |
| | `/math` | **contains just the mathematical part of the old dev-docs forest** |

The most critical thing to double check that I haven't broken is `deploy.yml`, which I've slightly reshuffled given that the consolidation step now also different between staging and production.

(Sorry for all the file changes, but you'll see that most of them are just from renaming `dev-docs` to `math-docs`)

epatters (changes requested):
Thanks Tim, this is a great re-org of the dev/math docs. Also impressed by your Markdown table summarizing the changes :)

My only comments concern the new folder `dev-docs`. The top-line item is that, for the sake of both documentation and testing, it should be as self-contained as possible. So:

- it should have a `package.json`, specifying the dependency `typedoc` (any others?)
- that file should include a `script` to build the page (currently only in CI)
- it should have a little README saying what to run
- the top-level CONTRIBUTING file should get copied somewhere that is git-ignored

.github/workflows/deploy.yml
```diff
@@ -15,8 +15,36 @@ permissions:
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: "pnpm"
+
+      - name: Build developer docs
+        run: |
+          pnpm install
+          mv CONTRIBUTING.md dev-docs/README.md
+          pnpm typedoc --options ./dev-docs/typedoc.json
```
epatters: To expand on the above, this would become something like:

```sh
pnpm install
pnpm --filter ./dev-docs run build
```

tim-at-topos: that's much tidier, will do!

PR #619, opened 2025-07-24
Stochastic Mass Action Dynamics

quffaro:


packages/catlog-wasm/src/theories.rs
```diff
@@ -335,6 +335,22 @@ impl ThSymMonoidalCategory {
                 .into(),
         ))
     }
+
+    /// Simulates a reaction network
+    #[wasm_bindgen(js_name = "reactionNetwork")]
+    pub fn reaction_network(
```
epatters: Should have a more descriptive name. I would suggest `stochastic_mass_action` / `stochasticMassAction`.

packages/catlog/Cargo.toml
```diff
@@ -16,11 +16,13 @@ derive_more = { version = "2", features = ["constructor", "from", "into"] }
 duplicate = "2"
 egglog = { version = "0.4", default-features = false }
 ego-tree = "0.10"
+getrandom = { version = "0.3", features = ["wasm_js"] }
```
epatters: You can remove this line. `getrandom` in both versions v0.2 and v0.3 should now be configured correctly in `catlog-wasm` on main.

packages/catlog/Cargo.toml
```diff
@@ -16,11 +16,13 @@ derive_more = { version = "2", features = ["constructor", "from", "into"] }
...
 itertools = "0.14"
 nalgebra = { version = "0.33", optional = true }
 nonempty = "0.12"
 num-traits = "0.2"
 ode_solvers = { version = "0.5.0", optional = true }
+rebop = { git = "https://github.com/Armavica/rebop.git", default-features = false, optional = true }
```
epatters: If we can't get an upstream release, we should pin to a specific commit for reproducibility.

packages/catlog/src/stdlib/analyses/ode/mass_action.rs
```diff
@@ -112,6 +116,87 @@ impl PetriNetMassActionAnalysis {
     ) -> ODEAnalysis<NumericalPolynomialSystem<u8>> {
         into_numerical_system(self.build_system(model), data)
     }
+
+    /// Creates a mass-action system in a reaction network.
+    pub fn build_reaction(
```
epatters: Would suggest a more descriptive name here too. Comparing with the names above, perhaps `build_stochastic_system`. Should also add the word "stochastic" in docstring.

packages/catlog/src/stdlib/analyses/ode/mass_action.rs
```diff
@@ -250,4 +335,17 @@ mod tests {
         "#]);
         expected.assert_eq(&sys.to_string());
     }
+
+    #[test]
+    fn sir_petri_dynamics() {
+        let th = Rc::new(th_sym_monoidal_category());
+        let model = sir_petri(th);
+        let sys = PetriNetMassActionAnalysis::default().build_system(&model);
```
epatters: Wait a second, this is testing the old method, not the new one! :)

packages/frontend/package.json
```diff
@@ -8,7 +8,7 @@
         "staging": "vite --mode staging",
         "dev": "vite --host",
         "build:wasm": "npm --prefix ../catlog-wasm run build:browser",
-        "build": "npm run build:wasm && tsc -b && vite build --sourcemap true",
+        "build": "RUSTFLAGS='--cfg getrandom_backend=\"wasm_js\"' npm run build:wasm && tsc -b && vite build --sourcemap true",
```
epatters: I've fixed the CI so that you can remove this too.

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -151,3 +151,131 @@ export function MassAction(
         </div>
     );
 }
+
+// type Simulator<...>
```
epatters: Delete stale line.

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -151,3 +151,131 @@ export function MassAction(
...
+export function configureReactionMassAction(options: {
```
epatters: Ditto about the name here and below.

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -151,3 +151,131 @@ export function MassAction(
...
+        initialContent: () => ({
+            rates: {},
+            initialValues: {},
+            duration: 10,
+        }),
+    };
+}
+
+/** Analyze a model using mass-action dynamics. */
+export function StochasticODEMassAction(
```
epatters: This whole function looks like a copy-paste of the existing [`MassAction`](https://github.com/ToposInstitute/CatColab/blob/9c10e5b63fa1dbe9be12c85fa4f29678c41038d5/packages/frontend/src/stdlib/analyses/mass_action.tsx#L59). Is it identical? Even if not, I would expect that with at most minor changes we could use the same component for both deterministic and stochastic simulations, passing only different `simulate` props.

packages/catlog/src/stdlib/analyses/ode/mass_action.rs
quffaro: ah you got me! this is now fixed.

epatters (approved):

PR #625, opened 2025-07-25
Allow vertical scrolling in theory selector

quffaro:
The list of theories is likely to grow over time, so this small PR adds scrolling on the theory selector when the screen height is small, like on floating windows (or mobile, for CatColabbators on the go).


epatters (approved):
Thanks Matt!

PR #635, opened 2025-07-28
Cobord/cleanup/clippy nits

Cobord:
Minor cleaning up, idiomatic


epatters:
Thanks for helping improve the quality of the codebase! A few questions/comments:

- What `clippy` settings are you using to detect these issues? On what I guess is the standard clippy settings, our CI is already enforcing that the linter passes without warnings.
- It seems that at least some of these features are unstable. E.g., [`must_use`](https://doc.rust-lang.org/beta/std/hint/fn.must_use.html) seems to be an unstable feature (first time I've encountered it). So far at least we've been avoiding unstable features: we aren't trying to beta test the Rust compiler :)

Cobord:
A while ago I got burned by a change on mutating a morphism to do
post-composition with another and then a couple weeks later deciding to
return something as well. So I have been extra with must_use until I have
understood how everything fits together to avoid that problem again.

epatters:
Thanks for the explanation! I found this page useful, so noting it here for others: https://std-dev-guide.rust-lang.org/policy/must-use.html

olynch:
Rather than reviewing all of these changes, I'd much prefer to review either:

1. A diff to an informal code policy document, which adds guidelines that when followed would produce these changes, and a rational for following these guidelines.
2. A diff to clippy settings which would warn when these changes weren't made.

Then we can discuss which of these high-level style changes are principles we want to follow for the CatColab code base, and then the high-level style changes that we like can be kept.

PR #636, opened 2025-07-28
Revert to building catcolab-next deployments on AWS

jmoggr:


epatters (approved):
Sorry this turned out to be so painful!

Issue #579, opened 2025-07-11
Drag-and-drop failure for rich text cells seem to duplicate their content on the UI


quffaro:
![Image](https://github.com/user-attachments/assets/731691c6-3dce-4143-9da4-5a0e6c874ef2)

If a rich text cell fails to drop*, it will sometimes duplicate its content. This appears to be a UI bug, since the component will return to its correct content after a successful drag-and-drop.

* the issue where cells fail to drop are tracked in #205

ShixuLuo:
Hi, after some investigation into this issue, here’s what I’ve found:

1. This issue occurs at an early stage and is reproducible on the current version of [catcolab.org](http://catcolab.org/).

2. The problem originates in `frontend/src/notebook/notebook_editor.tsx`, specifically in the `onDrop` handler of `NotebookEditor`.
If the computed `finalIndex` equals `sourceIndex`, the following update logic may trigger an unnecessary Automerge update:

```tsx
props.changeNotebook((nb) => {
    const [cell] = nb.cells.splice(sourceIndex, 1);
    nb.cells.splice(finalIndex, 0, deepCopyJSON(cell));
});
```
While I’m not deeply familiar with Automerge, this might explain the duplicated state across sessions.

3. In PR #637 , I implemented a minimal fix: disallowing a cell to be dropped onto itself. This causes the function to exit early:

```tsx
if (!(target && isCellDragData(source.data) && isCellDragData(target.data))) {
    setTether(null);
    setCurrentDropTarget(null);
    return;
}
````
This workaround is simple but intuitive to me—dropping a cell onto itself doesn't seem like a meaningful action anyway.

As further evidence this is an Automerge-related issue: when I reproduced the bug in one window, the duplicated content also appeared in a second window viewing the same document.

Let me know if a deeper fix is preferred—I’m happy to iterate.

**Comment by jmoggr** [comment]
I agree that something odd is going on with automerge and that the source of the issues is this:
```
    const [cell] = nb.cells.splice(sourceIndex, 1);
    nb.cells.splice(finalIndex, 0, deepCopyJSON(cell));
```

Automerge works using proxy objects and records changes using paths in JSON tree. I suspect that using `deepCopyJSON` combined with two updates on the same path is causing some sort of weird but technically correct edge case. I personally don't think it's worth investigating further.

This issues has already been indirectly fixed by #603 and #641, where we change to tracking the cell order by index only. The relevant changes are here:

[...]

PR #637, opened 2025-07-28
fix(notebook): prevent cell drop onto itself

ShixuLuo:
Fixes issue #579 

Dragging a cell onto itself causes a UI update, likely due to an internal Automerge change, even though the actual data remains the same.

This fix resolves the problem by preventing cells from being dropped onto themselves.

### DIFF
--- a/packages/frontend/src/notebook/notebook_cell.tsx
+++ b/packages/frontend/src/notebook/notebook_cell.tsx
@@ -19,6 +19,7 @@ import {
 } from "../components";
 
 import ArrowDown from "lucide-solid/icons/arrow-down";
+import Copy from "lucide-solid/icons/copy";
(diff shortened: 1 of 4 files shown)


jmoggr:
Thanks for the PR!

Could you please rebase this onto #641? It includes some complementary changes to the same code. Context for this request can be found here: https://github.com/ToposInstitute/CatColab/issues/579#issuecomment-3137087343

ShixuLuo:
I’ve rebased my fix onto #641 as requested.

I wasn’t totally sure which branch this PR should ultimately target — whether it’s `main` or `notebook-type-migrations` — so I went ahead and changed the base to `notebook-type-migrations` for now.
Let me know if I should change it back or adjust anything else!

epatters (approved):
Thanks for your careful analysis of this situation and for the fix!

PR #639, opened 2025-07-29
Refactor notebook accessors and mutators into separate namespace

jmoggr:
This removes the need for lint overrides, de-duplicates logic, and addresses feedback about adding utility functions. However it might be more "overkill" than desired.

I'm in favour of this pattern, it keeps logic out of UI code and provides a library of utility functions that can easily be re-used and extended.

Possible lesson learned:  this was worth effort of doing from the beginning.


packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -249,80 +249,65 @@ export function NotebookEditor<T>(props: {
...
+                                    NotebookUtils.tryMoveCellDown(nb, i());
                                 });
                             },
                             hasFocused() {
                                 setActiveCell(i());
                             },
                         };
 
+                        const cell = props.notebook.cellContents[cellId];
+                        invariant(cell, `Failed to find contents for cell '${cellId}'`);
```
jmoggr, line 309: I moved this from the top of the closure, it's not needed by `cellActions` and using this instance of `cell` in `cellActions` would likely result in bugs, since this `cell` is not in a automerge change context.

packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -249,80 +249,65 @@ export function NotebookEditor<T>(props: {
...
-                        const cell = props.notebook.cellContents[cellId];
-                        if (!cell) {
-                            throw `Failed to find contents for cell ${cellId}`;
-                        }
-
                         const isActive = () => activeCell() === i();
+
                         const cellActions: CellActions = {
                             activateAbove() {
-                                i() > 0 && setActiveCell(i() - 1);
```
jmoggr, line 260: This is bad code smell IMO. Principle of least surprise suggests that people would expect to see control common control flow statements where control flow is happening, and would not expect to see short circuiting outside of assignment or conditionals.

epatters: Agreed (even though I probably wrote this code :). Short-circuiting logical operators are a cute way to save a line or two, but better to make control flow explicit.

packages/frontend/src/notebook/notebook_editor.tsx
epatters: Good catch, I missed that!

epatters (approved):
Thanks Jason, this is great: much cleaner than what we had before!

PR #643, opened 2025-07-31
FIX: cached documents not being migrated on the client

jmoggr:
Cut and paste migration solution from `automerge-doc-server`. This is not ideal, but we currently don't have a ts package to share code between automerge and the frontend and for the sake of CI stability and my own sanity I'd rather not add one at this moment.

I'm personally okay with letting this copy-and-paste persist, since there is only one instance of it and it has a comment. However I could create an issue to track this if we'd like to address it the future.


jmoggr:
There is some sort of bad interaction between client migrations and the development server

If I do:
- open tab with localhost:5173/model/ref
- stop development server
- apply the code changes notebook migrations
- start the development server
- (open tab automatically reloads)

Then the snapshot in the db becomes temporarily corrupted.

If I do:
- close tab with localhost:5173/model/ref
- stop the frontend development server
- apply the code changes for notebook migrations
- start the development server
- open a new tab with localhost:5173/model/ref

Then everything works as expected.

I suspect that the way the client development site is being reloaded is interacting badly with the automerge sync when migrations are being run on both the client and the server.

Based on my understanding of how automerge works I see no reason why running a migration on a client and server simultaneously should lead to issues (though we should maybe ask?).

Given that the issue only seems to happen in the development context, I feel comfortable moving forward with running the migrations on the client as well.


Example of the snapshot corruption:
```
      "019860ae-0a1d-717a-9124-d3236dd0a20e": {
        "id": "019860ae-0a1d-717a-9124-d3236dd0a20e019860ae-0a1d-717a-9124-d3236dd0a20e",
        "tag": "rich-textrich-text",
        "content": "oneone"
      },
```

It seems that the corruption goes away as soon as any change is applied (or some other series of events?):
```
      "019860ae-0a1d-717a-9124-d3236dd0a20e": {
        "id": "019860ae-0a1d-717a-9124-d3236dd0a20e",
        "tag": "rich-text",
        "content": "onettt"
      },
```

PR #646, opened 2025-07-31
Replace `Theory` by `TheoryMeta` in `TheoryHelpButton`

### LINKED ISSUE #611: `TheoryHelpButton` uses `Theory` and `TheoryMeta`
Currently `TheoryHelpButton` takes either optional props of type `Theory` *and* type `TheoryMeta`. This is a hack arising from moving the help button from the main toolbar at the top of the page to next to the logic name in the logic selector. I don't think that the help button needs access to the actual `Theory`, so just rewriting this to only accept a `TheoryMeta` should suffice. As part of this, it might be sensible to write a function `theoryFromMeta: TheoryMeta -> Theory`.

https://github.com/ToposInstitute/CatColab/blob/c8ca5c359eb677117b5f03c63061cf25aced239a/packages/frontend/src/page/toolbar.tsx#L54

tim-at-topos:
fixes #611

### DIFF
--- a/packages/frontend/src/model/theory_selector.tsx
+++ b/packages/frontend/src/model/theory_selector.tsx
@@ -25,7 +25,7 @@ export function TheorySelectorDialog(props: TheorySelectorProps) {
             >
                 {props.theoryMeta.name}
             </Dialog.Trigger>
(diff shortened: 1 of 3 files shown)


epatters (approved):
Thanks for cleaning this up!

PR #648, opened 2025-07-31
Proptests of categorical data structures

Cobord:
Specifically proptest


Cobord:
Put in zig-zags and simplex/opposite of simplex (truncated with small n of course). My initial thought was about these being more extensive tests than graphs,sgraphs,hgraphs with many more objects, morphisms and equations. But they could be useful in other contexts as well.

epatters:
Thanks for your interest in the project. While we might well benefit from having more tests in certain places, tests like any other kind of code must be maintained and so are not without cost. I haven't reviewed the PR carefully, but my impression is that some of these tests are not actually increasing the coverage of the test suite.

In general, if you're considering proposing significant code changes or additions, I'd encourage you to reach out to the dev first (say by filing an issue or posting the Zulip) to ensure that there will be a path to getting the changes merged.

PR #649, opened 2025-07-31
Issue template for bug reports

tim-at-topos:
just trying to make the number at https://github.com/ToposInstitute/CatColab/community go up


epatters (approved):
Yay, number go up! ;)

I've sometimes found issue and PR templates to be annoyingly heavy-handed, but this one looks pretty light-weight. Happy to try it out for a while and see if we like it.

PR #651, opened 2025-07-31
Fix title of model being needlessly cut off

epatters:
It is currently being cut off when it exceeds ~50% of the available space.

cc @tim-at-topos since this was introduced in #553. Not sure what was the intended effect, but feel free to have another go if there's something you're after.


tim-at-topos:
oh woops, this was me trying to get the question mark button to render nicely but i'm so old that all my CSS knowledge predates flexboxes 👻

PR #655, opened 2025-08-01
Fix nodes+marks being possibly undefined in our PM schema

jmoggr:
I added a comment giving some background on the types we use for our schema.

Something in the library types changed that caused our types to become less defined again.

### DIFF
--- a/packages/frontend/src/components/rich_text_editor/basic_schema.ts
+++ b/packages/frontend/src/components/rich_text_editor/basic_schema.ts
@@ -22,7 +22,7 @@ const olDOM: DOMOutputSpec = ["ol", 0];
 const ulDOM: DOMOutputSpec = ["ul", 0];
 const liDOM: DOMOutputSpec = ["li", 0];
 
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/rich_text_editor/basic_schema.ts
```diff
@@ -348,4 +348,4 @@ export const basicSchema: MappedSchemaSpec = {
             },
         } as MarkSpec,
     },
-};
+} satisfies MappedSchemaSpec;
```
jmoggr, line 351: This fixes the problem. It should have been this way since the start, I'm not quite sure why our types worked before this.

epatters: Sorry, I changed that in my PR, because I thought it fixed type errors I was seeing but I think it only appeared to do so due to other issues.

epatters (approved):
Thanks for adding the further explanation!

PR #663, opened 2025-08-05
Warn when docs are missing in `catlog-wasm`

epatters:
The CI will now fail when public structs, functions, etc in `catlog-wasm` have no docstrings, like it already does for `catlog`.


tim-at-topos:
wonderful, thank you!

PR #664, opened 2025-08-05
Extended causal loop diagrams

tim-at-topos:
Implements a new logic of extended causal loop diagrams: causal loop diagrams labelled with differential degree and order of delay (both natural numbers).

This PR also adds a pre-processing function (`catlog/src/stdlib/analyses/ecld/atomisations.rs`) for future analyses, which is currently unused but included simply because it's already written (along with a simple test). If preferred, this could be moved to a separate PR :-)

To-do:
- [x] `/math-docs` documentation (@jrb137)
- [x] `/help` documentation (@tim-at-topos)

### DIFF
--- a/math-docs/trees/dct-000C.tree
+++ b/math-docs/trees/dct-000C.tree
@@ -16,4 +16,4 @@ monoid #{\mathbb{N} \times \mathbb{N} \times \mathbb{Z}_2} with the monoidal pro
 \transclude{thy-000C}
 
 \p{A free model of this double theory a causal loop diagram in which edges can
-are additionally marked with two natural numbers indicating the \em{degree} and \em{delay} of the influence, called an ]em{extended causal loop diagram}.}
+are additionally marked with two natural numbers indicating the \em{degree} and \em{delay} of the influence, called an \em{extended causal loop diagram}.}
--- a/math-docs/trees/thy-000C.tree
+++ b/math-docs/trees/thy-000C.tree
@@ -2,6 +2,6 @@
 \taxon{theory}
 \import{macros}
 
-\p{The \define{theory of degree-delay signed categories} is the simple double
+\p{The \define{theory of degree-delay signed categories} is the discrete double
 theory with a single object and with proarrows given by the elements of the
 monoid #{\mathbb{N} \times \mathbb{N} \times \mathbb{Z}_2}.}
--- a/packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
+++ b/packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
@@ -20,9 +20,9 @@ and call Y the *base* of this tower; we call the length n of the list
  */
 
 use crate::dbl::model::{DiscreteDblModel, FgDblModel, MutDblModel};
-use crate::one::{Path, category::FgCategory};
+use crate::one::{category::FgCategory, Path};
 use crate::stdlib::theories;
-use crate::zero::{QualifiedName, name};
+use crate::zero::{name, QualifiedName};
 use std::{collections::HashMap, rc::Rc};
 
 // Some helpful functions
@@ -31,6 +31,7 @@ fn deg_of_mor(model: &DiscreteDblModel, f: &QualifiedName) -> usize {
 }
 
 fn is_mor_pos(model: &DiscreteDblModel, f: &QualifiedName) -> bool {
+    // TO-DO: use th_deg_del_signed_category_to_signed_category()
     0 == model
         .mor_generator_type(f)
         .into_iter()
@@ -45,8 +46,7 @@ fn is_mor_pos(model: &DiscreteDblModel, f: &QualifiedName) -> bool {
  * track of the relation between the formal derivatives and the original objects
  */
 pub fn degree_atomisation(
-    // TODO: do we really need an Rc here?
-    model: Rc<DiscreteDblModel>,
+    model: DiscreteDblModel,
 ) -> (DiscreteDblModel, HashMap<QualifiedName, Vec<QualifiedName>>) {
     let mut atomised_model: DiscreteDblModel =
         DiscreteDblModel::new(Rc::new(theories::th_deg_del_signed_category()));
@@ -65,7 +65,7 @@ pub fn degree_atomisation(
     // an empty list of positive-degree (resp. zero-degree) incoming arrows
     for x in model.ob_generators() {
         tower_constructors.insert(
-            x.clone(),
+            x,
             TowerConstructor {
                 height: 1,
                 in_arrows_pos_deg: Vec::new(),
@@ -80,21 +80,17 @@ pub fn degree_atomisation(
     //      over the *morphisms* instead.
 
     for f in model.mor_generators() {
-        let f_cod = model.get_cod(&f).unwrap();
+        let f_cod = model.mor_generator_cod(&f);
         let f_degree = deg_of_mor(&model, &f);
 
         if f_degree != 0 {
-            let new_degree = std::cmp::max(tower_constructors.get(f_cod).unwrap().height, f_degree);
-            tower_constructors
-                .entry(f_cod.clone())
-                .and_modify(|tower_cons| tower_cons.height = new_degree)
-                // Since we already have the codomain, we can add this to our
-                // hash map of positive-degree incoming arrows
-                .and_modify(|tower_cons| tower_cons.in_arrows_pos_deg.push(f.clone()));
+            let tower_cons = tower_constructors.get_mut(&f_cod).unwrap();
+            let new_degree = std::cmp::max(tower_cons.height, f_degree);
+            tower_cons.height = new_degree;
+            tower_cons.in_arrows_pos_deg.push(f.clone());
         } else {
-            tower_constructors
-                .entry(f_cod.clone())
-                .and_modify(|tower_cons| tower_cons.in_arrows_zero_deg.push(f.clone()));
+            let tower_cons = tower_constructors.get_mut(&f_cod).unwrap();
+            tower_cons.in_arrows_zero_deg.push(f.clone());
         }
     }
 
@@ -110,7 +106,7 @@ pub fn degree_atomisation(
 
     // We have yet to build any of the towers so, right now, every base is
     // "unchecked".
-    let mut unchecked_bases: Vec<QualifiedName> = model.ob_generators().collect::<Vec<_>>().clone();
+    let mut unchecked_bases: Vec<QualifiedName> = model.ob_generators().collect::<Vec<_>>();
 
     while !unchecked_bases.is_empty() {
         // Since heights will change as we go, we start by resorting the list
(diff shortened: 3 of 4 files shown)


tim-at-topos:
@jrb137 I think that `catlog/src/stdlib/analyses/ecld/atomisations.rs` is now ready for us read through and review, and then we can open this PR for proper review :-)

.rustfmt.toml
```diff
@@ -1,3 +1,4 @@
 array_width = 80
 chain_width = 80
 fn_call_width = 80
+max_width = 100
```
epatters, line 4: Just noting (no action required) that this is the default value for `max_width`, but there's no harm in setting it explicitly.

math-docs/trees/dct-000C.tree
```diff
@@ -0,0 +1,19 @@
...
+
+\p{A \define{degree-delay signed category} is a category sliced over the product
+monoid #{\mathbb{N} \times \mathbb{N} \times \mathbb{Z}_2} with the monoidal product given by addition.}
+
+}
+
+\transclude{thy-000C}
+
+\p{A free model of this double theory a causal loop diagram in which edges can
+are additionally marked with two natural numbers indicating the \em{degree} and \em{delay} of the influence, called an ]em{extended causal loop diagram}.}
```
epatters: Tree is not building due to syntax error here.

The CI ought to catch this but apparently doesn't.

math-docs/trees/dct-000C.tree
```diff
@@ -0,0 +1,19 @@
+\title{Degree-delay signed categories}
+\taxon{doctrine}
+\import{macros}
+
+\p{Degree-delay signed categories are categories graded in the monoid #{\mathbb{N} \times \mathbb{N} \times \mathbb{Z}_2}. The two copies of #{\mathbb{N}} label edges by the \em{differential degree} and \em{order of delay}. The order of the differential degree is meant to indicate which derivative of the target is affected (linearly) by the source. For example, an edge #{A \to B} with degree 2 indicates a contribution of the form #{\ddot{B} \mathrel{+}= kA} for some #{k \in \mathbb{R}}. The \em{order of delay} only has meaning with the assumption that the delays are modelled stochastically by Erlang distributions with the same time-scale parameter (see, for example, [here](https://en.wikipedia.org/wiki/Erlang_distribution)). The labelling in #{\mathbb{Z}_2} indicates the "sign" of the influence, as described in \ref{dct-0002}.}
```
epatters, line 5: Perhaps worth explaining that the *reason* you're talking about the Erlang distribution is that it's a sum of i.i.d. Gammas, where the number of summands is the "order of delay" in the ECLD.

math-docs/trees/thy-000C.tree
```diff
@@ -0,0 +1,7 @@
+\title{Theory of degree-delay signed categories}
+\taxon{theory}
+\import{macros}
+
+\p{The \define{theory of degree-delay signed categories} is the simple double
```
epatters: "simple double theory" -> "discrete double theory"

packages/catlog/src/stdlib/analyses/ecld/mod.rs
```diff
@@ -0,0 +1,4 @@
+//! Utilities for analyses of ECLDs
```
epatters, line 1: Are you planning to add more submodules here? If not, I'd suggest removing the extra directory level and having just `stdlib/analyses/ecld.rs`.

packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
```diff
@@ -0,0 +1,292 @@
...
+}
+
+fn is_mor_pos(model: &DiscreteDblModel, f: &QualifiedName) -> bool {
+    0 == model
+        .mor_generator_type(f)
+        .into_iter()
+        .filter(|t| *t == name("Negative"))
+        .count()
+        % 2
+}
```
epatters, line 41: I feel like we ought to be doing such reasoning using general machinery rather than ad hoc calculations. However, the procedure that comes to mind, namely

1. Project onto the monoid of signs
2. Check whether the result is equal (in the e-graph) to the generator Negative

is almost certainly more code, so perhaps not worth the additional complexity.

epatters: Come to think of it, to enable the expected model migrations, we're going to have to define the needed theory map anyway, similar to [this one](https://github.com/ToposInstitute/CatColab/blob/49b3f4d9244d95e5388a14f28314e987d6d47161/packages/catlog/src/stdlib/theory_morphisms.rs#L40), so maybe it *is* worth it after all.

packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
```diff
@@ -0,0 +1,292 @@
...
+
+/** Atomisiation of an ECLD by degree: replace every degree-n arrow by a path
+ * of n-many degree-1 arrows, going via (n-1)-many new objects; all the
+ * degree-0 arrows are kept unchanged. Returns the derivative towers to keep
+ * track of the relation between the formal derivatives and the original objects
+ */
+pub fn degree_atomisation(
+    // TODO: do we really need an Rc here?
+    model: Rc<DiscreteDblModel>,
```
epatters: No, it should suffice to take `model: &DiscreteDblModel`.

packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
```diff
@@ -0,0 +1,292 @@
...
+        in_arrows_pos_deg: Vec<QualifiedName>,
+        in_arrows_zero_deg: Vec<QualifiedName>,
+    }
+    let mut tower_constructors: HashMap<QualifiedName, TowerConstructor> = HashMap::new();
+
+    // Every tower will be of height at least 1, and will have at the very least
+    // an empty list of positive-degree (resp. zero-degree) incoming arrows
+    for x in model.ob_generators() {
+        tower_constructors.insert(
+            x.clone(),
```
epatters: Unnecessary clone.

packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
```diff
@@ -0,0 +1,292 @@
...
+        );
+    }
+
+    // -------------------------------------------------------------------------
+    // 1.   For each base, calculate the maximum degree over all incoming
+    //      arrows. Note that this is most easily done by actually iterating
+    //      over the *morphisms* instead.
+
+    for f in model.mor_generators() {
+        let f_cod = model.get_cod(&f).unwrap();
```
epatters: Use `model.mor_generator_cod` instead to avoid `unwrap`.

packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
```diff
@@ -0,0 +1,292 @@
...
+        let f_degree = deg_of_mor(&model, &f);
+
+        if f_degree != 0 {
+            let new_degree = std::cmp::max(tower_constructors.get(f_cod).unwrap().height, f_degree);
+            tower_constructors
```
epatters: Simpler and more efficient to do:

```rust
let mut tower_cons = tower_constructors.get_mut(f_cod).unwrap();
tower_cons.height = ...
```

packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
```diff
@@ -0,0 +1,292 @@
...
+    //      the incoming arrows can be lifted so that their target is the
+    //      highest floor of the tower, i.e. that the tower over Y has enough
+    //      floors so that *every* arrow into Y can have the same target Y_max.
+    //      But note that if we lift an arrow of degree n from X to Y then we
+    //      need to know that X_n exists, i.e. we need to (potentially) add
+    //      more floors to the *source* of every arrow whose target is Y.
+
+    // We have yet to build any of the towers so, right now, every base is
+    // "unchecked".
+    let mut unchecked_bases: Vec<QualifiedName> = model.ob_generators().collect::<Vec<_>>().clone();
```
epatters: Never need to `clone` immediately after `collect`: you're already getting a fresh object.

packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
```diff
@@ -0,0 +1,292 @@
...
+                height: 1,
+                in_arrows_pos_deg: Vec::new(),
+                in_arrows_zero_deg: Vec::new(),
+            },
+        );
+    }
+
+    // -------------------------------------------------------------------------
+    // 1.   For each base, calculate the maximum degree over all incoming
+    //      arrows. Note that this is most easily done by actually iterating
```
epatters, line 79: Is the comment out of date? I don't see a calculation of the maximum degree.

packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
```diff
@@ -0,0 +1,292 @@
...
+        for f in &tower_cons.in_arrows_pos_deg {
+            let deg = deg_of_mor(&model, f);
+            let source = model.get_dom(f).unwrap();
+            let source_tower = towers.get(source).unwrap();
+            // Note that we could alternatively take height to be the length of
+            // towers.get(source), which is equal by construction/definition
+            let height = tower_cons.height;
+            let new_source = &source_tower[height - deg];
+            let new_target = tower.last().unwrap();
+            match is_mor_pos(&model, f) {
```
epatters: Use `if`-`else` instead of matching on a boolean.

packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
```diff
@@ -0,0 +1,292 @@
...
+    use std::{collections::HashMap, rc::Rc};
+
+    // Makes a hash map from objects in sample_ecld to tower heights
+    fn correct_heights() -> HashMap<QualifiedName, usize> {
+        let mut heights = HashMap::new();
+        heights.insert(name("a"), 1);
+        heights.insert(name("b"), 3);
+        heights.insert(name("c"), 3);
+        heights.insert(name("d"), 2);
+        heights
```
epatters, line 238: Idiomatic way to write this in one line is:

```rust
[(name("a"), 1), (name("b"), 3), ...].collect()
```

packages/catlog/src/stdlib/analyses/ecld/atomisations.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    // Makes a hash map from morphisms in sample_ecld to the correct index of
+    // the domain in the atomised tower
+    fn correct_domain_indices() -> HashMap<QualifiedName, usize> {
+        let mut domains = HashMap::new();
+        domains.insert(name("f"), 0);
+        domains.insert(name("g"), 2);
+        domains.insert(name("l"), 0);
+        domains
```
epatters, line 251: Ditto.

epatters (changes requested):

epatters:
@tim-at-topos, this PR has become pretty stale, unfortunately. Is this something that you'd still like to push forward? One thing that gives me pause is that I still don't really understand atomization or why it needs to be so complicated, though I'm sure there's a reason.

tim-at-topos:
atomization is complicated for simple reasons, but really this PR should be preceded by an RFC! feel free to close this and I'll write an RFC

epatters:
Ok, thanks! Happy to read such a document if/when you find the time for it.

Issue #376, opened 2025-02-04
Unclear that you can't edit a view-only model


KevinDCarlson:
The only feedback the user gets about trying to edit a view-only model (which is possible at the client side, it just doesn't propagate to the database!) is this symbol:

<img width="30" alt="Image" src="https://github.com/user-attachments/assets/76bee1bb-b25f-4f0d-8eca-bdccc834e858" />

epatters:
If you hover over that icon, you get a tooltip explaining that you can't edit, but I suppose it would also be useful to allow the icon to be clicked.

Issue #417, opened 2025-02-19
Refactor the Live[*]Document types so that there is a shared LiveDocument base type


jmoggr:
Currently the Live[*]Document types share many similarities but are created in an adhoc manner. This can lead to an underdefined and implicit interface for the union of Live[*]Document types. This can be solved by making the Live[*]Document types extend as shared  LiveDocument base type.

https://github.com/ToposInstitute/CatColab/pull/407#discussion_r1961028830

epatters:
I decided to take a different approach to this in #744. See the PR description for more info.

PR #670, opened 2025-08-06
Add dialog with suggestion to duplicate view-only model

kasbah:
Partly addresses #376. I think the model/doc should also auto-duplicate when the user starts editing it.

image

### DIFF
--- a/packages/frontend/src/api/duplicate_document.ts
+++ b/packages/frontend/src/api/duplicate_document.ts
@@ -1,0 +1,36 @@
+import { type LiveDiagramDocument, createDiagramFromDocument } from "../diagram/document";
+import { type LiveModelDocument, createModel } from "../model/document";
+import { assertExhaustive } from "../util/assert_exhaustive";
+import type { Api } from "./";
+
+/**
+ * Duplicates a document (diagram or model) and returns the new reference ID.
+ * The duplicated document will have " (copy)" appended to its name.
+ */
+export async function duplicateDocument(
+    api: Api,
+    liveDocument: LiveDiagramDocument | LiveModelDocument,
+): Promise<string> {
+    if (!liveDocument) {
+        throw new Error("Cannot duplicate: liveDocument not provided");
+    }
+
+    switch (liveDocument.type) {
+        case "diagram": {
+            const diagram = liveDocument.liveDoc.doc;
+            return createDiagramFromDocument(api, {
+                ...diagram,
+                name: `${diagram.name} (copy)`,
+            });
+        }
+        case "model": {
+            const model = liveDocument.liveDoc.doc;
+            return createModel(api, {
+                ...model,
+                name: `${model.name} (copy)`,
+            });
+        }
+        default:
+            assertExhaustive(liveDocument);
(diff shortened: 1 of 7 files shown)


packages/frontend/src/page/document_menu.tsx
```diff
@@ -66,28 +63,8 @@ export function DocumentMenu(props: {
...
-                const newRef = await createModel(api, {
-                    ...model,
-                    name: `${model.name} (copy)`,
-                });
-                navigate(`/model/${newRef}`);
-                break;
-            }
-            default:
-                assertExhaustive(props.liveDocument);
-        }
```
epatters, line 87: Thanks for relocating this logic so that it's not buried in a handler and can be reused.

packages/frontend/src/user/permissions.tsx
```diff
@@ -272,17 +278,59 @@ const AnonPermissionsTrigger = (props: ComponentProps<"button">) => {
...
+    const [open, setOpen] = createSignal(false);
+    const api = useApi();
+    const navigate = useNavigate();
+
+    const onDuplicateDocument = async () => {
+        try {
+            const newRef = await duplicateDocument(api, props.liveDocument);
+            navigate(`/${props.liveDocument.type}/${newRef}`);
+        } catch (error) {
+            console.error("Failed to duplicate document:", error);
```
epatters: For now, our error handling strategy is to let them propagate and be caught and display by a top-level error boundary. So let's remove this `catch` block so that failures to duplicate (or navigate) aren't silently swallowed.

packages/frontend/src/api/duplicate_document.ts
```diff
@@ -0,0 +1,36 @@
...
+import { assertExhaustive } from "../util/assert_exhaustive";
+import type { Api } from "./";
+
+/**
+ * Duplicates a document (diagram or model) and returns the new reference ID.
+ * The duplicated document will have " (copy)" appended to its name.
+ */
+export async function duplicateDocument(
+    api: Api,
+    liveDocument: LiveDiagramDocument | LiveModelDocument,
```
epatters, line 12: Note to self (no action required): we've now got several places where we have to consider a union over document types or live document types; here's [another](https://github.com/ToposInstitute/CatColab/blob/fd55208f02cf12231a9090a1080ba8e89e71efe3/packages/frontend/src/page/document_breadcrumbs.tsx#L12). In a future PR, we should implement #417 and consolidate this kind of logic.

epatters (changes requested):
Thank you, this is very clean! I just had one small request (plus two comments in passing).

packages/frontend/src/user/permissions.tsx
kasbah: Ok, I've removed the try/catch.

kasbah:
Thanks for the comments and explanations.

packages/frontend/src/api/duplicate_document.ts
epatters: I took a different approach to this in #744.

PR #687, opened 2025-08-14
Fix poor performance of Jupyter kernels in the frontend

jmoggr:


packages/frontend/src/stdlib/analyses/decapodes.tsx
```diff
@@ -352,14 +352,18 @@ const makeInitCode = () =>
...
     system = Analysis(ThDecapode(), raw"""${JSON.stringify(data)}""");
     simulator = evalsim(system.pode);
 
     f = simulator(system.geometry.dualmesh, system.generate, DiagonalHodge());
 
     soln = run_sim(f, system.init, system.duration, ComponentArray(k=0.5,));
 
-    JsonValue(SimResult(soln, system))
+    result = JSON3.write(SimResult(soln, system))
+    println(result)
```
epatters: It's slighter shorter, and also likely slightly more efficient, to replace these two lines with:

```julia
JSON3.write(stdout, SimResult(soln, system))
```

packages/algjulia-interop/jupyter_server.sh
```diff
@@ -53,4 +53,5 @@ jupyter server \
     --ServerApp.disable_check_xsrf=True \
     --ServerApp.allow_origin="$ORIGIN" \
     --ServerApp.allow_credentials=True \
+    --ServerApp.iopub_data_rate_limit=1000000000 \
```
epatters, line 56: Just noting @quffaro that this flag will need to be carried over to your branch replacing the bash script with a Julia script.

packages/algjulia-interop/src/kernel_support.jl
```diff
@@ -2,6 +2,26 @@ import JSON3
...
+# performance, but one might be worse than the other. Whether or not the performance is different might
+# help inform a future investigation.
+#
+# NOTE: the use of the `show` methods is the only way I was able to get the contents of the message
+# (`content["data"]?.["application/json"]`) to be interpreted as JSON in the browser. using something
+# like `display(MIME"application/json"(), JSON3.write(...))` caused the
+# content to be the JSON in string form, even though the mime type was "application/json"
+#
+# NOTE: It looks like IJulia is delivering payloads for each mime type available according to this
+# docstring https://github.com/JuliaLang/IJulia.jl/blob/master/src/display.jl
```
epatters, line 18: Pursuant to this comment, we should probably remove the `Base.show` method below for the `text/plain` MIME type.

packages/frontend/src/stdlib/analyses/jupyter.ts
```diff
@@ -21,18 +22,51 @@ export function createKernel(
...
+        // Useful for debugging jupyter stuff
+        const t0 = Date.now();
+        kernel.anyMessage.connect((_, msg) => {
+            console.log(
+                "[Jupyter message]",
+                (Date.now() - t0) / 1000,
+                msg.direction,
+                msg.msg.header.msg_type,
+            );
+        });
```
epatters: Is this intended to stay here or were you planning to delete/comment it except when debugging?

packages/frontend/src/stdlib/analyses/jupyter.ts
```diff
@@ -71,25 +105,53 @@ export function executeAndRetrieve<S, T>(
...
+                // biome-ignore lint/suspicious/noExplicitAny: this should type narrow, but doesn't
+                const text = (msg.content as any).text;
+                if (text) {
+                    streamedResult += text;
+                }
+            }
+
+            // In this case msg_type "stream" will be collected but not used. This only works for
+            // relatively small amounts of data, it could be the case that nothing should be returned
+            // like this. The json seriailization done by IJulia appears to be cripillingly slow for
```
epatters: Typos: 

- `seriailization` -> `serialization`
- `cripillingly` -> `cripplingly`

packages/frontend/src/stdlib/analyses/jupyter.ts
jmoggr: I was thinking of commenting out but leaving it in. It took me longer than I'd like to admit to figure out that this was possible, and it will likely be needed again in the future.

epatters: Sounds good, I'm happy to leave it in but commented out.

epatters (approved):

PR #692, opened 2025-08-19
Refactor migration execution

jmoggr:


packages/automerge-doc-server/src/migrations/m20250805230408_fix_automerge_storage.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        if (!handle) {
+            console.error(`Failed to create document for ref ${id}`);
+            continue;
+        }
+
+        await pool.query("UPDATE snapshots SET doc_id = $2 WHERE id = $1", [id, handle.documentId]);
+
+        // await handle.whenReady();
+        await sleep(500);
```
epatters, line 53: Is it worth asking the Automerge team whether this hack can be avoided?

If it ends up staying, we should add a further comment explaining in what way `await handle.whenReady()` doesn't work. (It doesn't ensure that the doc is written to the storage adapter?)

packages/migrator/README.md
```diff
@@ -1,36 +0,0 @@
...
-cargo run -p migrator help
-```
-
-To apply all migrations, run
-
-```sh
-cargo run -p migrator apply
-```
-
-## Writing new migrations
```
epatters, line 26: Is it worth adding these instructions on how to write migrations back somewhere, such as in the `README` for `backend`?

packages/automerge-doc-server/src/migrations/m20250805230408_fix_automerge_storage.ts
```diff
@@ -0,0 +1,58 @@
+import { Repo, type RepoConfig } from "@automerge/automerge-repo";
+import dotenv from "dotenv";
+import pgPkg from "pg";
+import { PostgresStorageAdapter } from "../postgres_storage_adapter.js";
+
+const { Pool } = pgPkg;
+
+dotenv.config();
+
+export async function fixAutomergeStorage() {
```
epatters, line 15: Can we add a doc-comment explaining that this migration is doing the same thing as the previous one (`automergeStorage`), but fixing a problem with that one? And explain what the problem was.

As it stands, it's hard to know what's going on without diff-ing the two migrations in your head.

packages/backend/src/main.rs
```diff
@@ -11,10 +24,19 @@ use tracing_subscriber::filter::{EnvFilter, LevelFilter};
...
 mod socket;
 mod user;
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum AppStatus {
+    Starting,
+    Migrating,
+    Running,
+    Failed(String),
+}
```
epatters: I'd move the definition of `AppStatus` to `app.rs` and import it here.

packages/backend/src/main.rs
```diff
@@ -42,62 +80,182 @@ async fn main() {
...
+    status_tx: watch::Sender<AppStatus>,
+) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let _ = status_tx.send(AppStatus::Migrating);
+    info!("Applying database migrations...");
+
+    let mut conn = db.acquire().await?;
+    migrator.run(&mut conn, &Plan::apply_all()).await.unwrap();
+
+    let _ = status_tx.send(AppStatus::Running);
+    let _ = sd_notify::notify(false, &[sd_notify::NotifyState::Ready]);
```
epatters: Why do `let _ = ` here?

jmoggr: `send` returns a `Result<(), SendError<T>>`, I wasn't sure if we should fail the whole process on something that is not necessary for the consistency of that process. `let = _ ` was the easiest way of doing that.

I'm not really sure what best practices are here, but given the confusion I'll change it to a `?`.

packages/automerge-doc-server/src/migrations/m20250805230408_fix_automerge_storage.ts
jmoggr: > Is it worth asking the Automerge team whether this hack can be avoided?

We'd like to know if there is a simple answer, but if the answer is not simple then that could turn into a time sink (~4hrs if there are multiple back and forths with testing) or possibly be seen as rude if we drop the issue. We can also just ignore the issue. I don't feel confident unilaterally making a decision like this. I'm slightly inclined to opening an issue re-evaluating if it gets complicated.

> If it ends up staying

If we don't want to block this PR on this issue, then I think we have to make this assumption. The comment is easy to add.

> It doesn't ensure that the doc is written to the storage adapter?

I don't have a confident explanation, all of my observations are now included in the comments.

packages/automerge-doc-server/src/migrations/m20250805230408_fix_automerge_storage.ts
jmoggr: done!

packages/migrator/README.md
jmoggr: Yes, that was my intention. done!

packages/backend/src/main.rs
jmoggr: done!

packages/backend/src/app.rs
```diff
@@ -17,6 +18,17 @@ pub struct AppState {
...
+
+    pub app_status: watch::Receiver<AppStatus>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum AppStatus {
+    Starting,
+    Migrating,
+    Running,
+    #[allow(dead_code)]
```
jmoggr, line 30: This warning didn't manifest when `AppStatus` was defined in `main.rs`. We currently don't have manual error handling so the `Failed` variant is never constructed, however code exists for handling it (a scar from the development process). I think it's okay to leave in.

packages/automerge-doc-server/src/migrations/m20250805230408_fix_automerge_storage.ts
epatters: I agree: let's not block this PR on a possible upstream issue. If this becomes a recurring problem as we do more migrations (hopefully not too soon :) we can invest more time into getting a less hacky fix.

epatters (approved):

PR #695, opened 2025-08-20
Replace Jupyter server bash script with Julia code

quffaro:
This PR moves the Jupyter kernel management to the CatColabInterop.jl package. This allows the user to manage and monitor the Jupyter kernel without needing to clone the CatColab repo.


packages/algjulia-interop/Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.1.0"
 
 [deps]
 ACSets = "227ef7b5-1206-438b-ac65-934d6da304b8"
+ArgParse = "c7e460c6-2fb9-53a9-8c5b-16f535851c63"
```
epatters: Despite being imported, this package is no longer being used.

packages/algjulia-interop/Project.toml
```diff
@@ -19,16 +20,19 @@ JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MLStyle = "d8e11817-5142-5d16-987a-aa16d5891078"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
+PackageCompiler = "9b87118b-4619-50d2-8e1e-99f35a4d4d9d"
```
epatters: Shouldn't need to depend on `PackageCompiler` (or a sys image) for ordinary use of this package. Can we make `build_sysimage` available through a package extension when `PackageCompiler` is imported?

packages/algjulia-interop/Project.toml
```diff
@@ -19,16 +20,19 @@ JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
...
+Preferences = "21216c6a-2e73-6563-6e65-726566657250"
+REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
```
epatters, line 23: Same comment about `REPL` as about `ArgParse`.

quffaro: `REPL` introduces `TerminalMenu`, which is used.

epatters (approved):
Looks great!

PR #696, opened 2025-08-20
Petri net reachability analysis

kris-brown:
This draft PR showcases the basic functionality of Petri net reachability analysis.

The details of it are succinctly described in the Handbook of Model checking (Ch 31, pg 1130-1131). The language for talking about regions of the space of petri net markings are upward closed sets (so the marking $(0,3,1)$ is a presentation of $$(x\geq 0, y\geq 3, z\geq 1)$$). 

There is an example in the book which is fully worked out (though frustratingly the petri net is drawn incorrectly). It involves a three place petri net with transitions `[]->[p1]`, `[p1,p2]->[p3]`, and `[p3]->[p2]` and a forbidden state of `(0,0,2)`, i.e. two tokens in `p3`. The space of tokenings which can reach the forbidden state is described by $[(0,0,2), (0,1,1),(0,2,0)]$.  

It's a very unpolished user interface presently.

Ideally one would have a visualization of the petri net and be able to click on the places to add or remove tokens from them, and one could have a set of forbidden regions (implicitly their union) and a set of initial tokenings which are all checked against it. Also just flipping the boolean at the end turns the spec into a _required_ region.

# TODO
- There is some reactivity bug: if you create a petri net, create some states, then create reachability analysis, you'll hit an error in Rust (it's looking in a dictionary for a key that doesn't exist). If you refresh, it's fine again. 
- ~Move main logic to catlog, not catlog-wasm~
- ~UI should represent tokenings with a table with rows for each state~

### DIFF
--- a/packages/frontend/src/stdlib/analyses/reachability.tsx
+++ b/packages/frontend/src/stdlib/analyses/reachability.tsx
@@ -15,7 +15,7 @@ import "./simulation.css";
 /** Configuration for a reachability analysis of a model. */
 export type ReachabilityContent = ReachabilityProblemData;
 
-type Simulator = (model: DblModel, data: ReachabilityContent) => boolean;
+type Checker = (model: DblModel, data: ReachabilityContent) => boolean;
 
 /** Configure a reachability analysis for use with models of a theory. */
 export function configureReachability(options: {
@@ -23,11 +23,11 @@ export function configureReachability(options: {
     name?: string;
     description?: string;
     help?: string;
-    simulate: Simulator;
+    check: Checker;
 }): ModelAnalysisMeta<ReachabilityContent> {
     const {
         id = "subreachability",
-        name = "Sub-reachability model checking",
+        name = "Sub-reachability check",
         description = "Check that forbidden tokenings are unreachable",
         help = "subreachability",
         ...otherOptions
@@ -45,7 +45,7 @@ export function configureReachability(options: {
 /** Check a reachability property in a model. */
 export function Reachability(
     props: ModelAnalysisProps<ReachabilityContent> & {
-        simulate: Simulator;
+        check: Checker;
         title?: string;
     },
 ) {
@@ -82,25 +82,24 @@ export function Reachability(
         }),
     ];
 
-    const isForbiddenUnreachable = createMemo<boolean | undefined>(() => {
+    const isChecked = createMemo<boolean | undefined>(() => {
         const validated = props.liveModel.validatedModel();
         if (validated?.tag !== "Valid") {
             return;
-        } else {
-            return props.simulate(validated.model, props.content);
         }
+        return props.check(validated.model, props.content);
     }, undefined);
 
     return (
         <div class="simulation">
-            <PanelHeader title="Subreachability analysis" />
+            <PanelHeader title={props.title} />
             <FixedTableEditor rows={obGenerators()} schema={obSchema} />
             <Switch>
-                <Match when={isForbiddenUnreachable() === false}>
-                    <p>{"\u274C forbidden tokening is reachable"}</p>
-                </Match>
-                <Match when={isForbiddenUnreachable() === true}>
+                <Match when={isChecked() === true}>
                     <p>{"\u2705 forbidden tokening is not reachable"}</p>
+                </Match>
+                <Match when={isChecked() === false}>
+                    <p>{"\u274C forbidden tokening is reachable"}</p>
                 </Match>
             </Switch>
         </div>
--- a/packages/frontend/src/stdlib/theories/petri-net.ts
+++ b/packages/frontend/src/stdlib/theories/petri-net.ts
@@ -48,7 +48,7 @@ export default function createPetriNetTheory(theoryMeta: TheoryMeta): Theory {
                 },
             }),
             analyses.configureReachability({
-                simulate(model, data) {
+                check(model, data) {
                     return thSymMonoidalCategory.subreachability(model, data);
                 },
             }),


KevinDCarlson:
Yay! Can you post a screenshot or gif showing an example input?

kris-brown:
reach

packages/catlog-wasm/Cargo.toml
```diff
@@ -20,6 +20,7 @@ egglog = { version = "0.4", default-features = false, features = [
 ] }
 getrandom_v03 = { package = "getrandom", version = "0.3", features = ["wasm_js"] }
 getrandom_v02 = { package = "getrandom", version = "0.2", features = ["js"] }
+itertools = "0.14.0"
```
epatters: This dependency is no longer needed in `catlog-wasm`, is it?

packages/catlog-wasm/src/analyses.rs
```diff
@@ -10,3 +10,8 @@ use catlog::stdlib::analyses;
 #[derive(Serialize, Deserialize, Tsify)]
 #[tsify(into_wasm_abi, from_wasm_abi)]
 pub struct ODEResult(pub JsResult<analyses::ode::ODESolution, String>);
+
+/// Input data for a mass-action dynamics analysis of a model.
+#[derive(Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct ReachabilityModelData(pub analyses::reachability::ReachabilityProblemData);
```
epatters: You should be able to delete this boilerplate entirely and directly use `ReachabilityProblemData` in the frontend code.

packages/catlog-wasm/src/theories.rs
```diff
@@ -335,6 +335,16 @@ impl ThSymMonoidalCategory {
                 .into(),
         ))
     }
+    /// Solve the subreachability problem for petri nets.
+    #[wasm_bindgen(js_name = "reachability")]
+    pub fn reachability(
```
epatters: Perhaps we should call this `subreachability` in hope that we eventually have genuine reachability analysis, its complexity notwithstanding.

I'm also wondering where the terminology came from. I mean, it makes sense, but Googling "subreachability Petri net" turns up very few results.

packages/catlog/src/dbl/modal/model.rs
```diff
@@ -75,8 +75,10 @@ impl MorListData {
 #[derive(Clone)]
 pub struct ModalDblModel {
     theory: Rc<ModalDblTheory>,
-    ob_generators: HashFinSet<QualifiedName>,
-    mor_generators: ComputadTop<ModalOb, QualifiedName>,
+    /// Object generators
+    pub ob_generators: HashFinSet<QualifiedName>,
+    /// Morphism generators
+    pub mor_generators: ComputadTop<ModalOb, QualifiedName>,
```
epatters: It shouldn't be necessary to make these fields public. In particular, the fact the morphism generators are stored using a separate computad data structure is really an implementation detail that could well change.

I'll try to indicate below what you can call instead.

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// Map from morphism IDs to rate coefficients (nonnegative reals).
+    pub tokens: HashMap<QualifiedName, i32>,
+
+    /// Map from object IDs to initial values (nonnegative reals).
+    pub forbidden: HashMap<QualifiedName, i32>,
+}
+
+/// The "Region Algebra for Petri Nets" algorithm from Ch 31 of
+/// the Handbook of Model Checking: "Symbolic Model Checking in Non Boolean
+/// Domains".
```
epatters: Thanks for including the citation. We can formalize it a bit using my cute little [module for references](https://next.catcolab.org/dev/rust/catlog/refs/).

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+pub fn reachability(m: &ModalDblModel, data: ReachabilityProblemData) -> Result<bool, String> {
+    // Convert model into a pair of matrices
+    //--------------------------------------
+
+    // Get a canonical ordering of the objects
+    let ob_vec: Vec<QualifiedName> = data.tokens.keys().cloned().sorted().collect();
```
epatters: IMO, it's nice style to avoid writing types when they can be inferred. In this case, you should be able to write

```rust
let ob_vec: Vec<_> = ...;
let ob_inv: HashMap<_, _> = ...;
```

epatters: Note to self (no action needed): in analyses, we keep have to write the same boilerplate to index objects by integer so that they can used with vectors. We should write a helper struct or function for this.

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    let ob_inv: HashMap<QualifiedName, usize> =
+        ob_vec.iter().enumerate().map(|(x, y)| (y.clone(), x)).collect();
+    let n_p: usize = ob_vec.len();
+    if n_p == 0 {
+        return Ok(true);
+    }
+
+    // Get a canonical ordering of the homs
+    let hom_vec: Vec<QualifiedName> = m.mor_generators.edge_set.iter().sorted().collect();
```
epatters: To avoid the private field access, the RHS can be replaced with `m.mor_generators().sorted().collect()`

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    let hom_inv: HashMap<QualifiedName, usize> =
+        hom_vec.iter().enumerate().map(|(x, y)| (y.clone(), x)).collect();
+    let n_t = hom_vec.len();
+
+    // Populate the I/O matrices from the hom src/tgt data
+    let mut i_mat = vec![vec![0; n_t]; n_p];
+    let mut o_mat = vec![vec![0; n_t]; n_p];
+
+    for e in m.mor_generators.edge_set.clone() {
```
epatters: Besides the private field access, you definitely shouldn't have to clone the whole field. I think `for e in m.mor_generators()` should suffice.

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+        let e_idx =
+            *hom_inv.get(&e).unwrap_or_else(|| panic!("Hom inv missing {e:?}\n{hom_inv:?}"));
+        if let Some(vs) = m.mor_generators.src_map.get(&e).unwrap().clone().collect_product(None) {
```
epatters: The RHS should be `m.mor_generator_dom(&e).collect_product(None)`.

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+            for v in vs.iter() {
+                if let ModalOb::Generator(u) = v {
+                    i_mat[*ob_inv
+                        .get(u)
+                        .unwrap_or_else(|| panic!("Ob inv missing {u:?}\n{ob_inv:?}"))][e_idx] += 1;
+                }
+            }
+        }
+
+        if let Some(vs) = m.mor_generators.tgt_map.get(&e).unwrap().clone().collect_product(None) {
```
epatters: Likewise, the RHS should be `m.mor_generator_cod(&e).collect_product(None)`.

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+        hom_vec.iter().enumerate().map(|(x, y)| (y.clone(), x)).collect();
+    let n_t = hom_vec.len();
+
+    // Populate the I/O matrices from the hom src/tgt data
+    let mut i_mat = vec![vec![0; n_t]; n_p];
+    let mut o_mat = vec![vec![0; n_t]; n_p];
+
+    for e in m.mor_generators.edge_set.clone() {
+        let e_idx =
+            *hom_inv.get(&e).unwrap_or_else(|| panic!("Hom inv missing {e:?}\n{hom_inv:?}"));
```
epatters: The idiomatic way to unwrap with a custom error message is [`expect`](https://doc.rust-lang.org/std/option/enum.Option.html#method.expect).

That being said, it should be an invariant of this code, due to way that `hom_inv` is constructed, that `hom_inv` is never undefined on an edge, so if it were me I'd probably just call `unwrap`.

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+            ),
+            ModalOb::App(ModalOb::List(List::Symmetric, vec![z.clone()]).into(), op.clone()),
+            ModalMorType::Zero(ob_type.clone()),
+        );
+        model.add_mor(
+            t3,
+            ModalOb::App(ModalOb::List(List::Symmetric, vec![z.clone()]).into(), op.clone()),
+            ModalOb::App(ModalOb::List(List::Symmetric, vec![y.clone()]).into(), op.clone()),
+            ModalMorType::Zero(ob_type),
+        );
```
epatters: Note to self (no action needed): we should write a helper function for defining Petri nets, if only to make the tests easier to write. This is painful.

Or better, now that I think of it, once @olynch's #716 we should have textual syntax to define models, which should make writing this sort of test much more convenient.

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// [(0,0,2)] -> [(0,0,2),(1,1,1)] -> [(0,0,2),(0,1,1),(2,2,0)]
+    /// -> [(0,0,2),(0,1,1),(0,2,0)]
+    /// So the three ways one can reach the forbidden state are:
+    /// 1.) starting in the forbidden state (or any superset)
+    /// 2.) having two tokens in p2
+    /// 3.) having one token in each p2 and p3
+    ///
+    /// Consider using algorithm from "Minimal Coverability Tree Construction
+    /// Made Complete and Efficient" for a more efficient algorithm which allows
+    /// "inf" as a possible specification of an invalid state.
```
epatters: Thanks for the detailed description of the example!

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        // Update f with new stuff and remove extraneous old stuff
+        f.retain(|v| newstuff.iter().all(|n| (0..n_p).any(|p| v[p] < n[p])));
+        f.extend(newstuff);
+    }
+
+    // Check whether input tokening lies within the region which can access
+    // the forbidden state, `init`.
+    let init_in_forbbiden = f.iter().any(|v| (0..n_p).all(|p| v[p] <= init[p]));
+    Ok(!init_in_forbbiden)
```
epatters: As it stands, this function never returns the `Err` variant. If that is indeed the desired state of affairs, it would be better to return `bool` rather than `Result<bool, String>`.

In general, here's the approach to error handling that I'm trying to take:

- if the error can occur because of user provided data, then return `Err` via a `Result` so that error can be handled gracefully downstream or at least displayed to the user
- if the error is due to a failure of an internal invariant, then just panic because that's a bug in the code and we shouldn't try to hide it

In this function, I think there's a mix of these types, but I suspect that you can avoid the former by using default values, a common pattern in these analyses. E.g., instead of `forbidden.get(u).unwrap()` above do `forbidden.get(u).unwrap_or_default(0)`.

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+/// Data defining a mass-action ODE problem for a model.
+#[derive(Clone)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde-wasm", derive(Tsify))]
+#[cfg_attr(
+    feature = "serde-wasm",
+    tsify(into_wasm_abi, from_wasm_abi, hashmap_as_object)
+)]
+pub struct ReachabilityProblemData {
+    /// Map from morphism IDs to rate coefficients (nonnegative reals).
```
epatters: Mismatched docstring from copy-paste.

packages/catlog/src/stdlib/analyses/reachability/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub tokens: HashMap<QualifiedName, i32>,
+
+    /// Map from object IDs to initial values (nonnegative reals).
```
epatters: Ditto.

packages/frontend/src/stdlib/analyses/reachability.tsx
```diff
@@ -0,0 +1,98 @@
...
+/** Analyze a model using Reachability formula. */
+export function Reachability(
+    props: ModelAnalysisProps<ReachabilityContent> & {
+        simulate: Simulator;
+        title?: string;
+    },
+) {
+    const obDecls = createMemo<ObjectDecl[]>(() => {
+        return props.liveModel.formalJudgments().filter((jgmt) => jgmt.tag === "object");
+    }, []);
```
epatters: Sorry for the churn, but as of #702, we're no longer directly accessing the formal judgments in the notebook, but rather using the elaborated model. This is necessary for analyses to work with compositional models once that feature lands.

Have a look at the diff to an analysis in #702 to see what you need to change or ask me if it's not clear.

epatters (changes requested):
Thanks Kris, I'm excited to have our first analysis that does model checking!

Are you still having the reactivity issue? It's not immediately obvious to me why that would be, but if so I can poke around more.

packages/catlog/src/stdlib/analyses/reachability/mod.rs
epatters: Ditto for `sir_petri` defined in #619.

PR #698, opened 2025-08-21
Implementing `petgraph` topological sort

quffaro:


packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -168,6 +168,118 @@ where
     result
 }
 
+fn out_neighbors<G>(graph: &G, v: &G::V) -> impl Iterator<Item = G::V>
```
epatters: Since they're generally useful, might as well as add `out_neighbors` and `in_neighbors` as methods to the `FinGraph` trait with these bodies as default implementations.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -168,6 +168,118 @@ where
...
+/** Computes a topological sorting for a given graph.
+
+This algorithm was borrowed from `petgraph`.
+ */
+pub fn toposort<G>(graph: &G) -> Result<Vec<G::V>, String>
+where
+    G: FinGraph,
+    G::V: Hash + std::fmt::Debug,
+{
+    let n = graph.vertices().collect::<Vec<_>>().len();
```
epatters: Use `graph.vertex_count()` here.

In general, you shouldn't collect to a vector to determine the length of an iterator; you can use the `count` method of iterator instead.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -211,6 +323,39 @@ mod tests {
...
+    #[test]
+    fn toposorting() {
+        let g = SkelGraph::path(5);
+        assert_eq!(toposort(&g), Ok(vec![0, 1, 2, 3, 4]));
+
+        let mut g = SkelGraph::path(3);
+        g.add_vertices(1);
+        let _ = g.add_edge(2, 3);
+        let _ = g.add_edge(3, 0);
+        assert_eq!(toposort(&g), Err("cycle detected involving node 3".to_owned()));
```
epatters: I'd avoid baking in the exact error message by doing:

```rust
assert!(toposort(&g).is_err());
```

Or, if you prefer to the test the error message (not a bad idea), you can use `expect_test`.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -211,6 +323,39 @@ mod tests {
...
+
+        let g = SkelGraph::triangle();
+        assert_eq!(toposort(&g), Ok(vec![0, 1, 2]));
+
+        let mut g = SkelGraph::path(4);
+        g.add_vertices(2);
+        let _ = g.add_edge(1, 4);
```
epatters: Remove the `let _ = ` and just write `g.add_edge(...)`.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -211,6 +323,39 @@ mod tests {
...
+        let mut g: HashGraph<u8, &str> = Default::default();
+        g.add_vertices(vec![0, 1, 2, 3, 4, 5]);
+        g.add_edge("0-1", 0, 1);
+        g.add_edge("1-2", 1, 2);
+        g.add_edge("2-3", 2, 3);
+        g.add_edge("1-4", 1, 4);
+        g.add_edge("4-3", 4, 3);
+        g.add_edge("5-2", 5, 2);
+        // TODO non-deterministic
+        // assert_eq!(toposort(&g), Ok(vec![5, 0, 1, 2, 4, 3]));
```
epatters: Lingering TODO. Perhaps just test that one of the known constraints is satisfied, e.g., 5 comes before 2.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -168,6 +168,118 @@ where
...
+    fn init(n: usize) -> Self {
+        Self {
+            visited: vec![false; n],
+        }
+    }
+}
+
+/** Computes a topological sorting for a given graph.
+
+This algorithm was borrowed from `petgraph`.
```
epatters: Add GitHub permalink to the relevant code in petgraph.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -168,6 +168,118 @@ where
...
+        let previous = self.visited[idx];
+        self.visited[idx] = true;
+        !previous
+    }
+
+    fn is_visited(&self, idx: usize) -> bool {
+        self.visited[idx]
+    }
+
+    fn init(n: usize) -> Self {
```
epatters: The idiomatic name for a constructor in Rust is `new`, not `init`.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -168,6 +168,118 @@ where
...
+    G: FinGraph,
+    G::V: Hash + std::fmt::Debug,
+{
+    let n = graph.vertices().collect::<Vec<_>>().len();
+    let mut discovered = VisitMap::init(n);
+    let mut finished = VisitMap::init(n);
+    let mut finish_stack: Vec<G::V> = Vec::new();
+    let mut stack = Vec::new();
+
+    let gmap: HashMap<_, _> = HashMap::from_iter(graph.vertices().enumerate().map(|(k, v)| (v, k)));
```
epatters: Wouldn't it be simpler to have `VisitMap` directly store `HashMap<V, bool>` than to maintain a separate mapping between vertices and integers?

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -168,6 +168,118 @@ where
...
+            }
+        }
+    }
+    finish_stack.reverse();
+
+    let mut discovered = VisitMap::init(n);
+    for i in &finish_stack {
+        stack.clear();
+        stack.push(i.clone());
+        //
```
epatters: Stray empty comment.

PR #704, opened 2025-08-28
Fixes and QoL improvements for CatColabInterop kernel management

quffaro:
- Added `uninstall_kernel!` function
- Bumped IJulia to v1.30.1, resolving a bug where `IJulia.set_max_stdio` breaks from unspecified kernel (c.f., [this PR](https://github.com/JuliaLang/IJulia.jl/pull/1178#event-19377499266))
- Added `manual` keyword argument to `start_server!` to select kernel and mode before starting the server
- Removed terminal colors to cut down ANSI codes appearing in Julia kernel output

### DIFF
--- a/packages/algjulia-interop/src/kernel_management.jl
+++ b/packages/algjulia-interop/src/kernel_management.jl
@@ -10,20 +10,20 @@ end
 
 KernelNotFoundException() = KernelNotFoundException(IJulia.kerneldir())
 
-function Base.showerror(err::KernelNotFoundException)
-    """
-    IJulia cannot find any kernels in the directory $(err.dir). To install a kernel, you may run `CatColabInterop.install_ccl_kernel()`. Refer to the [IJulia documentation](https://julialang.github.io/IJulia.jl/stable/library/public/#IJulia.installkernel) for more information about managing Jupyter kernels in IJulia.
+function Base.showerror(io::IO, err::KernelNotFoundException)
+    print(io, """
+    IJulia cannot find any kernels in the directory `$(err.dir)`.
+
+    To install a kernel, you may run `CatColabInterop.install_ccl_kernel()`.
+    Refer to the [IJulia documentation](https://julialang.github.io/IJulia.jl/stable/library/public/#IJulia.installkernel)
+    for more information about managing Jupyter kernels in IJulia.
 
     If you wish to install a sysimage instead, run
     ```julia
     using PackageCompiler
     install_ccl_kernel(Val(:sysimge))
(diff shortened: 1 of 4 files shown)


packages/algjulia-interop/src/kernel_management.jl
```diff
@@ -2,24 +2,63 @@ using IJulia
 using Preferences
 import REPL
 using REPL.TerminalMenus
+using MLStyle
```
epatters, line 5: This is fun: I like having a `Result` type in Julia. But how about we move this to a separate file `result.jl` to avoid distraction in this one?

packages/algjulia-interop/src/kernel_management.jl
```diff
@@ -2,24 +2,63 @@ using IJulia
...
     dir::String
 end
 
-function Base.showerror(io, e::KernelNotFoundException)
-    print(io, """IJulia cannot find any kernels in
-    $(e.dir)
-    To install a kernel, you may run `CatColabInterop.install_ccl_kernel()`. Refer to the [IJulia documentation](https://julialang.github.io/IJulia.jl/stable/library/public/#IJulia.installkernel) for more information about managing Jupyter kernels in IJulia.""")
+KernelNotFoundException() = KernelNotFoundException(IJulia.kerneldir())
+
+function Base.String(err::KernelNotFoundException)
```
epatters: I believe that overloading the constructor `Base.String` is not the preferred way to control how an object is displayed in Julia. When I load Julia in a fresh REPL, `Base.String` has only 13 methods, whereas `Base.show` has 234.

In this case, I would just directly define the new method for `Base.showerror`.

packages/frontend/src/stdlib/analyses/decapodes.tsx
```diff
@@ -341,6 +341,10 @@ type SimulationData = {
 /** Julia code run after kernel is started. */
 const makeInitCode = () =>
     `
+    for k in keys(Base.text_colors)
+		Base.text_colors[k] = ""
+	end
+
```
epatters, line 347: Have you tested that this works as expected by triggering a stack trace and seeing how it appears in the frontend?

quffaro: image
Yes, there are still a few ANSI codes remaining but the stacktrace was actually legible!

epatters: Much improved!

epatters (approved):

PR #705, opened 2025-08-31
Configure Simulation Mesh from Decapodes Analysis

quffaro:
The user should be able to configure the simulation mesh from the frontend.


epatters:
Closing due to staleness. Could be a good thing to revive when the Decapodes analysis is running again.

Issue #13, opened 2024-05-21
Consider using `indexmap` for hash maps


epatters:
The Rust package [indexmap](https://lib.rs/crates/indexmap) provides drop-in replacements for `HashSet` and `HashMap` that maintain an ordering on keys. That's almost always what we want, so we should probably just make this our default data structure for hash sets and maps.

olynch:
Question: do we want insertion-order, or do we want to sort them? Arguably, sorting is even *more* deterministic.

epatters:
Good point. I'm not sure which is preferable. One advantage to insertion order is that it will preserve any order coming from the user interface, which might be convenient for debugging.

PR #716, opened 2025-09-09
DoubleTT: type theory for models and their morphisms

olynch:
This PR implements a type theory for double models and their morphisms.

### DIFF
--- a/packages/catlog/src/tt/mod.rs
+++ b/packages/catlog/src/tt/mod.rs
@@ -287,12 +287,12 @@ and normalization of the `.g1.V` field of a model of `Graph2`.
 Specialization is a way of creating *subtypes* of a record type by setting the
 type of a field to be a subtype of its original type. So in order to understand
 specialization, one must first understand subtyping. In doublett, there are
-two ways for form subtypes. We write `A <: B` for "`A` is a subtype of `B`".
+two ways to form subtypes. We write `A <: B` for "`A` is a subtype of `B`".
 
 1. If `a : A`, then `@sing a <: A`.
 2. If `A` is a record type with a field `.x`, and `B` is a subtype of the type of
-`a.x` for a generic element `a : A`, then `A & [ .x : B ] <: A`. The notation `A
-& [ .x := y ]` is just syntactic sugar for `A & [ .x : @sing y ]`.
+   `a.x` for a generic element `a : A`, then `A & [ .x : B ] <: A`. The notation `A
+   & [ .x := y ]` is just syntactic sugar for `A & [ .x : @sing y ]`.
 
 Crucially, the type of `a.x` may depend on the values of other fields of `a`, so
 it is important that the subtyping check is performed in the context of a *generic*
--- a/packages/catlog/src/tt/batch.rs
+++ b/packages/catlog/src/tt/batch.rs
@@ -160,16 +160,16 @@ pub fn run(path: &str, output: &BatchOutput) -> io::Result<bool> {
 /// Run the doublett elaborator in batch mode
 pub fn elaborate(src: &str, path: &str, output: &BatchOutput) -> io::Result<bool> {
     let reporter = Reporter::new();
-    let source_info = SourceInfo::new(Some(path), &src);
+    let source_info = SourceInfo::new(Some(path), src);
     let start_t = Instant::now();
     let _unwind_guard = guard((), |_| {
         output.report(&reporter, &source_info);
     });
     let mut succeeded = true;
-    let _ = PARSE_CONFIG.with_parsed_top(&src, reporter.clone(), |topntns| {
+    let _ = PARSE_CONFIG.with_parsed_top(src, reporter.clone(), |topntns| {
         let mut toplevel = Toplevel::new(stdlib::th_schema());
         for topntn in topntns.iter() {
-            output.log_input(&src, topntn);
+            output.log_input(src, topntn);
             let mut should_fail = false;
             for annot in topntn.annotations {
                 // We allow single_match here because in the future we might want


olynch:
@epatters do you know why the frontend build is failing here?

epatters:
Looks like a dependency that you've added, `notify-debouncer-full`, is failing to compile to Wasm. Judging by the error message, it might be trying to compile something for Windows. Regardless, many packages require setting feature flags to compile to Wasm.

In this case, likely `notify` should be a dev dependency not compiled for the Wasm target at all?

olynch:
Ah, that makes sense!

olynch:
@epatters Maybe it makes sense to just merge this now, and then make a new PR for integrating it into the frontend.

olynch:
Oh, there's still one more thing to do in the elaborator, which is implement morphism composition.

packages/catlog/Cargo.toml
```diff
@@ -5,31 +5,44 @@ authors = ["Evan Patterson"]
...
 
+
 [features]
 ode = ["dep:ode_solvers", "dep:nalgebra"]
 serde = ["dep:serde", "nonempty/serialize", "ustr/serde", "uuid/serde"]
 serde-wasm = ["serde", "dep:wasm-bindgen", "dep:tsify", "dep:instant", "instant/wasm-bindgen"]
 
 [dependencies]
+bwd = "0.2.1"
+clap = { version = "4.5.47", features = ["derive"] }
```
epatters: I figure that `clap`, like `notify`, should be a dev dependency. No reason to include it in Wasm blob.

packages/catlog/Cargo.toml
```diff
@@ -5,31 +5,44 @@ authors = ["Evan Patterson"]
...
 derivative = "2"
 derive_more = { version = "2", features = ["constructor", "from", "into"] }
 duplicate = "2"
 egglog = { version = "0.5", default-features = false }
 ego-tree = "0.10"
 instant = { version = "0.1", optional = true } # Needed for egglog v0.5
+fnotation = "0.8.1"
+indexmap = "2.11.1"
```
epatters, line 23: Note to self (no action needed): if we're adding `indexmap` as a dep, perhaps a good time to revisit one of our earliest issues, #13.

packages/catlog/src/tt/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+pub mod eval;
+#[allow(unused)]
+mod prelude;
+#[allow(unused)]
+pub mod stx;
+#[allow(unused)]
+pub mod toplevel;
+#[allow(unused)]
+pub mod util;
+#[allow(unused)]
```
epatters: These `allow` directives should be removed before merging. Or, if necessary, localized to a few definitions intended to be used later.

packages/catlog/examples/check.rs
```diff
@@ -0,0 +1,160 @@
...
+        if let Err(e) = watcher.watch(Path::new(&args.path), RecursiveMode::Recursive) {
+            eprintln!("could not watch {}: {}", &args.path, e)
+        }
+
+        for res in rx {
+            match res {
+                Ok(es) => {
+                    let mut modified = false;
+                    for e in es.iter() {
+                        #[allow(clippy::single_match)]
```
epatters: I'm not too bothered about this, but I guess the idiomatic way to write this is:

```rust
if matches!(e.kind, notify::EventKind::Modify(_)) {
    modified = true;
}
```

While I'm bikeshedding, the whole block could be replaced with

```rust
let modified = es.iter().any(|e| matches!(e.kind, notify::EventKind::Modify(_)));
```

Similarly for `should_fail` above.

packages/catlog/examples/test.dbltt
epatters, line 1: This is a cool and readable way to test while developing! If I'm not mistaken, this file is currently only run by the `check` program. Can we get an automated test doing the same?

packages/catlog/examples/test.dbltt
```diff
@@ -0,0 +1,82 @@
+type u := Unit
+
+type u2 := u
+
+def t : u := 'tt
```
epatters, line 6: I don't get this line. Was `tt` previously defined? What does the `'` mean?

packages/catlog/examples/test.dbltt
```diff
@@ -0,0 +1,82 @@
...
+  Weight : AttrType,
+  src : (Id Entity)[E, V],
+  tgt : (Id Entity)[E, V],
+  weight : Attr[E, Weight]
+]
+
+type EntityArr := [
+  src : Entity,
+  tgt : Entity,
+  arr : (Id Entity)[src, tgt]
```
epatters: I get what's going on here, but IMO it'd be easier to understand if you called these `dom` and `cod`, to avoid confusion with the `src` and `tgt` in `WeightedGraph`. Unless you really want to test the case where the identifiers happen to be the same :)

packages/catlog/examples/test.dbltt
olynch: `tt` is a traditional name for the unique element of `Unit`. In fnotation, `'` is used to form "tags" which are the constructors of sum types. E.g., `'some 1 : Option Int`. As `Unit` can be thought of as a sum type with one constructor that has no arguments, `'tt` is that constructor.

packages/catlog/src/tt/mod.rs
olynch: Done!

packages/catlog/src/tt/stx.rs
```diff
@@ -0,0 +1,364 @@
...
+                    write!(f, "{segment}")?;
+                }
+                Ok(())
+            }
+        }
+    }
+}
+
+impl MorphismType {
+    fn to_doc<'a>(&self) -> D<'a> {
```
epatters, line 40: The `pretty` crate seems great. I'm going to start using it.

That said, I'm wondering why you define this method and then implement `fmt::Display` above directly, when it seems like you could call `self.to_doc().render_fmt(f)` instead.

packages/catlog/src/tt/stx.rs
```diff
@@ -0,0 +1,364 @@
+/*! Syntax for types and terms.
+
+See [crate::tt] for what this means.
```
epatters, line 3: Noting that the referenced docs have not been written yet. I won't block this PR on that, especially if you suspect the formalism to be documented to evolve considerably in the near future, but it would be great to get some high-level documentation of the overall approach.

packages/catlog/src/tt/util/idx.rs
```diff
@@ -0,0 +1,77 @@
+/*! Indices.
+
+We find it useful to distinguish between backward indices (used in syntax, where
+0 refers to the *end* of the context) and forward indices (used in values, where
+0 refers to the *beginning* of the context).
+
+In the literature, backward indices are known as DeBruijn indices and forward
+indices are known as DeBruijn levels, but we think that "backwards and forwards"
+is more clear, and it corresponds with "backwards and forwards linked lists". A forwards linked list uses `cons` to put a new element on the front; a backwards
```
epatters, line 9: NGL, I really appreciate this because on the usual terminology I cannot remember which are "indices" and "levels".

packages/catlog/src/tt/util/idx.rs
```diff
@@ -0,0 +1,77 @@
...
+use std::ops::Deref;
+
+/** Forward indices (aka DeBruijn levels)
+
+Get the underlying `usize` using the [Deref] implementation.
+*/
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub struct FwdIdx(usize);
+
+impl From<usize> for FwdIdx {
```
epatters: A shortcut I've come to appreciate: you can save on this boilerplate by importing

```rust
use derive_more::From;
```

then deriving `From`.

packages/catlog/src/tt/util/idx.rs
```diff
@@ -0,0 +1,77 @@
...
+    fn from(value: usize) -> Self {
+        Self(value)
+    }
+}
+
+impl Deref for FwdIdx {
```
epatters, line 25: It appears that `derive_more` also supports [`Deref`](https://docs.rs/derive_more/latest/derive_more/derive.Deref.html), though I haven't tried that one (we'd need to enable the feature flag).

packages/catlog/src/tt/stx.rs
```diff
@@ -0,0 +1,364 @@
...
+pub enum Ty0 {
+    /// The type of (objects of a given object type).
+    Object(ObjectType),
+    /// Non-dependent record type.
+    Record(Row<Ty0>),
+    /// Unit type.
+    Unit,
+}
+
+/** Context of record type syntax. */
```
epatters: "Context" should be "Content", right? I'm comparing with `RecordV`.

packages/catlog/src/tt/stx.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    Each of these types is meant to be evaluated in an environment
+    where the last element is a value of type `fields0`.
+    */
+    pub fields1: Row<TyS>,
+}
+
+impl RecordS {
+    /// Constructor for [RecordS].
+    pub fn new(fields0: Row<Ty0>, fields1: Row<TyS>) -> Self {
```
epatters: You can also derive these with `derive_more::Constructor`.

packages/catlog/src/tt/val.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    Precondition: assumes that this produces a subtype.
+    */
+    pub fn add_specialization(&self, path: &[FieldName], ty: TyV) -> Self {
+        Self {
+            specializations: merge_specializations(
+                &self.specializations,
+                &Dtry::singleton(path, ty),
+            ),
+            ..self.clone()
```
epatters, line 47: Nice, I didn't know that Rust supported JS-style spread syntax.

Given that `add_specialization` and `specialize` clone `self`, I wonder whether it would be more natural to make these take `self` by value rather than by reference.

packages/catlog/src/tt/stx.rs
```diff
@@ -0,0 +1,364 @@
...
+        Self { fields0, fields1 }
+    }
+}
+
+/** Inner enum for [TyS]. */
+pub enum TyS_ {
+    /** A reference to a top-level declaration */
+    TopVar(TopVarName),
+    /** Type constructor for object types.
```
epatters, line 87: Thanks for these detailed docstrings, very helpful!

packages/catlog/src/tt/stx.rs
olynch: In progress!

packages/catlog/src/tt/stx.rs
olynch: Probably just a holdover from before I had defined the fmt::Display trait XD

packages/catlog/src/tt/stx.rs
olynch: Yes, fixed!

packages/catlog/src/tt/stx.rs
olynch: Nice!

packages/catlog/src/tt/stx.rs
olynch: No problem!

packages/catlog/src/tt/val.rs
olynch: I close `self` as a shortcut for cloning the non-specialization fields of `self`. I think it makes sense that this is a "non-consuming" method; it's supposed to create a modified version of self, and it's fine that the original version sticks around.

packages/catlog/src/tt/util/idx.rs
olynch: I'll keep this one manual for now.

epatters (approved):

PR #718, opened 2025-09-11
Patchwork tools for model and analysis editors

### LINKED ISSUE #714: Patchwork module with model and analysis tools
The Ink & Switch prepared an initial integration with Patchwork for the August workshop:

https://github.com/inkandswitch/CatColab/tree/main/packages/patchwork

To make it maintainable, we need to upstream this integration, consisting of a Patchwork module with a few plugins that provide the CatColab model and analysis editors as tools. To do this cleanly, we'll need to lightly refactor how API objects are passed through our components to avoid needless dependency on things that play no role in Patchwork, such as the CatColab backend and Firebase auth.

epatters:
Closes #714. This PR started from the Patchcolab demo created by Ink & Switch.

### DIFF
--- a/packages/patchwork/package.json
+++ b/packages/patchwork/package.json
@@ -1,7 +1,7 @@
 {
-    "name": "@patchwork/solidjs-demo",
+    "name": "catcolab-patchwork",
     "version": "0.0.1",
-    "description": "A demo showing how to use SolidJS inside a React app",
+    "description": "A Patchwork module for CatColab",
     "type": "module",
     "main": "src/index.ts",
     "exports": {
--- a/packages/patchwork/src/model_datatype.ts
+++ b/packages/patchwork/src/model_datatype.ts
@@ -1,7 +1,6 @@
 import * as A from "@automerge/automerge";
 import type {
     HasVersionControlMetadata,
-    Annotation,
     TextPatch,
     DecodedChangeWithMetadata,
 } from "@patchwork/sdk/versionControl";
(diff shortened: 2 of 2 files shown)


quffaro:
There is a small UX regression introduced in this branch. It's not critical, but I'll record my observation here.

This gif records drag-and-drop on the main branch. Notice that dragging the cell outside the model editor (list of cells) preserves the state of the drop indicator, which allows the dragged cell to drop onto the first element despite the drop location being outside a valid zone. The `notebookRef` allows us to do this.
main

This gif records drag-and-drop on this branch, where the drop indicator no longer points to the topmost cell when the cell is dragged outside the model editor
patchcolab

packages/patchwork/src/model_datatype.ts
```diff
@@ -0,0 +1,187 @@
...
+    name: string;
+    theory: string;
+    type: string;
+    notebook: {
+        cells: Cell<unknown>[];
+    };
+    analysisDocUrl: AutomergeUrl;
+};
+
+export const patchesToAnnotations = (doc: ModelDoc, _docBefore: ModelDoc, patches: A.Patch[]) => {
```
epatters: @paulsonnentag, this is last vestige of the annotations code from the original branch. Can/should I remove this too? It's used in the definition of `datatype` below.

paulsonnentag: You can delete this function. It's optional in the datatype and with the new annotations API it will be no longer necessary to implement a manual conversion from patches to annotations.

packages/patchwork/package.json
```diff
@@ -0,0 +1,44 @@
+{
+    "name": "@patchwork/solidjs-demo",
+    "version": "0.0.1",
+    "description": "A demo showing how to use SolidJS inside a React app",
```
paulsonnentag: name should be probably something like "frontend-patchwork"

packages/patchwork/src/model_datatype.ts
```diff
@@ -0,0 +1,187 @@
...
+                changedCells.add(after.id);
+                return;
+            }
+        }
+    });
+
+    return annotations;
+};
+
+const valueOfAnchor = (doc: ModelDoc, anchor: Uuid): Cell<unknown> => {
```
paulsonnentag: valueOfAnchor, sortAnchorsBy can be deleted as well

PR #722, opened 2025-09-17
Generate `catlog` models from doublett types

olynch:
This is surprisingly easy. This can be seen in action at `packages/catlog/examples/test.dbltt`, results in `test.dbltt.snapshot`.

### DIFF
--- a/packages/catlog/src/tt/elab.rs
+++ b/packages/catlog/src/tt/elab.rs
@@ -1,12 +1,20 @@
 /*! Elaboration for doublett */
-use crate::dbl::{category::VDblCategory, theory::DblTheory};
+use crate::{
+    dbl::{
+        category::VDblCategory,
+        model::{DblModel, FgDblModel},
+        theory::DblTheory,
+    },
+    one::FgCategory,
+};
 use fnotation::*;
 use nonempty::nonempty;
 use scopeguard::{ScopeGuard, guard};
+use std::fmt::Write;
 use tattle::declare_error;
 
 use crate::{
-    tt::{eval::*, prelude::*, stx::*, toplevel::*, val::*},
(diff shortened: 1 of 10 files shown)


packages/catlog/src/tt/elab.rs
```diff
@@ -158,6 +166,34 @@ impl<'a> TopElaborator<'a> {
                 let (tm_s, _) = elab.chk(&ty_v, tm_n)?;
                 Some(TopElabResult::Output(format!("{tm_s}")))
             }
+            "generate" => {
+                let mut elab = self.elaborator();
+                let (_, ty_v) = elab.ty(tn.body)?;
+                let model = generate(self.toplevel, &ty_v);
+                let mut out = String::new();
```
epatters, line 194: Might be cleaner to extract the string generation into a separate function (or even a `Display` method) that returns `std::fmt::Result` so that you can handle errors with `?` instead of having to call `unwrap` everywhere.

olynch: Given that the way of displaying here is specific to this usage (e.g., each line is prefixed with the fnotation comment character), I think it makes most sense to have this as a function in this file, not as a display method, but I agree it makes sense to factor it out, and that is what I did!

packages/catlog/src/tt/elab.rs
```diff
@@ -27,6 +36,30 @@ pub struct TopElaborator<'a> {
     reporter: Reporter,
 }
 
+fn model_output(out: &mut impl fmt::Write, model: &DiscreteDblModel) -> fmt::Result {
+    writeln!(out)?;
+    writeln!(out, "#/ object generators: ")?;
+    let mut obgens = model.ob_generators().collect::<Vec<_>>();
+    obgens.sort();
```
epatters: As of #725 (started after this PR), sorting is no longer necessary! I'll fix this now.

PR #726, opened 2025-09-21
Convert block comments to line comments

olynch:
The recommended comment style in rust is line comments (https://doc.rust-lang.org/nightly/style-guide/#doc-comments). This PR converts all of our existing comments to line comments, and additionally changes the format action in CI to enforce this style in the future.


epatters (approved):
Thanks Owen!

I changed the CI to use the latest nightly, rather than a pinned nightly, for `rustfmt` since I don't want to have to remember to update the build date at some frequency. If the latest nightly turns out to be too unreliable, we can revert to a pinned one.

PR #727, opened 2025-09-21
DoubleTT elaboration for notebooks

olynch:
This is a draft PR for notebook elaboration.


packages/catlog/src/tt/context.rs
```diff
@@ -0,0 +1,50 @@
...
+
+/// The variable context during elaboration
+pub struct Context {
+    /// Stores the value of each of the variables in context
+    pub env: Env,
+    /// Stores the names and types of each of the variables in context.
+    ///
+    /// We allow the type to be "none" as a hack for the `self` variable before we
+    /// know the type of the `self` variable.
+    pub scope: Vec<(VarName, LabelSegment, Option<TyV>)>,
```
epatters: IMO, it would increase legibility to replace this triple with a struct, something like the following if I'm understanding the meaning each of field correctly:

```rust
struct VarInContext {
  name: VarName,
  label: LabelSegment,
  type: Option<TyV>
}
```

Also feels slightly incongruous to have `VarName` as a type alias for `NameSegment` but use `LabelSegment` directly. Why not also use `NameSegment` directly?

packages/catlog/src/tt/mod.rs
```diff
@@ -352,11 +352,14 @@
 //! be semantically the same type.
 
 pub mod batch;
-pub mod elab;
+pub mod context;
 pub mod eval;
 pub mod modelgen;
+#[allow(unused)]
+pub mod notebook_elab;
```
epatters, line 358: Can you remove `allow(unused)` here?

packages/catlog/src/tt/notebook_elab.rs
```diff
@@ -0,0 +1,332 @@
...
+        let src = fs::read_to_string(format!("examples/{name}.json")).unwrap();
+        let doc: ModelDocumentContent = serde_json::from_str(&src).unwrap();
+        let cells = doc
+            .notebook
+            .cells()
+            .filter_map(|c| match c {
+                notebook_types::v1::NotebookCell::Formal { id, content } => Some((*id, content)),
+                _ => None,
+            })
+            .collect::<Vec<_>>();
```
epatters, line 292: Should the notebook elaborator have an entry point that does this filtering? We'll need to do the same thing when processing a notebook from the frontend.

packages/catlog/src/tt/prelude.rs
```diff
@@ -20,3 +18,7 @@ pub type FieldName = NameSegment;
 pub(super) fn text_seg(s: impl Into<Ustr>) -> NameSegment {
     NameSegment::Text(s.into())
 }
+
+pub(super) fn label_seg(s: impl Into<Ustr>) -> LabelSegment {
+    LabelSegment::Text(s.into())
+}
```
epatters: These helper functions should be moved to the module `crate::zero::qualified`, where there are already analogous helpers `name` and `label` to construct `QualifiedName`s and `QualifiedLabel`s. Also, for consistency, `text_seg` should be renamed to `name_seg`.

packages/catlog/src/tt/modelgen.rs
```diff
@@ -1,28 +1,38 @@
...
+    ty: &TyV,
+) -> (DiscreteDblModel, HashMap<QualifiedName, QualifiedLabel>) {
     let eval = Evaluator::new(toplevel, Env::Nil, 0);
-    let (elt, eval) = eval.bind_neu(text_seg("self"), ty.clone());
+    let (elt, eval) = eval.bind_self(ty.clone());
     let elt = eval.eta_neu(&elt, ty);
     let mut out = DiscreteDblModel::new(toplevel.theory.clone());
-    extract_to(&eval, &mut out, vec![], &elt, ty);
-    out
+    let mut name_translation = HashMap::new();
```
epatters, line 26: Not sure how all the pieces will ultimately fit together, but noting in case you didn't see that there is a [`Namespace`](https://github.com/ToposInstitute/CatColab/blob/b889f9daa5ce68cca05de6f87d149f9edee064d0/packages/catlog/src/zero/qualified.rs#L352) type in `crate::zero::qualified` to do bidirectional lookups between qualified names and qualified labels. Currently, as a result of incremental refactoring, it is used only at the `catlog-wasm` level, but we might want to put that data somewhere else.

epatters (changes requested):
Very nice stuff! Just a few comments below.

packages/catlog/src/tt/modelgen.rs
olynch: Ah, excellent, yes I will use that Namespace type!

packages/catlog/src/tt/mod.rs
olynch: Yes, I believe so, I just had that on in development so it wouldn't be annoying.

packages/catlog/src/tt/context.rs
olynch: The answer is that I wanted to be self-documenting when I was originally writing this, but then I got lazy when I was adding labels everywhere...

I'm happy to have `VarInContext` though.

olynch: If you find this inconsistency ugly, I can add `VarLabel` and `FieldLabel` aliases.

packages/catlog/src/tt/notebook_elab.rs
olynch: I think that the frontend should do the filtering, because then it will be tracked by the solid incremental stuff?

packages/catlog/src/tt/prelude.rs
olynch: Sounds good, to both!

packages/catlog/src/tt/notebook_elab.rs
epatters: Yes, you're right. I got confused because I went back-and-forth on how this was handled in the frontend, but that is what we're currently doing.

packages/catlog/src/tt/modelgen.rs
olynch: Looking more into this, I don't really understand what's going on with `Namespace`. Specifically, why is there both a hashmap and an indexedhashcolumn? Is that for bidirectionality? Why do you only have bidirectionality part of the time? Maybe this should be an explicit enum?

epatters: The short is answer is that we need bidirectionality in the frontend:

- forward direction (name -> label) to *display* a name
- inverse direction (label -> name) to take a name as *input* from the user

epatters: I grant that the data structure itself is a little weird. There might be a better way to do it.

Issue #500, opened 2025-05-12
More interactions with "My Documents" page


tim-at-topos:
There are a few things that it would be nice to be able to do with my documents on the "My Documents" page. In order of (my) personal preference:

- Delete documents
- Re-order the list of documents (by each column)
- Rename documents

tim-at-topos:
moved to roadmap (needs strategic design)

PR #735, opened 2025-09-24
Minimum viable document deletion

kasbah:
I am working on document deletion which is one of the points in #500 . 

- This will be done via soft deletion using a `deleted_at` column in `refs` and possibly on `snapshots` (still need to look into that)

- The initial UI will simply add a delete button to the My Documents table and have a dialog for "are you sure?". When deleting a model

Mockups:

image

image

## Implementation status

So far I have only added the `deleted_at` column via a DB migration. 

## Further future enhancements (possibly going to to be on this PR, or a new one)

- Show the user an informative message when a document has been deleted and they are trying to access it
- Display analysis as a sub-row of its model and give more details on what would be deleted when

Mockups: 

image

Link to mockups: https://excalidraw.com/#json=H0ASC9D2Yg4rHphcXQPjk,20XDpsaAxY0cMZTRw-rZGA

### DIFF
--- a/packages/backend/src/migrations/m20250924133640_add_refs_deleted_at.rs
+++ b/packages/backend/src/migrations/m20250924133640_add_refs_deleted_at.rs
@@ -1,0 +1,46 @@
+use sqlx::{PgConnection, Postgres};
+use sqlx_migrator::Operation;
+use sqlx_migrator::error::Error;
+use sqlx_migrator::migration;
+use sqlx_migrator::vec_box;
+
+pub(crate) struct AddRefsDeletedAt;
+
+migration!(
+    Postgres,
+    AddRefsDeletedAt,
+    "backend",
+    "20250924133640_add_refs_deleted_at",
+    vec_box![],
+    vec_box![MigrationOperation]
+);
+
+struct MigrationOperation;
+#[async_trait::async_trait]
+impl Operation<Postgres> for MigrationOperation {
+    async fn up(&self, conn: &mut PgConnection) -> Result<(), Error> {
+        sqlx::query(
+            "
+            ALTER TABLE refs
+            ADD COLUMN deleted_at TIMESTAMPTZ NULL;
+            ",
+        )
+        .execute(conn)
+        .await?;
+
+        Ok(())
+    }
+
+    async fn down(&self, conn: &mut PgConnection) -> Result<(), Error> {
+        sqlx::query(
+            "
+            ALTER TABLE refs
+            DROP COLUMN deleted_at;
+            ",
+        )
+        .execute(conn)
+        .await?;
+
+        Ok(())
+    }
+}
--- a/packages/backend/src/migrations/mod.rs
+++ b/packages/backend/src/migrations/mod.rs
@@ -7,6 +7,7 @@ mod m20241025030906_users;
 mod m20250409171833_add_permissions_object_subject_idx;
 mod m20250516154702_automerge_storage;
 mod m20250805230408_fix_automerge_storage;
+mod m20250924133640_add_refs_deleted_at;
 
 pub fn migrations() -> Vec<Box<dyn Migration<Postgres>>> {
     vec_box![
@@ -15,5 +16,6 @@ pub fn migrations() -> Vec<Box<dyn Migration<Postgres>>> {
         m20250409171833_add_permissions_object_subject_idx::AddPermissionsObjectSubjectIdx,
         m20250516154702_automerge_storage::AutomergeStorage,
         m20250805230408_fix_automerge_storage::FixAutomergeStorage,
+        m20250924133640_add_refs_deleted_at::AddRefsDeletedAt,
     ]
 }


kasbah:
I amended this to no longer use cascading deletes and show a warning on deleted documents.

image

Because I needed to consider how versioning works, I went on a long detour learning about how the snapshotting is implemented but in the end I went back to just have a deleted_at column on refs.

One smaller UX issue with how this is currently: deleting a document will make the my documents pagination slightly wrong until you refresh the page or navigate away. So if you delete a doc and go to the next page you are skipping the first doc on the second page. I don't think it's a big issue but I thought I would mention it.

packages/backend/src/rpc.rs
```diff
@@ -112,6 +122,16 @@ async fn create_snapshot(ctx: AppCtx, ref_id: Uuid) -> RpcResult<()> {
...
 
+#[handler(mutation)]
+async fn delete_ref(ctx: AppCtx, ref_id: Uuid) -> RpcResult<()> {
+    async {
+        auth::authorize(&ctx, ref_id, PermissionLevel::Own).await?;
+        doc::delete_ref(ctx.state, ref_id).await
+    }
+    .await
+    .into()
+}
```
epatters, line 133: I'm not in the habit of using `async` blocks (as opposed to `async` functions), but I think the body of this function can be simplified to:

```rust
auth::authorize(&ctx, ref_id, PermissionLevel::Own).await?;
doc::delete_ref(ctx.state, ref_id).await.into()
```

I.e., the `async` block is extraneous.

packages/frontend/src/api/types.ts
```diff
@@ -93,6 +95,17 @@ export class Api {
         return permissions;
     }
 
+    /** Check if the document has been deleted. */
+    async checkDocumentIsDeleted(refId: Uuid): Promise<boolean> {
```
epatters: Not a big deal either way, but I like to signpost that a function returns a boolean by having it start with `is`, so that this method name would shorten to `isDocumentDeleted`.

packages/frontend/src/page/document_page.css
```diff
@@ -1,3 +1,20 @@
+.warning-banner {
```
epatters, line 1: Noting that we have generic "alert" components in `src/component/alert.tsx`. I'm not attached to the styling currently there (which IIRC I borrowed from GitHub), but we should consider consolidating this stuff to ensure a consistent look and feel. Happy to defer that to a future PR, though.

packages/frontend/src/user/documents.tsx
```diff
@@ -152,12 +155,17 @@ function DocumentRowsPagination(props: {
     setPage: (p: number) => void;
     pageSize: number;
 }) {
+    const [items, setItems] = createSignal<RefStub[]>(props.items);
```
epatters: Reactivity in Solid is subtle, but eventually you get used to it. In this case, the problem is that you're accessing a prop outside a reactive context. The correct way to write this line is:

```typescript
const [items, setItems] = createSignal<RefStub[]>([]);
createEffect(() => setItems(props.items));
```

packages/frontend/src/user/documents.tsx
```diff
@@ -181,34 +189,127 @@ function DocumentRowsPagination(props: {
...
+            if (result.tag === "Ok") {
+                api.clearCachedDoc(props.stub.refId);
+                props.onDelete();
+            } else {
+                setErrorMessage(`Failed to delete document: ${result.message}`);
+                setShowError(true);
+            }
+        } catch (error) {
+            setErrorMessage(`Error deleting document: ${error}`);
+            setShowError(true);
```
epatters, line 234: No action needed, but noting for future consideration: the way we handle RPC errors in the app is currently a bit random. Often we just throw an error and let it percolate to the top-level error boundary. We should think about our desired strategy and then apply it consistently.

epatters (changes requested):
Thanks Kaspar, this is great! Very clean. A few minor comments below.

packages/backend/src/rpc.rs
kasbah: I was confused by this too but I am following an existing pattern, e.g. `create_snapshot`  and it's necessary for using the `?` operator this way. The async block returns a `Result` which we turn `into()` an `RpcResult` and we can't use `?` to get a `RpcResult` directly.

https://github.com/ToposInstitute/CatColab/blob/b4dcd7e8da716cc3fce3ca2da7620bdefb9d6e98/packages/backend/src/rpc.rs#L106-L113

packages/frontend/src/page/document_page.css
kasbah: Thanks for pointing this out, I did miss that those were available. Using it as is does look worse to me though due to lack of separation from the document text. 

image

vs

image

I could work on making a variant with a background color maybe? I should probably also put the delete buttons into components.

packages/frontend/src/api/types.ts
kasbah: Changed it.

packages/frontend/src/user/documents.tsx
kasbah: Thanks, changed.

packages/backend/src/rpc.rs
jmoggr: This is on me. I had discussed this with Evan a while ago [here](https://github.com/ToposInstitute/CatColab/pull/446#discussion_r2087786685) and never made the followup PR/issue. iirc Kaspar's analysis is correct and I'm still leaning towards applying the everywhere. To track this I created issue https://github.com/ToposInstitute/CatColab/issues/794

packages/frontend/src/page/document_page.css
kasbah: I've decided I will add to and modify our components for this but first I am creating a [Storybook](https://storybook.js.org/)-esque view to help with this sort of work in general (I don't want to actually use Storybook due to size/complexity). I'll open another PR for that first.

packages/backend/src/document.rs
```diff
@@ -118,6 +132,19 @@ pub async fn create_snapshot(state: AppState, ref_id: Uuid) -> Result<(), AppErr
     Ok(())
 }
 
+pub async fn delete_ref(state: AppState, ref_id: Uuid) -> Result<(), AppError> {
```
kasbah, line 136:
```suggestion
/// Soft-deletes a ref by setting `deleted_at`.
pub async fn delete_ref(state: AppState, ref_id: Uuid) -> Result<(), AppError> {
```

packages/backend/src/rpc.rs
epatters: Thanks for chiming in, Jason. I had forgotten about that discussion but now I remember. It makes sense: the `?` operator in Rust is strictly limited to `Option` and `Result`, so we have to do something like this.

I agree also that we should apply this pattern consistently rather than creating dedicated functions for the inner block (as I had originally done).

packages/frontend/src/page/document_page.css
epatters: > Using it as is does look worse to me though due to lack of separation from the document text.

Agreed: your alert looks much better in this context, so happy to stick with it until our component library has settled.

PR #736, opened 2025-09-24
Math docs for the double theory of fibrations

epatters:
Thanks to Corinthia Aberlé for suggesting this theory.

### DIFF
--- a/math-docs/trees/thy-000C.tree
+++ b/math-docs/trees/thy-000C.tree
@@ -1,0 +1,100 @@
+\title{Fibrations}
+\taxon{theory}
+\import{macros}
+
+\p{The \define{theory of fibrations} is presented by
+
+\ul{
+  \li{objects #{b} and #{e}}
+  \li{an arrow #{p: e \to b}}
+  \li{an arrow #{r: \top p^* \to e}}
+  \li{unit and counit cells #{\eta} and #{\varepsilon} plus associated equations
+  making #{r} right adjoint to the arrow #{i: e \to \top p^*} induced by the
+  universal property of the tabulator: \quiver{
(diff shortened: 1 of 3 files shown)


math-docs/trees/thy-000C.tree
```diff
@@ -0,0 +1,100 @@
...
+	\arrow["{\pi_0}"', from=3-1, to=4-1]
+	\arrow["{\pi_1}", from=3-2, to=4-2]
+	\arrow[""{name=5, anchor=center, inner sep=0}, "{p^*}"'{inner sep=.8ex}, "\shortmid"{marking}, from=3-4, to=3-5]
+	\arrow[""{name=6, anchor=center, inner sep=0}, "{p^*}"'{inner sep=.8ex}, "\shortmid"{marking}, from=4-1, to=4-2]
+	\arrow["\varepsilon"{description}, draw=none, from=0, to=4]
+	\arrow["{=}"', draw=none, from=1, to=3]
+	\arrow["\pi"{description}, draw=none, from=2, to=5]
+	\arrow["\pi"{description}, draw=none, from=4, to=6]
+  \end{tikzcd}
+  }
```
KevinDCarlson, line 69: Let's consider a model of this in Prof. OK, you have a functor $P:E\to B$ equipped with the canonical $I:E\to B\downarrow P$, that is, $I(e)=(P(e),e,1_{P(e)}).$ You're asking for a right adjoint $R:B\downarrow P\to E$ to $I,$ which is over $B$, that is, sends $(b,e,x:b\to P(e))$ to some $\bar b$ with $P(\bar b)=b.$ Then $IR(b,e,x)=(b,\bar b,1_b)$ and the counit $\varepsilon$ consists *a priori* of some map $y:b\to b$ and some map $\bar x:\bar b\to e$ such that $P(\bar x)=y\cdot x.$ Now, writing $\pi$ as you do for the canonical cell out of the tabulator, $\pi\circ \epsilon (b,e,x)=y\cdot x$ (it's the *diagonal* of the commutative square defining maps in $B\downarrow P$) so I don't think this axiom is quite what you want. Rather, you want $y=1_b,$ and $y=\pi_0 *\varepsilon.$ So I think you want to be pasting on $\mathrm{id}_{\pi_0}$ above where you currently write $\pi.$ This is nicely consistent with the axiom you write below for $\eta,$ so it shows we have a general way to make adjunctions "over" some guy.

epatters: Thanks Kevin, very helpful. This is precisely the point that I was worrying about. I've fixed this up.

epatters:
I also like your idea to formulate the theory more simply using a rari functor $R$. I wonder why the nLab doesn't do it that way. Maybe I'll follow up later with a variant of the theory formulated along these lines.

KevinDCarlson:
> I also like your idea to formulate the theory more simply using a rari functor R . I wonder why the nLab doesn't do it that way. Maybe I'll follow up later with a variant of the theory formulated along these lines.

I think it's probably just something that's an obscure technical decoration for most purposes but became pretty useful for Riehl and Verity.

Issue #606, opened 2025-07-22
Help page about Petri nets


epatters:
As a followup to #597, we should write a help page about Petri nets in the format of #553.

tim-at-topos:
This should now also include analysis documentation for:

- Reachability https://github.com/ToposInstitute/CatColab/pull/696
- Mass-action https://github.com/ToposInstitute/CatColab/pull/601
- Stochastic mass-action https://github.com/ToposInstitute/CatColab/pull/619

PR #737, opened 2025-09-25
Help page for Petri nets

tim-at-topos:
addresses #606

### DIFF
--- a/packages/frontend/src/help/analysis/stochastic-mass-action.mdx
+++ b/packages/frontend/src/help/analysis/stochastic-mass-action.mdx
@@ -2,8 +2,6 @@ import { A } from "@solidjs/router";
 
 #### User inputs
 
-TO-DO
-
 <dl>
   <dt>Initial value: $\mathbb{R}_{\geqslant0}$</dt>
   <dd>The initial value of the concentration.</dd>
--- a/packages/frontend/src/help/analysis/stochastic-mass-action.mdx
+++ b/packages/frontend/src/help/analysis/stochastic-mass-action.mdx
@@ -2,6 +2,8 @@ import { A } from "@solidjs/router";
 
 #### User inputs
 
+TO-DO
+
 <dl>
   <dt>Initial value: $\mathbb{R}_{\geqslant0}$</dt>
   <dd>The initial value of the concentration.</dd>
--- a/packages/frontend/src/help/logics/petri-net.mdx
+++ b/packages/frontend/src/help/logics/petri-net.mdx
@@ -1,0 +1,35 @@
+import { A } from "@solidjs/router";
+import { HelpAnalysisById } from "../logic_help_detail"
+
+## Description
+
+TO-DO
+
+
+### Examples
+
+- TO-DO
+
+
+### Further reading
+
+- TO-DO
+
+
+## Instances and analyses
+
+### Analyses
+
+<HelpAnalysisById theory={props.theory} analysisId="diagram"/>
+
+<HelpAnalysisById theory={props.theory} analysisId="subreachability">
+TO-DO
+</HelpAnalysisById>
+
+<HelpAnalysisById theory={props.theory} analysisId="mass-action">
+TO-DO
+</HelpAnalysisById>
+
+<HelpAnalysisById theory={props.theory} analysisId="stochastic-mass-action">
+TO-DO
+</HelpAnalysisById>
--- a/packages/frontend/src/help/logics/petri-net.mdx
+++ b/packages/frontend/src/help/logics/petri-net.mdx
@@ -27,7 +27,12 @@ TO-DO
 </HelpAnalysisById>
 
 <HelpAnalysisById theory={props.theory} analysisId="mass-action">
-TO-DO
+Places are interpreted as quantities that can vary in time ("concentrations"), and the transitions ("reactions") in the model determine equations constraining the time derivatives of the stocks.
+Each place is given an **initial value**.
+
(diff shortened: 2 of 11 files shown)


tim-at-topos:
Aside from docs on stochastic mass-action, this is a complete first draft. I figure you might have some thoughts on changing a few sentences @epatters so feel free to leave any feedback (or just made direct changes!)

packages/frontend/src/help/analysis/stochastic-mass-action.mdx
```diff
@@ -0,0 +1,12 @@
...
+
+#### User inputs
+
+<dl>
+  <dt>Initial value: $\mathbb{R}_{\geqslant0}$</dt>
+  <dd>The initial value of the concentration.</dd>
+  <dt>Rate: $\mathbb{R}_{\geqslant0}$</dt>
+  <dd>The rate coefficient ($r$) of the reaction.</dd>
+  <dt>Duration: $\mathbb{R}_{\geqslant0}$</dt>
+  <dd>The total duration of the simulation in units of time.</dd>
```
epatters: I noticed that you removed spurious periods from other definition lists, but a few periods remain here.

(This is pedantry but I appreciate this level of pedantry, so please keep bringing it :)

packages/frontend/src/help/logics/petri-net.mdx
```diff
@@ -0,0 +1,46 @@
+import { A } from "@solidjs/router";
+import { HelpAnalysisById } from "../logic_help_detail"
+
+## Description
+
+**Petri nets** are traditionally a way of expressing discrete-event dynamical systems with concurrency, or networks of resources and processes, describing how various states (called **places**) relate to one another in terms of **transitions**.
```
epatters: Minor point, but perhaps "are traditionally a way of expressing" -> "were invented to express" since "traditionally" always has a faint connotation of outdatedness to me.

packages/frontend/src/help/logics/petri-net.mdx
```diff
@@ -0,0 +1,46 @@
...
+Each transition has incoming and outgoing arrows connected to places, which describe the **input** and **output** places of the transition (respectively).
+
+One way of understanding Petri nets is through their **token semantics**.
+We place a number of "tokens" in each place and then check all the transitions to see if all their input places have a token: if so, then we can "fire" the transition, moving the tokens from the input places to the output places; if not, then nothing happens.
+In many classical references the token semantics are built directly in to the definition of Petri nets (often called a **marking** or **tokening**), but from a categorical point of view --- and, in particular, from the CatColab point of view --- the underlying shape of Petri nets is itself an expressive formal gadget.
```
epatters: As a matter of style, I don't think we should call unnecessary attention to the "categorical point of view" (even if we like it). I'd just say something like "but an unmarked Petri net is already a useful formal gadget, just like it can be useful to consider an ordinary differential equation without giving its initial conditions."

packages/frontend/src/help/logics/petri-net.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+
+### Further reading
+
+- Wikipedia: [Petri net](https://en.wikipedia.org/wiki/Petri_net)
+- Series of blog posts by John Baez on Petri nets, all listed
+  [here](https://johncarlosbaez.wordpress.com/?s=petri)
+
```
epatters, line 20: Best not to overdo it on academic papers in the user docs, but Baez-Pollard is readable enough that it might be worth including here.

packages/frontend/src/help/logics/petri-net.mdx
```diff
@@ -0,0 +1,46 @@
...
+*Note:* we are *not* checking that the forbidden state is reached exactly, just whether *some* state which is strictly larger than the forbidden state can be reached.
+</HelpAnalysisById>
+
+<HelpAnalysisById theory={props.theory} analysisId="mass-action">
+Places are interpreted as quantities that can vary in time ("concentrations"), and the transitions ("reactions") in the model determine equations constraining the time derivatives of the stocks.
+Each place is given an **initial value**.
+
+A transition $A\to T\to B$ between places $A$ and $B$ is interpreted as the equations $\{\dot{A}=-T,\dot{B}=T\}$, i.e. the transition $T$ goes out of $A$ and into $B$.
+Transitions add linearly, i.e. in an arbitrary Petri net, the equation governing a place $A$ is $\dot{A}=\sum_{A_\text{in}}T_i - \sum_{A_\text{out}}T_j$, where $A_\text{in}$ (resp. $A_\text{out}$) is the set of all transitions whose target (resp. source) is $A$.
+Transitions are determined by their **rate** coefficient $r$, i.e. a transition $A\to T\to B$ is defined to satisfy $T=rA$.
```
epatters: Would strongly suggest giving an example at least with two inputs to illustrate that these are _nonlinear_ dynamics. Less important, but also nice, would be to have multiple outputs in your example too.

epatters (commented):
Thanks Tim (and other contributors), this is great! Some comments below. The only one think is actually important is changing the example to illustrate mass-action ODE dynamics.

tim-at-topos:
@epatters thanks for the feedback! I've now merged in Matt's docs for stochastic mass-action, which are slightly more long-form than the other analysis docs I've been writing, but I think that we shouldn't aim for perfect agreement in style and vibe and everything at this stage. I think it's sensible to just have *good* documentation out there rather than none.

epatters (approved):
Thanks for the updates, Tim! I agree that it's better to have some docs than no docs. We can work on finding our "house style" over time.

PR #738, opened 2025-09-25
Links in rich text editor

jmoggr:


packages/frontend/src/components/rich_text_editor/link_editor.tsx
```diff
@@ -0,0 +1,367 @@
...
+type LinkState = {
+    href: string;
+    title: string;
+    linkStart: number;
+    linkEnd: number;
+    left: number;
+    top: number;
+};
+
+function getLinkAtPos(view: EditorView, pos: number): LinkState | null {
```
epatters: This is quite complicated and it's sad that (I presume) ProseMirror doesn't have this functionality built in. How specific is it to links though? Would the same logic be used for inline math?

packages/frontend/src/components/rich_text_editor/link_editor.tsx
```diff
@@ -0,0 +1,367 @@
...
+
+    return (
+        <Show when={props.linkState} fallback={null}>
+            {(linkState) => (
+                <div
+                    class="link-editor-popup"
+                    style={{
+                        position: "absolute",
+                        top: `${linkState().top}px`,
+                        left: `${linkState().left}px`,
```
epatters, line 164: Note to self (no action required): it seems unfortunate to rely on manual coordinate calculations and absolute positioning, but a little searching suggests that this is recommended practice as the ProseMirror tooltip example works similarly: https://prosemirror.net/examples/tooltip/

epatters (changes requested):
Thanks Jason, this is great! I don't have much that's useful to say about the code, but I noticed two bugs/usability problems:

1. If you click on a link with no URL set, a new tab is opened with anyway and it redirects to the current URL. I would expect nothing to happen when you click a link with empty URL
2. If you click a link when the rich text editor is not activated, the rich text tool bar pops up and the click is intercepted. Thus, you can't open links without being in the rich text editor. For me, the expected behavior is that clicking a link in this situation should open the link but not the rich text editor.

packages/frontend/src/components/rich_text_editor/link_editor.tsx
jmoggr: > How specific is it to links though?

Fairly specific, we are using "marks" for links whereas the inline math is using "blocks". Marks are properties on blocks, so there is no single "mark" object that we can grab like we can for a block. I suspect the large majority of custom components that you would be interested in would use blocks.

packages/frontend/src/components/rich_text_editor/link_editor.tsx
jmoggr: I would argue that we are functionally doing the same thing as the example: get the local html coordinates of something in the prosemirror doc and rendering something else with those coordinates relative to a parent. The math is basically the same and we're just being explicit about which div we are positioning relative to.

We could possibly use `.offsetParent.getBoundingClientRect()`, but I wasn't able to quickly get it working and I'm more comfortable/used to doing this kind of positioning explicitly (less magic).

jmoggr:
How we got here:

tl;dr:
- custom things with prosemirror marks should be avoided in favor of blocks
- problems with syncing inline blocks currently make it challenging to build custom components
- this plugin showcases some of the "hacky" things you can do to just-make-something-work
- most of the struggle in working with prosemirror is identifying the shape of a workable solution

I thought I had a happy path to work around the [inline blocks issue in automerge-prosemirror](https://github.com/automerge/automerge-prosemirror/issues/30), however I ran into issue when using backspace on the title field. I was using the wrong method for updating the prosemirror document and I'm surprised it worked at all. Editing was completely broken after changing it to the correct method. This continued problem may have been my fault, but I was feeling pretty tired of second guessing my code vs possible bugs in `prosemirror-automerge`.

I was attempting to implement the link as a prosemirror block, but in the example schemas links are treated as prosemirror marks. I thought that maybe by trying to use blocks as links I was doing something "wrong", so I tried using marks instead. That was an unfortunate choice: marks are much harder and should probably be avoided.

This plugin represents a worst case scenario: we have no tools for working with the marks in the document and no neat way of changing how marks are rendered (mark views seem like a non-starter, but that could again be user error). I just hacked until it worked. Given the context, I don't think the end result is that bad; a similar solution could be reasonably used in the future.

jmoggr:
Issues 1 and 2 should be fixed!

jmoggr:
I'm recalling now that the original intention of implementing links was to test out the fix/workaround for syncing inline blocks with `prosemirror-automerge`. I definitely lost sight of that intention along the way...

packages/frontend/src/components/rich_text_editor/link_editor.tsx
jmoggr: > (I presume) ProseMirror doesn't have this functionality built in

To the best of my knowledge is does not. The only way to collect consecutively marked blocks is to traverse the document.

packages/frontend/src/components/rich_text_editor/link_editor.tsx
epatters: Yep, I agree that we're doing basically the same thing as the example.

PR #746, opened 2025-09-30
Support elaboration to multiple theories

olynch:
Previously, one could only elaborate to a single theory; now multiple theories are supported, along with the logic to reject instantiating a model of theory `X` in a model of theory `Y` when `X != Y`.


epatters (approved):
Thanks Owen, this looks great!

PR #749, opened 2025-10-06
Notion-like paned workflow for all document types

jmoggr:
This PR adds stored procedures for getting max permission levels of a document and getting a list of `RefStub` given a list of `refId`. I have mixed feelings about stored procedures. In this case I think they're worthwhile because I can't see another reasonable way to re-use SQL across different queries.

### DIFF
--- a/packages/frontend/src/page/document_page.tsx
+++ b/packages/frontend/src/page/document_page.tsx
@@ -169,7 +169,7 @@ function SplitPaneToolbar(props: {
     );
 }
 
-function DocumentPane(props: { document: AnyLiveDocument }) {
+export function DocumentPane(props: { document: AnyLiveDocument }) {
     const theories = useContext(TheoryLibraryContext);
     invariant(theories, "Library of theories should be provided as context");
 
--- a/packages/patchwork/src/model_pane.tsx
+++ b/packages/patchwork/src/model_pane.tsx
@@ -2,7 +2,7 @@ import type { AnyDocumentId } from "@automerge/automerge-repo";
 import { createResource, Switch, Match } from "solid-js";
 
 import { createModelLibraryWithRepo } from "../../frontend/src/model";
-import { ModelPane } from "../../frontend/src/model/model_editor";
+import { DocumentPane } from "../../frontend/src/page/document_page";
 import { TheoryLibraryContext } from "../../frontend/src/theory";
 import { stdTheories } from "../../frontend/src/stdlib";
 import type { SolidToolProps } from "./tools";
@@ -36,7 +36,7 @@ export function ModelPaneComponent(props: SolidToolProps) {
                 <Match when={liveModel()}>
                     {(liveModel) => (
                         <TheoryLibraryContext.Provider value={stdTheories}>
-                            <ModelPane liveModel={liveModel()} />
+                            <DocumentPane document={liveModel()} />
                         </TheoryLibraryContext.Provider>
                     )}
                 </Match>


packages/backend/src/document.rs
```diff
@@ -262,53 +263,80 @@ pub async fn search_ref_stubs(
 
     let results = sqlx::query!(
         r#"
-        WITH effective_permissions AS (
```
jmoggr, line 269: I rewrote the filtering logic to divide it into more digestible cases, I found this to be more readable and it should be faster in practice. The logic is the same.

packages/migrator/src/migrations/m20251006141026_get_ref_stubs.rs
```diff
@@ -0,0 +1,111 @@
...
+        )
+        .execute(&mut *tx)
+        .await?;
+
+        tx.commit().await?;
+        Ok(())
+    }
+
+    async fn down(&self, conn: &mut PgConnection) -> Result<(), Error> {
+        sqlx::query("DROP FUNCTION IF EXISTS get_ref_stubs(uuid, uuid[])")
```
epatters, line 107: Should also drop `get_max_permission`?

packages/backend/src/document.rs
```diff
@@ -349,3 +376,150 @@ pub async fn search_ref_stubs(
...
     })
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize, TS)]
+pub struct RelatedRefStub {
+    pub stub: RefStub,
+    pub children: Vec<RelatedRefStub>,
+}
+
+fn find_root_ref<'a>(
```
epatters: I understand why you've added these procedures, but I'd like to think more about where we're implementing this logic, motivated by two concerns:

- That as much as possible we avoid making reference in SQL queries to fields of the JSON content like `diagramIn` or `analysisOf` that are specific to certain document types, since it distributes that logic across the codebase and we have no type checking ability at this point
- In general, we should try to keep our backend as dumb as possible, preferring a local-first strategy based on Automerge. This will help us meet external requirements that CatColab interop with other systems based on Automerge but with a different backend (see #714 for a first step in this direction)

epatters: To get more specific, we could start by moving the procedure `find_root_ref` from the backend to the frontend.

The apparent downside to this suggestion---that we'll have to hit the backend with more calls to `rpc.get_doc` while fetching more data than necessary---is obviated by the following points:

- The parent documents will usually get loaded anyway because doing anything with a diagram or analysis typically requiring loading the parent model or diagram
- With the caching in #756, these fetches won't require hitting the backend at all, because they'll be pulled from Automerge local storage!

packages/backend/src/document.rs
```diff
@@ -349,3 +376,150 @@ pub async fn search_ref_stubs(
...
+            };
+            let children = fetch_children(db, user_id.clone(), stub.ref_id).await?;
+            result.push(RelatedRefStub { stub, children });
+        }
+
+        Ok(result)
+    })
+}
+
+pub async fn get_related_ref_stubs(ctx: AppCtx, ref_id: Uuid) -> Result<RelatedRefStub, AppError> {
```
epatters: The case of `get_related_ref_stubs` (and `fetch_children` that it's based on) is trickier because we clearly need to hit the backend to follow the back references. So I don't expect these Rust/SQL procedures to go away.

OTOH, I'm still hopeful that we can eliminate the specific mentions of `diagramIn` and `analysisOf` because I designed the [`StableRef`](https://github.com/ToposInstitute/CatColab/blob/3d180ebe7460778f53b9f6cf04750a5520cc52c4/packages/notebook-types/src/v0/api.rs#L17)/[`Link`](https://github.com/ToposInstitute/CatColab/blob/3d180ebe7460778f53b9f6cf04750a5520cc52c4/packages/notebook-types/src/v0/api.rs#L43) types to have a special form (inspired by document DBs) that would make them detectable anywhere in our documents without having to know the document schema. Namely, any object not at the top level containing a field called `_id` should be interpreted as a cross-reference/link to another document. So, I think we could rewrite `get_related_ref_stubs` to iterate over the top-level fields of the JSON data and extract any link detected, irrespective of their type. We would thus be implicitly defining the "related" documents as those referencing it from the top-level.

packages/migrator/src/migrations/m20251006141026_get_ref_stubs.rs
jmoggr: done

packages/backend/src/document.rs
jmoggr: I've added a new backend function which just returns children of a given ref. This does result in more round trips to the server, but the loading speed of the sidebar contents should hopefully not be too important.

The definition of a child is now: any document snapshot which has an object anywhere inside it which contains fields: `_id` = parent.id and `type` = `diagram-in | analysis-of`

packages/backend/src/document.rs
```diff
@@ -349,3 +374,59 @@ pub async fn search_ref_stubs(
...
+            stubs.name,
+            stubs.type_name,
+            stubs.created_at       AS "created_at!",
+            stubs.permission_level AS "permission_level!: PermissionLevel",
+            stubs.owner_id,
+            stubs.owner_username,
+            stubs.owner_display_name
+        FROM
+            child_refs,
+            get_ref_stubs($2, child_refs.child_ids) AS stubs
```
jmoggr, line 413: Currently this use of `get_ref_stubs` is overkill since we're only using `ref_id` on the client. The original idea was to ensure that whenever the server was returning a minimal reference to a ref it would return a `RefStub` to avoid a proliferation of utility types containing a subset of the same information. I'm no longer sure this is worth doing.

epatters: Thanks for pointing this out. From the point of view of efficiency, it seems a small matter to return `RefStub`s instead of just ref IDs. Is the issue whether it's worth having a stored function `get_ref_stubs` if we only actually need it in one query, at least right now? I don't know. I leave it to your judgment.

packages/backend/src/document.rs
```diff
@@ -349,3 +374,59 @@ pub async fn search_ref_stubs(
...
+        r#"
+        WITH child_refs AS (
+            SELECT ARRAY_AGG(refs.id) AS child_ids
+            FROM refs
+            JOIN snapshots ON snapshots.id = refs.head
+            WHERE (
+                get_max_permission($2, refs.id) >= 'read'
+                AND jsonb_path_exists(
+                    snapshots.content,
+                    '$.*[*] ? (@._id == $id && (@.type == "diagram-in" || @.type == "analysis-of"))',
```
epatters: Playing around in a JSON Path editor, I believe that the first part of this query, namely

```
$.*[*] ? (@._id == $id)
```

will find any top-level object value containing an `_id` key with the given value. Is that right? If so, great. Can you then:

- add a comment or docstring explaining that
- remove the clause `(@.type == "diagram-in" || @.type == "analysis-of")`: for now I would rather be under-specific than over-specific and it's nice to avoid hard-coding these types in the query

packages/frontend/src/analysis/analysis_editor.tsx
```diff
@@ -1,180 +1,23 @@
...
+    LiveAnalysisDocument,
+    LiveDiagramAnalysisDocument,
+    LiveModelAnalysisDocument,
 } from "./document";
 import type { Analysis } from "./types";
 
-import PanelRight from "lucide-solid/icons/panel-right";
-import PanelRightClose from "lucide-solid/icons/panel-right-close";
-
-export default function AnalysisPage() {
```
epatters, line 43: Love to see all this specialized code deleted!

packages/frontend/src/page/document_type_icon.tsx
epatters, line 1: We should try to avoid putting stuff in `util` since it tends to become a dumping ground for random stuff. Ideally we'll eventually get rid of `util` entirely.

More to the point, in this case, I think we should move this component to `frontend/src/page/`.

packages/frontend/src/page/page_container.css
epatters, line 1: This file being added is empty. Should be removed.

packages/frontend/src/page/layout.tsx
```diff
@@ -0,0 +1,75 @@
...
+                <IconButton onClick={props.openSidebar}>
+                    <MenuIcon />
+                </IconButton>
+            )}
+            {props.children}
+        </header>
+    );
+}
+
+function IconButton(props: {
```
epatters: We've got an `IconButton` in `src/components/icon_button.tsx` which includes a nice tooltip feature. You should be able to use that one and delete this one.

packages/frontend/src/page/layout.tsx
```diff
@@ -0,0 +1,75 @@
+import { type JSX, createSignal } from "solid-js";
+
+import "./layout.css";
+
+import ChevronsLeft from "lucide-solid/icons/chevrons-left";
+import MenuIcon from "lucide-solid/icons/menu";
+
+export function Layout(props: {
```
epatters: Would suggest a more descriptive name, such as `PaneWithSidebar`.

packages/frontend/src/page/document.tsx
```diff
@@ -0,0 +1,236 @@
...
+import { TheoryLibraryContext } from "../theory";
+import { PermissionsButton } from "../user";
+import { assertExhaustive } from "../util/assert_exhaustive";
+import { DocumentSidebar } from "./document_sidebar";
+
+import "./document.css";
+
+type AnyLiveDocument = LiveModelDocument | LiveDiagramDocument | LiveAnalysisDocument;
+
+export default function Document() {
```
epatters: Between our document type and Automerge's, the name `Document` is already overloaded. We should call this component `DocumentPage` or something similar.

packages/frontend/src/analysis/analysis_widget.tsx
```diff
@@ -0,0 +1,34 @@
+import { A } from "@solidjs/router";
+
+import type { LiveAnalysisDocument } from "./document";
+
+/** Widget in the top right corner of a diagram document pane.
+ */
+export function AnalysisWidget(props: { liveAnalysis: LiveAnalysisDocument }) {
```
epatters: The word "widget" can mean pretty much anything in an UI, so we should find something more specific. Not immediately sure what, perhaps something like `AnalysisInfo`.

Ditto for `ModelWidget` and `DiagramWidget`.

packages/frontend/src/model/model_editor.tsx
```diff
@@ -34,82 +24,6 @@ import {
...
-                <PermissionsButton liveDoc={props.liveModel.liveDoc} />
-            </Toolbar>
-            <ModelPane liveModel={props.liveModel} />
-        </div>
-    );
-}
-
-/** Pane containing a model notebook plus a header with the title and theory.
- */
-export function ModelPane(props: {
```
epatters, line 74: Deleting this component and `AnalysisPane` will break the corresponding components in `packages/patchwork`. I've created #779 to make such problems more obvious in the future.

Can you update the Patchwork module as appropriate? I'm hoping that's as simple as using your new `DocumentPane`.

packages/frontend/src/page/document_sidebar.tsx
```diff
@@ -0,0 +1,161 @@
...
+function DocumentsTreeNode(props: {
+    liveDoc: LiveDoc;
+    indent: number;
+    currentLiveDoc: LiveDoc;
+    refreshCounter: number;
+    triggerRefresh: () => void;
+}) {
+    const api = useApi();
+
+    const [childDocs] = createResource(
```
epatters: I believe the more idiomatic way to manually cause a resource to refresh is to retain the refetch callback

```typescript
const [childDocs, { refetch }] = createResource(...)
```

and then call `refetch()` when needed. See: https://docs.solidjs.com/reference/basic-reactivity/create-resource

packages/frontend/src/page/document_sidebar.tsx
```diff
@@ -0,0 +1,161 @@
...
+        () => [props.liveDoc, props.refreshCounter] as const,
+        async ([liveDoc, _counter]) => {
+            const docRefId = liveDoc.docRef?.refId;
+            invariant(docRefId, "Doc must have a valid ref");
+
+            const results = await api.rpc.get_ref_children_stubs.query(docRefId);
+
+            if (results.tag !== "Ok") {
+                throw "couldn't load child documents!";
```
epatters: `throw ...` -> `throw new Error(...)`

packages/frontend/src/page/layout.css
```diff
@@ -0,0 +1,179 @@
...
+    height: 28px;
+    width: 28px;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    border-radius: 6px;
+    transition: background 20ms;
+}
+
+.icon-button:hover > div {
```
epatters: Reminder to remove the CSS for `icon-button` too.

packages/backend/src/document.rs
epatters: Oh wait, now I see your comment above:

> The definition of a child is now: any document snapshot which has an object anywhere inside it which contains fields: _id = parent.id and type = diagram-in | analysis-of

which disagrees with the "top-level" part of what I wrote above. Are you sure about your interpretation of the JSON Path?

packages/backend/src/document.rs
jmoggr: Let's keep it in for now, Kaspar was planning on using it for something. It's easy to remove later if it's causing friction and it's nice to have a stored procedure to call when debugging things.

packages/frontend/src/page/document_sidebar.tsx
jmoggr: nice! React brain...

packages/backend/src/document.rs
jmoggr: You are correct, good catch! I added the comment with the correct explanation to `get_ref_children_stubs`.

kasbah:
There is something wrong with the truncation in the tree menu when names have dashes in them:

image

EDIT: Would also suggest a bit more spacing between the hamburger menu and the document name. I have a lot more UX quibbles but I don't want to derail this right now.

EDIT2: I'll just mention one more. This looks like a back button and the expectation is "take me back to my documents list".

image

epatters:
Thanks for the comments, Kaspar. Due to the long lifespan of this PR (plus its predecessor), I'm aiming to merge this promptly, but if you have ideas about how to improve the UI, I'd love to get follow-up PR(s) along those lines.

PR #751, opened 2025-10-06
Morphisms between diagrammatic instances: data structures and validation

kris-brown:
This PR introduces a data structure for morphisms of diagrams $D:J \rightarrow C$ to $D': J'\rightarrow C$, i.e. morphisms of instances of double models. We only treat morphisms of discrete double models.

The key features associated with this are checking whether a purported diagram morphism is natural and whether two diagrammatic morphisms are presentations of the same instance morphism.

### DIFF
--- a/packages/catlog/src/dbl/model_diagram.rs
+++ b/packages/catlog/src/dbl/model_diagram.rs
@@ -31,11 +31,10 @@ use crate::one::{
 use crate::validate::{self};
 use crate::zero::{HashColumn, QualifiedName, column::MutMapping};
 
-/** A diagram in a model of a double theory.
-
-This struct owns its data, namely, the domain of the diagram (a model) and the
-model mapping itself.
-*/
+/// A diagram in a model of a double theory.
+///
+/// This struct owns its data, namely, the domain of the diagram (a model) and the
+/// model mapping itself.
 #[derive(Clone, Into, PartialEq)]
 #[into(owned, ref, ref_mut)]
 pub struct DblModelDiagram<Map, Dom>(pub Map, pub Dom);
@@ -135,24 +134,24 @@ impl DiscreteDblModelDiagram {
     /// morphism `D(f): D(j0)→D(j1)` in `𝒞`. This adds a *single* element to
     /// the connected component because the following triangle commutes:
     ///
-    ///```text
+    /// ```text
     ///                D(f)
     ///           D(j0) -> D(j1)
     ///         D(f);m ↘   ↙ m
     ///                  c
-    ///```
+    /// ```
     ///
     /// Move 2: there is a morphism `f: j1→j2` in `J`. Therefore there is a
     /// morphism `D(f): D(j1)→D(j2)` in `𝒞`. We then have new `𝒞/c` objects in
     /// our connected component for *every* `m'` in `𝒞` such that the following
     /// diagram in `𝒞` commutes:
     ///
-    ///```text
+    /// ```text
     ///               D(f)
     ///            D(j1) -> D(j2)
     ///           m  ↘   ↙  m'
     ///                c
-    ///```
+    /// ```
     ///
     /// We represent elements of `𝒞/c` with a path of generating morphisms in
     /// `𝒞`. There is a possibility of redundant work because multiple paths of
@@ -289,10 +288,9 @@ impl DiscreteDblModelDiagram {
 }
 
 impl<'a> InstanceMorphism<'a> {
-    /** Validates that the instance morphism is well-defined.
-
-    Assumes that the dom/codom diagrams are valid. If not, this may panic.
-     */
+    /// Validates that the instance morphism is well-defined.
+    ///
+    /// Assumes that the dom/codom diagrams are valid. If not, this may panic.
     pub fn validate_in(
         &self,
         model: &DiscreteDblModel,
@@ -492,7 +490,7 @@ mod tests {
     /// The letters before names indicate whether something is an entity,
     /// attrtype, hom, or attr.
     ///
-    ///```text
+    /// ```text
     ///    D1(J)  D2(J')
     ///    -----  ------
     ///
@@ -505,18 +503,18 @@ mod tests {
     ///     v  v ✓ | h01
     ///     a0 <- e0
     ///       a00
-    ///```
+    /// ```
     ///
     /// The bottom triangle commutes always, and the input parameter `commutes`
     /// controls whether or not the top square commutes.
     ///
     /// Viewed as tabular instances, D1(J) is:
     ///
-    ///```text
+    /// ```text
     /// | e0 |  | e1 |  | e2 | | a1 | o10 |  |  a0 |  
     /// ======  ======  ====== ============  =======
     ///                        |ja1 | ja0 |  | ja0 |   
-    ///```
+    /// ```
     ///
     /// And `D2(J')` (if both polygons in `C` commute) is the terminal instance:
     ///
@@ -528,7 +526,7 @@ mod tests {
     /// |    a1   |   o10   |     |    a0    |  
     /// =====================     ============
     /// |a21(je2) | a00(je0)|     | a00(je0) |  
-    ///```
+    /// ```
     ///
     /// However, if the top square in `C` does not commute, `D2(J')` becomes:
     ///
@@ -541,7 +539,7 @@ mod tests {
     /// =====================     ================
     /// |a21(je2) | a00(je0)|     |    a00(je0)  |  
     ///                           | o10(a21(je1))|
-    ///```
+    /// ```
     ///
     /// Returns a tuple: `(C,D1,D2)`.
     fn create_diagrams(


packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
     /// Mapping underlying the diagram is invalid.
     Map(MapErr),
 }
+
+/// A morphism between instances (presented as diagrams in some model of a
+/// double theory) consists in an assignment, for each representable `j` in the
```
KevinDCarlson, line 57: "representable" doesn't really parse here for me. Is it not just for any _object_?

KevinDCarlson: Generally I think this will need to name the instances to be parseable.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+/// domain instance, of a choice of representable `j'` in the codomain instance
+/// plus a morphism in the underlying category out of the representing object of
+/// `j'`.
+///
+/// For example, consider the schema for graphs as a model of the trivial double
+/// theory. One sends some domain representable vertex `v` to the source of some
+/// codomain representable edge, `e`. There may be no outgoing morphism for `e`
```
KevinDCarlson, line 64: I don't really follow this either.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+/// in the codomain shape category, but it's implicitly there because the the
+/// morphism data of the v component comes from the underlying category.
+///
+///
+/// Let `D: J → 𝒞` and `D': J' → 𝒞` be two diagrams. For any `j' ∈ J'`, let
+/// `D'j'` be a functor `1 → 𝒞`, picking out `D'(j') ∈ 𝒞`. Consider the comma
```
KevinDCarlson, line 70: It's a bit overwrought to distinguish between D'j' and D'(j') like this.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+/// category `D'j' ↓ D`:
+/// - Its objects are pairs `(j ∈ J, D'j' → Dj ∈ Hom 𝒞)`
+/// - Its morphisms are triples `(f: j₁ → j₂, s: D'j → Dj₁, t: D'j → Dj₂)` such
+///   that `s;Df = t`
+///
```
KevinDCarlson, line 75: You started out talking about instance morphisms, but now the rest just looks like morphisms of diagrams in categories? 

Anyway, this is a bit heavy for a docstring, I think--I need to write it up in the math docs this week and then this can point there, hopefully.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+/// We are interested in `Hom(D, D')`, the morphisms between the instances
+/// presented by `D` and `D'`. This is equal to `lim(π₀(D'j' ↓ D))` The
+/// solutions to the limit problem are choices, for each `j ∈ J`, of some object
+/// of `π₀(D'j' ↓ D)`, satisfying a universal property. These choices are the
+/// components of the instance morphism.
+///
+/// This struct borrows its data to perform validation. The domain and codomain
+/// are assumed to be valid models of double theories. If that is in question,
+/// the models should be validated *before* validating this object.
+pub struct InstanceMorphism<'a>(
```
KevinDCarlson, line 92: I think if this is going to be in `Dbl::ModelDiagram`, it has to be  parametetric, like `Dbl::ModelMorphism`. If it's going to have `Discrete` inside of it, then it should be in `Discrete::ModelDiagram`.

KevinDCarlson: Ah, `DblModelMorphism` is for some reason in `Discrete` itself, but then all the moreso this seems like it's in the wrong place.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+    HashColumn<QualifiedName, (QualifiedName, QualifiedPath)>,
```
KevinDCarlson, line 93: I guess it wasn't possible to get this guy using `FpFunctorData`?

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+/// MissingComponent marks if one is missing, whereas IllSpecifiedComponent
+/// marks that the component is improperly specified somehow.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(tag = "tag", content = "err"))]
+#[cfg_attr(feature = "serde-wasm", derive(Tsify))]
+#[cfg_attr(feature = "serde-wasm", tsify(into_wasm_abi, from_wasm_abi))]
+pub enum InvalidInstanceMorphism {
+    /// Domain of the diagram is invalid.
+    MissingComponent(QualifiedName),
```
KevinDCarlson, line 108: The docstring and the enumerant name seem like different things to me. We're not allowing for the domain being invalid to fall through to even building an instance morphism, are we?

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+    /// Mapping underlying the diagram is invalid.
+    IllSpecifiedComponent(String, QualifiedName),
+
+    /// Mapping underlying the diagram is unnatural: (f: a->b, α(a), α(b)).
+    UnnaturalComponent(QualifiedName),
+}
+
+impl DiscreteDblModelDiagram {
+    /// Because morphisms within the shape `J` of a diagram `J → 𝒞` (viewed as
+    /// an instance) are representing equalities, they can be traversed in
```
KevinDCarlson, line 119: I might say "are representing equalities in the generated instance" rather than the slightly vaguer "(viewed as an instance)". Or maybe I'm just bad at reading.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+    /// either direction. Thus the same object `c` in `𝒞` be presented in
+    /// multiple different ways, `(j1,D(j1)→c)` and `(j2,D(j2)→c)`, and showing
```
KevinDCarlson, line 121: I don't really see how objects of $D\downarrow \mathcal{C}$ present objects of $\mathcal{C}$...

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+    /// they are different presentations of the same object involves traversing
+    /// these morphisms in either direction.
+    ///
+    /// Given a diagram `D: J → 𝒞`, and a pair of `𝒞/c` objects `src: D(j1)→c`
```
KevinDCarlson, line 125: Some ungrammatical constructions here

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+    /// There are some concerns about the accuracy of this approach when `𝒞`
+    /// has loops, as it is not possible to fully enumerate all paths of
+    /// morphism generators between two objects in `𝒞`. The `len` keyword
+    /// prevents us from infinite iteration at the cost of some false negatives.
+    /// It sets the maximum length of the objects in `𝒞/c` that
+    /// we use to iteratively expand the connected component. Likewise, to
+    /// prevent infinite iteration, `ziglen` controls the max length of zigzags.
+    ///
+    /// Note this algorithm does not need to look at the object / morphism
+    /// *types* in `𝒞`.
```
KevinDCarlson, line 172: Note that people will have to try pretty hard to see this beautiful docstring rendered on a private function!

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+    /// the connected component because the following triangle commutes:
+    ///
+    /// ```text
+    ///                D(f)
+    ///           D(j0) -> D(j1)
+    ///         D(f);m ↘   ↙ m
+    ///                  c
+    /// ```
+    ///
+    /// Move 2: there is a morphism `f: j1→j2` in `J`. Therefore there is a
```
KevinDCarlson, line 144: Surely you want to introduce a j3 here, not jump all the way from src to tgt?

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+    /// ```
+    ///
+    /// We represent elements of `𝒞/c` with a path of generating morphisms in
+    /// `𝒞`. There is a possibility of redundant work because multiple paths of
+    /// generating morphisms may be equal to the same morphism; however, there
+    /// is no canonical representative for such equivalence classes, and working
+    /// with paths of generating morphisms allows us to control the max length
+    /// of morphism we are willing to consider.
+    ///
+    /// There are some concerns about the accuracy of this approach when `𝒞`
```
KevinDCarlson, line 163: "there are some concerns" is a bit weaselly...

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+}
+
+impl DiscreteDblModelDiagram {
+    /// Because morphisms within the shape `J` of a diagram `J → 𝒞` (viewed as
+    /// an instance) are representing equalities, they can be traversed in
+    /// either direction. Thus the same object `c` in `𝒞` be presented in
+    /// multiple different ways, `(j1,D(j1)→c)` and `(j2,D(j2)→c)`, and showing
+    /// they are different presentations of the same object involves traversing
+    /// these morphisms in either direction.
+    ///
```
KevinDCarlson, line 124: The below paragraph should come first; Evan was explaining to me that people will often see a "short docstring" rendered which is everything down to the first blank line.

packages/catlog/src/dbl/model_diagram.rs
KevinDCarlson: Not sure it's quite clear what type $\mathcal{C}/c$ has right now. Are you passing to underlying categories or taking a comma object in the diagram category?

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+    ///            D(j1) -> D(j2)
+    ///           m  ↘   ↙  m'
+    ///                c
+    /// ```
+    ///
+    /// We represent elements of `𝒞/c` with a path of generating morphisms in
+    /// `𝒞`. There is a possibility of redundant work because multiple paths of
+    /// generating morphisms may be equal to the same morphism; however, there
+    /// is no canonical representative for such equivalence classes, and working
+    /// with paths of generating morphisms allows us to control the max length
```
KevinDCarlson, line 160: I'm confused about why you want to control length on the objects of $\mathcal{C}/c$ rather than the morphisms. Seems like it would be more natural to compose these lists up when they're representing the formal, wouldn't it?

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+        // have seen before).
+        let mut seen: HashSet<(QualifiedName, QualifiedPath)> = HashSet::new();
+
+        let jg = self.1.generating_graph();
+        let g = model.generating_graph();
+
+        // This is the distinguished object in 𝒞 that we care about. We are
+        // only interested in 𝒞/c for the purposes of determining whether `src`
+        // and `tgt` are mutually reachable.
+        let c = src.tgt(g);
```
KevinDCarlson, line 211: `src.tgt` is a distressing sequence of characters. I suggest not using `src` and `tgt` for names of the possibly-linked guys.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+    }
+
+    /// Determine if two presentations of instance morphisms present the same
+    /// instance morphism. This is testing whether, for fixed `D` and `D'`,
+    /// whether the assignments given by the underlying mapping of instance
+    /// morphism produce equal instance morphisms. This presumes that both
+    /// inputs have been validated: it may give incorrect results or panic
+    /// otherwise.
+    ///
+    /// Note this is very strict> it requires domain and codomain categories of
```
KevinDCarlson, line 395: I'm not really bothered by this strictness. It's normal to have an identity-for-parallel-arrows algorithm.

packages/catlog/src/dbl/model_diagram.rs
KevinDCarlson: I wonder whether these fields should be named, since they get accessed so much in the zigzag algorithm. I'm also confused why the mapping is private while the domain and codomain are pub?

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+        // only interested in 𝒞/c for the purposes of determining whether `src`
+        // and `tgt` are mutually reachable.
+        let c = src.tgt(g);
+
+        let dj1 = f.ob_generator_map().get(&j1).unwrap();
+        let dj2 = f.ob_generator_map().get(&j2).unwrap();
+
+        // Confirm that src and tgt are a cospan D(j1) → c ← D(j2)
+        if *dj1 != src.src(g) {
+            panic!("Bad input D({j1:?}) = {dj1:?} != src({src:?}")
```
KevinDCarlson, line 218: I don't think these should panic, this is going to get called from a public method. Return Result, probably? There's some fancier stuff about ThisError in this crate that I don't know much about but might be even better.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(tag = "tag", content = "err"))]
+#[cfg_attr(feature = "serde-wasm", derive(Tsify))]
+#[cfg_attr(feature = "serde-wasm", tsify(into_wasm_abi, from_wasm_abi))]
+pub enum InvalidInstanceMorphism {
+    /// Domain of the diagram is invalid.
+    MissingComponent(QualifiedName),
+
+    /// Mapping underlying the diagram is invalid.
+    IllSpecifiedComponent(String, QualifiedName),
```
KevinDCarlson, line 111: This hocking of a string into a variant is pretty sketchy compared to what we have in, say `InvalidDblModelMorphism`. I think it should be more like that, probably using `#derive(Error)`.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+
+        for j in cat_j.ob_generators() {
+            let Some((j_, f)) = self.0.get(&j) else {
+                errs.push(InvalidInstanceMorphism::MissingComponent(j));
+                continue;
+            };
+
+            let Some(dj) = d.get(&j) else {
+                errs.push(InvalidInstanceMorphism::IllSpecifiedComponent(
+                    format!("component {j:?} wrong "),
```
KevinDCarlson, line 327: I wouldn't want a user to get these error messages, again cf InvalidModelMorphism.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+            Some(n) => Box::new(0..n + 1),
+        };
+
+        for _ in iter {
+            let mut next_worklist: Vec<(QualifiedName, QualifiedPath)> = vec![];
+
+            // Process worklist until it is empty, building up next round's worklist
+            while let Some((j, m)) = worklist.pop() {
+                // Check if this pair is the tgt we have been trying to reach
+                if model.morphisms_are_equal(m.clone(), tgt.clone()) {
```
KevinDCarlson, line 236: Lots of clones but not fixable within this PR.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+
+        let dj1 = f.ob_generator_map().get(&j1).unwrap();
+        let dj2 = f.ob_generator_map().get(&j2).unwrap();
+
+        // Confirm that src and tgt are a cospan D(j1) → c ← D(j2)
+        if *dj1 != src.src(g) {
+            panic!("Bad input D({j1:?}) = {dj1:?} != src({src:?}")
+        } else if *dj2 != tgt.src(g) {
+            panic!("Bad input D({j2:?}) = {dj2:?} != src({tgt:?}")
+        } else if c != tgt.tgt(g) {
```
KevinDCarlson, line 221: Don't we also need to check `c != src.tgt`?

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+        // not `src` and `tgt` are in the same connected component does not need
+        // us to track which particular sequence of zigs and zags led us to this
+        // pair, so there is no need to track this data (any zigzag with a
+        // cycle can be replaced by a zigzag without a cycle, thus we can
+        // immediately stop when our search comes across an element of `𝒞/c` we
+        // have seen before).
+        let mut seen: HashSet<(QualifiedName, QualifiedPath)> = HashSet::new();
+
+        let jg = self.1.generating_graph();
+        let g = model.generating_graph();
```
KevinDCarlson, line 206: `g` gets used a bunch to find the src and tgt of a path, but you can just use `model.cod(&src)` and stuff like that.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+        } else if *dj2 != tgt.src(g) {
+            panic!("Bad input D({j2:?}) = {dj2:?} != src({tgt:?}")
+        } else if c != tgt.tgt(g) {
+            panic!("Bad input tgt({tgt:?})!={c:?}")
+        }
+        // Iterate ziglen many times, or unboundedly if ziglen is none
+        let iter: Box<dyn Iterator<Item = usize>> = match ziglen {
+            None => Box::new(0..),
+            Some(n) => Box::new(0..n + 1),
+        };
```
KevinDCarlson, line 228: I think it's much easier to do something like this:
```
for round in 0.. {
      if ziglen == Some(round) {                                  
          break;                                                              
      }                                                                       
      // ...                                                                  
  }                                                                           
 ```

File: packages/catlog/src/dbl/model_diagram.rs [hunk 25]
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+
+        for _ in iter {
+            let mut next_worklist: Vec<(QualifiedName, QualifiedPath)> = vec![];
```
KevinDCarlson, line 231: Probably shouldn't allocate a fresh worklist every time through the loop. Allocate this next to worklist, and then I hear you can use `std::mem::swap` to freely put `next_worklist` into `worklist` instead of manually rebuilding it below.

packages/catlog/src/dbl/model_diagram.rs
```diff
@@ -38,3 +52,664 @@ pub enum InvalidDblModelDiagram<DomErr, MapErr> {
...
+        // composite)
+        let mut worklist: Vec<(QualifiedName, QualifiedPath)> = vec![(j1.clone(), src.clone())];
+
+        // If we ever come across a pair `(j, D(j) → c)` we have seen before,
+        // then we do not need to add it to the worklist. Determining whether or
+        // not `src` and `tgt` are in the same connected component does not need
+        // us to track which particular sequence of zigs and zags led us to this
+        // pair, so there is no need to track this data (any zigzag with a
+        // cycle can be replaced by a zigzag without a cycle, thus we can
+        // immediately stop when our search comes across an element of `𝒞/c` we
```
KevinDCarlson, line 201: "can immediately stop" makes it sound like you're going to stop the whole search, not just do a normal BFS pruning.

PR #765, opened 2025-10-21
Fix failing integration test and improve error handling

jmoggr:
The frontend tests were failing because they were creating outdated documents:
```
{
    type: "model"
    name: "name"
}
```

These invalid models were still getting added to automerge, and were failing to parse during the migration that occurs when we attempt to connect to a ref. The errors logged by `automerge-doc-server` were not very useful and similar to the errors recently seen on the `catcolab-next`.

Changes:
- Validate document contents in `new_ref`
- Validate document contents `createDoc`
- Improve handling of uncaught error in `automerge-doc-server`
- Update tests to use the new `Document` type

Since we did not have these guards previously, it is likely that `catcolab-next` has invalid documents, production might as well. This could explain the errors seen on `catcolab-next`. I will investigate this further.

Note:
I was running into some difficulty with the different sources for our generated typescript types:
- `ts_rs` for `backend`
- `tsify` for `frontend`

In rust I can't write:
```
async fn new_ref(ctx: AppCtx, document: VersionedDocument)
```
since the `TS` trait from `ts_rs` is not implement on `VersionedDocument`. 

We could implement `TS` on everything that currently implements `Tsify`, but that would still generate types in two different places and they are not guaranteed to be the same (though I see no reason why they wouldn't be).

This is only an issue for some rpc endpoints that deal with more complicated types. Given that, the surface area for this problem is quite small, so living with it could be less painful than a fix, especially since I see no obvious fix.

### DIFF
--- a/packages/automerge-doc-server/src/server.ts
+++ b/packages/automerge-doc-server/src/server.ts
@@ -15,10 +15,20 @@ import type { Pool as PoolType } from "pg";
 import { PostgresStorageAdapter } from "./postgres_storage_adapter.js";
 import type { NewDocSocketResponse, StartListeningSocketResponse } from "./types.js";
 import type { SocketIOHandlers } from "./socket.js";
+import { serializeError } from "./socket.js";
 import jsonpatch from "fast-json-patch";
 
 // Load environment variables from .env
 dotenv.config();
+
+/** Attempt to migrate a document, returning the migrated document or an error message. */
+function migrateDocument(doc: unknown): { Ok: any } | { Err: string } {
+    try {
+        return { Ok: notbookTypes.migrateDocument(doc) };
+    } catch (e) {
+        return { Err: `Failed to migrate document: ${serializeError(e)}` };
+    }
+}
 
 export class AutomergeServer implements SocketIOHandlers {
     private docMap: Map<string, DocHandle<unknown>>;
@@ -66,7 +76,12 @@ export class AutomergeServer implements SocketIOHandlers {
     }
 
     async createDoc(content: unknown): Promise<NewDocSocketResponse> {
-        const handle = this.repo.create(content);
+        const migrateResult = migrateDocument(content);
+        if ("Err" in migrateResult) {
+            return migrateResult;
+        }
+
+        const handle = this.repo.create(migrateResult.Ok);
         if (!handle) {
             return {
                 Err: "Failed to create a new document",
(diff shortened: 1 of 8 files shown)


packages/automerge-doc-server/src/server.ts
```diff
@@ -15,11 +15,21 @@ import type { Pool as PoolType } from "pg";
 import { PostgresStorageAdapter } from "./postgres_storage_adapter.js";
 import type { NewDocSocketResponse, StartListeningSocketResponse } from "./types.js";
 import type { SocketIOHandlers } from "./socket.js";
+import { serializeError } from "./socket.js";
```
epatters, line 18: That imports from `socket.js` appear across two lines made me realize that we need to:

- #768

No action needed here: this and likely other small nits should get fixed along with #768.

packages/automerge-doc-server/src/server.ts
```diff
@@ -133,14 +148,18 @@ export class AutomergeServer implements SocketIOHandlers {
         //
         // XXX: frontend/src/api/document.ts needs to be kept up to date with this
```
epatters, line 149: Given the comment, I have to ask the obvious question: does the frontend migration code need to be updated in light of this change?

I'm not blocking this PR on that, but if it is a problem, please take appropriate action by filing an issue or making a followup PR.

epatters (approved):
Thanks for the fixes, Jason! Merging now but note the two comments below.

> I was running into some difficulty with the different sources for our generated typescript types:
>
> - ts_rs for backend
> - tsify for frontend

Yes, it's unfortunate that the code paths for generating type defs for the backend and Wasm bindings are completely different. `ts-rs` has an issue to support `wasm-bindgen` but no movement so far:

- https://github.com/Aleph-Alpha/ts-rs/issues/313

It's too bad that the Rust community hasn't converged on a preferred package for generating TS type defs, but I don't see any easy way around this at the moment.

packages/automerge-doc-server/src/server.ts
jmoggr: I didn't think it needed to be updated, since we are just wrapping the migrateDocument function in `automerge-doc-server` to make it slightly easier to use in that context. The rest of the process for migrating a doc in TS (getting the migrated doc, creating a diff, applying a diff), remains the same.

However this is now messier than a simple cut and paste. Unfortunately I don't see an easy way of factoring out the common code without importing either `frontend` or `automerge-doc-server` into the other.

I'm inclined to let this slide, but would understand if you felt differently.

epatters: Thanks for explaining. I think it's fine to leave it.

PR #780, opened 2025-10-24
Run Biome CI for `automerge-doc-server`

### LINKED ISSUE #768: Run formatter/linter on `automerge-doc-server` in CI
This looks like an oversight. The fix should be as simple as adding a line

```json
"ci": "biome ci",
```

to the `package.json` in `automerge-doc-server`, like in `frontend` package. And then you'll likely have to fix some lints!

epatters:
Closes #768.

### DIFF
--- a/packages/automerge-doc-server/biome.json
+++ b/packages/automerge-doc-server/biome.json
@@ -1,0 +1,10 @@
+{
+    "extends": ["../../biome.json"],
+    "linter": {
+        "rules": {
+            "suspicious": {
+                "noExplicitAny": "off"
+            }
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


packages/automerge-doc-server/biome.json
```diff
@@ -0,0 +1,10 @@
+{
+    "extends": ["../../biome.json"],
+    "linter": {
+        "rules": {
+            "suspicious": {
+                "noExplicitAny": "off"
```
epatters, line 6: For better or worse, `any` is used quite deliberately in several places in this package, so it seemed simpler to disable this lint which is enabled in the root Biome config.

PR #781, opened 2025-10-24
Add missing `wasm-pack` dev dependency to `notebook-types`

epatters:
I was trying to compile the `automerge-doc-server` package locally and the first problem I encountered was that I couldn't even build `notebook-types` for Node.

### DIFF
--- a/packages/notebook-types/package.json
+++ b/packages/notebook-types/package.json
@@ -6,5 +6,8 @@
     "type": "module",
     "scripts": {
         "build:node": "wasm-pack build ./. --target nodejs -d ./dist/pkg-node"
+    },
+    "devDependencies": {
+        "wasm-pack": "^0.13.1"
     }
 }
--- a/packages/notebook-types/pnpm-lock.yaml
+++ b/packages/notebook-types/pnpm-lock.yaml
@@ -1,0 +1,206 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+
+  .:
+    devDependencies:
+      wasm-pack:
+        specifier: ^0.13.1
+        version: 0.13.1
+
+packages:
+
+  axios@0.26.1:
+    resolution: {integrity: sha512-fPwcX4EvnSHuInCMItEhAGnaSEXRBjtzh9fOtsE6E1G6p7vl7edEeZe11QHf18+6+9gR5PbKV/sGKNaD8YaMeA==}
+
+  balanced-match@1.0.2:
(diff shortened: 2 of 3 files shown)


kasbah:
I assumed because it's a Rust dependency we are handling it differently. `wasm-pack` is also [used for catlog-wasm](https://github.com/ToposInstitute/CatColab/blob/1b6dd6457c46c188c3842c1df07d169466ef8f29/packages/catlog-wasm/package.json#L8) for instance.

1. If we want to handle installation via pnpm it should probably go in the (root?) package.json?
2. If we want to handle installation via Cargo then it should go in the (root?) Cargo.toml
3. If we want to handle installation via Nix then... I believe we just need to document that fact

jmoggr:
I'm able to repro this issue locally by removing `wasm-pack` from the Nix dev shell and running `npm run build:node` in `automerge-doc-server`.

Evan: are you using the Nix dev shell? How are you trying to build `automerge-doc-server`?

If you were to run `npm run build:node` in `automerge-doc-server` without the Nix development shell being active, then I would expect this issue. [`wasm-pack` is provided by the Nix dev shell](https://github.com/ToposInstitute/CatColab/blob/1b6dd6457c46c188c3842c1df07d169466ef8f29/flake.nix#L114) and as a build dependency to the Nix package. The nix shell preventing the issue along with  `notebook-types` not being built frequently could explain why this problem has gone unnoticed.

As an artifact of my development process, I've had a bias towards including "devDependencies" type dependencies as nix packages. I don't have a good reason for doing this anymore, and should probably avoid it in the future. My not so good reasons: I don't have to worry about `PATH` issues, I don't have to update the pnpm dependencies hash.

epatters:
Thanks both for your comments.

> Evan: are you using the Nix dev shell? How are you trying to build `automerge-doc-server`?

No, I don't use the Nix dev shell for local development.

> If you were to run `npm run build:node` in `automerge-doc-server` without the Nix development shell being active, then I would expect this issue. [`wasm-pack` is provided by the Nix dev shell](https://github.com/ToposInstitute/CatColab/blob/1b6dd6457c46c188c3842c1df07d169466ef8f29/flake.nix#L114) and as a build dependency to the Nix package.

Ok, that explains why it works in Nix. I figured something like that must be going on because after all we can build the packages in CI.

> As an artifact of my development process, I've had a bias towards including "devDependencies" type dependencies as nix packages. I don't have a good reason for doing this anymore, and should probably avoid it in the future.

Yeah, we could conceivably make the argument that all devs should go through the Nix shell, and it might even be a good argument, but I'm not enough bought into Nix to want to make that argument. Just as a matter of personal preference, I'd like to preserve the possibility of using old-fashioned dev workflows. With that in mind, I think we should prefer the standard packaging mechanisms provided by Rust and NPM when possible and aim to keep the Nix infrastructure as a lightweight layer on top of those mechanisms.

So, I'll merge this PR, and I'll likely follow up with other issues I encountered when trying to build the backend components after a hiatus in doing so.

PR #795, opened 2025-10-31
FIX: Use document type instead of hardcoded 'model' in menu text

### LINKED ISSUE #791: Sidebar menu item for analysis says "Duplicate model"
It should say "Duplicate analysis" instead. See:

Image

Follow up to #749.

jmoggr:
closes #791


epatters (approved):
Thanks for the fix!

PR #798, opened 2025-10-31
Click an already open doc in the sidebar to maximize that doc

jmoggr:
PR #797 implicitly adds some new behavior: if two documents are open in panels and you click on the primary document, then the primary document becomes maximized. However when the secondary document is clicked in the sidebar, nothing happens.

This change unifies the behavior, clicking on the primary or secondary document in the sidebar will always maximize that document.

### DIFF
--- a/packages/frontend/src/page/document_page.tsx
+++ b/packages/frontend/src/page/document_page.tsx
@@ -103,7 +103,12 @@ export default function DocumentPage() {
                             closeSidePanel={closeSidePanel}
                         />
                     }
-                    sidebarContents={<DocumentSidebar liveDoc={liveDocument().liveDoc} />}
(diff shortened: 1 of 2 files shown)


kasbah:
It feels much better but still a little inconsistent in how it's going to open. E.g. when I go back to the model and back to analysis:


https://github.com/user-attachments/assets/2ab9d69e-545e-48c9-86e4-8af2201e02da

jmoggr:
That was a bug, good catch. It should be fixed now. The SolidJS reactivity tripped me up. I don't think I fully understand why this worked as well as it did before wrapping the `refId` accessors with `createMemo`.

epatters (approved):

PR #801, opened 2025-10-31
Prevent notebook cells shifting when focussing a rich text editor cell

jmoggr:
Previously on focus a notebook cell containing a rich text editor would resize to display the menubar and border, which would cause many UI elements to jump slightly.

This change prevents the resizing, but causes the menubar to possibly overlap other cells. To help ameliorate the possible problem of a menubar obscuring an element that a user wants to interact with, a button to manually hide the menubar (and defocus the RTE cell) was added.

I think the annoyance of sometimes having the menubar overlap things is less than the annoyance of the UI always jumping slightly.

### DIFF
--- a/packages/frontend/src/components/rich_text_editor/link_editor.tsx
+++ b/packages/frontend/src/components/rich_text_editor/link_editor.tsx
@@ -116,7 +116,7 @@ function getLinkCoords(
 
     const editorRect = view.dom.getBoundingClientRect();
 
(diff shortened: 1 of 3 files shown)


jmoggr:
I made both changes!

tim-at-topos:
I like this behaviour a lot!

epatters (approved):
I agree, this is looking really nice now.

PR #808, opened 2025-11-03
Split out a `ui-components` package, showcase and test it with Storybook

kasbah:
This makes a new package: `catcolab-ui-components` where our components can live and where we use [Storybook](https://storybook.js.org/) as a dev/test server. It includes some automatic testing that just checks that the "stories" run without errors and highlights possible accessibility issues (these will not lead to test failures yet). Additionally it acts as a showcase of components and is also deployed to `next.catcolab.org/dev/ui-components`. 

Mostly I just moved components over except for `Button` and `IconButton` where I consolidated some existing usage into components. I stopped short of moving components that depend on catlog-wasm types and the rich-text-editor though I actually do think we should move these too.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,6 +1,12 @@
 {
     "files": {
-        "ignore": ["dist/**", "docs/**", "catlog-wasm/**", "**/queries.ts"]
+        "ignore": [
+            "dist/**",
+            "docs/**",
+            "catlog-wasm/**",
+            "**/queries.ts",
+            "storybook-static/**"
+        ]
     },
     "$schema": "https://biomejs.dev/schemas/1.8.3/schema.json",
     "organizeImports": {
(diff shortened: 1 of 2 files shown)


kasbah:
It's a little hard to be sure I haven't affected any styles of any components. One I'm not sure how to even bring up in actual use is the `FixedTableEditor`. I got as far as this. (No Julia in next.catcolab.org either?)

image

epatters:
> It's a little hard to be sure I haven't affected any styles of any components.

That's understandable. I imagine we'll shake out little issues over time.

> One I'm not sure how to even bring up in actual use is the `FixedTableEditor`. I got as far as this.

The Decapodes analysis is difficult to run since it requires an external Julia setup, but most of the other quantitative analyses use the `FixedTableEditor`. For example, try creating a causal loop diagram with a few variables and arrows and then create a linear or Lotka-Volterra ODE analysis.

BTW, as your screenshot shows, this component is currently kind of ugly :)

epatters (approved):
Thanks Kaspar, this is excellent. Also happy to see the scattered styling for buttons get consolidated.

PR #813, opened 2025-11-04
FIX: Move doc components back to frontend to fix regressions

kasbah:
There were some *severe* regressions with the document editing section on my ui-components refactor (#808). I really need to be more careful! 

Since the error messages were not very helpful I haven't figured out exact causes yet. This restores previous components associated with the document pages and seems to fix the regressions. 

@epatters had reported:

- When I create a new model, the title field is small/doesn't grow with input
  - seems to be fixed by this
- notebook drag handles aren't appearing
  - I couldn't recreate this one on next because of much more severe errors. I think it's all fixed by this though.
- when I create a new place in a Petri net, I get an error in the console
  - I believe it is fixed. Please double check. 

Fixes https://github.com/ToposInstitute/CatColab/issues/814


kasbah:
Yes, it's quite confusing why moving these files around would cause these issues. I am putting off digging deeper because the main error message was so unhelpful and I want to get some things into the release right now. 

I do want to fix this properly after the release.

PR #815, opened 2025-11-04
Delete document from sidebar menu

### LINKED ISSUE #805: Delete documents from sidebar menu
The sidebar context menu should include a menu item to delete the document. This is much more convenient than finding the document in the document pages, especially if it has no title.

kasbah:
Will fix #805 once it works. 

A bit stumped by this bug though. If I delete from the sidebar mouse interaction with elements stops working. If delete from My Documents it's fine but in the sidebar it freezes. The freezing even happens if I make delete a no-op!

Recording (with sound):

https://github.com/user-attachments/assets/bd51dff2-8498-4ebd-97c3-86a47f09aa08

- I tried making delete not actually happen, or anything to happen really. Just opening the modal here seems to cause this.
- I tried untangling corvu from ui-components f8eab5502322e397bfdcf524f2d7d402ec715b6d but it didn't help
- I tried clearing out ui-components dependency from this completely
- I tried using `@corvu/dialog` directly instead of our wrapper

### DIFF
--- a/packages/frontend/src/page/document_menu.tsx
+++ b/packages/frontend/src/page/document_menu.tsx
@@ -1,7 +1,6 @@
 import { DropdownMenu } from "@kobalte/core/dropdown-menu";
 import { useNavigate } from "@solidjs/router";
 import Ellipsis from "lucide-solid/icons/ellipsis";
-import X from "lucide-solid/icons/x";
 import { Match, Switch, createMemo, createResource, createSignal } from "solid-js";
 import { useContext } from "solid-js";
 import { Show } from "solid-js";
@@ -10,10 +9,10 @@ import invariant from "tiny-invariant";
 import { IconButton } from "catcolab-ui-components";
 import { createAnalysis } from "../analysis";
 import { type LiveDoc, useApi } from "../api";
-import { useDeleteDocument } from "../components/delete_document_dialog";
 import { createDiagram } from "../diagram";
 import {
     CopyJSONMenuItem,
+    DeleteMenuItem,
     DuplicateMenuItem,
     ExportJSONMenuItem,
     MenuItem,
@@ -31,13 +30,6 @@ export function DocumentMenu(props: {
 
     const navigate = useNavigate();
     const docType = () => props.liveDoc.doc.type;
-
-    invariant(props.liveDoc.docRef, "No document reference found");
-    const deleteDocument = useDeleteDocument({
-        refId: props.liveDoc.docRef.refId,
-        name: props.liveDoc.doc.name,
-        typeName: props.liveDoc.doc.type,
-    });
 
     const onNewDiagram = async () => {
         let modelRefId: string | undefined;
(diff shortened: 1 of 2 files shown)


kasbah:
Ok, so 1bec941 was the fix / workaround. I believe regressions seen in #813 were also related to interaction between kobalte and corvu but I haven't confirmed.

packages/frontend/src/page/document_menu.tsx
```diff
@@ -80,41 +89,68 @@ export function DocumentMenu(props: {
...
-                </DropdownMenu.Content>
-            </DropdownMenu.Portal>
-        </DropdownMenu>
+                        <MenuItem
+                            disabled={!canDelete()}
+                            onSelect={async () => {
+                                // Explicitly closing the menu avoids some
+                                // strange conflict between kobalte and corvu.
+                                // Our UI locks if we don't close the menu
+                                // _first_.
```
epatters: Ah, I've seen something like this before but I can't recall the exact circumstances. It's unfortunate.

packages/frontend/src/page/document_menu.tsx
```diff
@@ -80,41 +89,68 @@ export function DocumentMenu(props: {
...
+                        <CopyJSONMenuItem doc={props.liveDoc.doc} />
                         <MenuSeparator />
-                    </Show>
-                    <DuplicateMenuItem doc={props.liveDoc.doc} />
-                    <ExportJSONMenuItem doc={props.liveDoc.doc} />
-                    <CopyJSONMenuItem doc={props.liveDoc.doc} />
-                </DropdownMenu.Content>
-            </DropdownMenu.Portal>
-        </DropdownMenu>
+                        <MenuItem
```
epatters: I'd suggest encapsulating this item as a self-contained component `DeleteMenuItem` in `menubar.tsx`.

At the moment, it'll just be for consistency with other menu items, but the more important reason is that I'd prefer to have separate menus for each document type (as I implemented in #747). Somehow this got undone in #749 but I intend to re-separate the menus because (1) I find all the conditionals confusing and (2) we're going to add more document types and this is not a sustainable strategy to manage the menus for all of them.

packages/frontend/src/components/delete_document_dialog.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+function noop(_: boolean): void {}
+
+export function useDeleteDocument(docInfo: DeleteDocumentInfo) {
+    const api = useApi();
+
+    const [showDeleteConfirm, setShowDeleteConfirm] = createSignal(false);
+    const [showError, setShowError] = createSignal(false);
+    const [errorMessage, setErrorMessage] = createSignal("");
+    const [resolveDeletion, setResolveDeletion] = createSignal<(success: boolean) => void>(noop);
```
epatters: Do we need `noop` defined above? Seems more idiomatic to have

```typescript
    const [resolveDeletion, setResolveDeletion] = createSignal<(success: boolean) => void | undefined>();
```

and then call

```typescript
resolveDeletion()?.(...);
```

In fact, you're already doing the latter in some but not all places below.

epatters (commented):
Thanks Kaspar! A few small questions/comments below.

packages/frontend/src/components/delete_document_dialog.tsx
kasbah: Right, I had the `?.()` way before but then thought the `noop` would be nicer. I don't feel strongly about it. I will change it to use `?.()`.

packages/frontend/src/page/document_menu.tsx
kasbah: That's definitely cleaner. I've refactored it into a DeleteMenuItem.

packages/frontend/src/page/menubar.tsx
```diff
@@ -229,3 +231,41 @@ function DocumentsMenuItem() {
...
+
+/** Menu item to delete a document. */
+export function DeleteMenuItem(props: {
+    refId: string | undefined;
+    name: string | null;
+    typeName: string;
+    canDelete: boolean;
+    onBeforeDelete?: () => void;
+}) {
+    invariant(props.refId, "No document reference found");
```
kasbah, line 243: I think we should clean up `LiveDoc` into two separate types. One with the `refDoc` (so we *know* we have a `refId` here) and one without for usage directly with automerge.

epatters: Good idea. I added the optional `refDoc` field as a quick solution during a refactor to allow Automerge-only usage of certain components (in our collaboration with Ink & Switch, see the `patchwork` package), but I haven't thought through a principled way to partially decouple the frontend from the backend. Currently some components will work without the `Api` object and others won't, and it's far from obvious for which components that is, or *should* be, the case.

PR #816, opened 2025-11-04
Run all frontend tests in CI with NixOS VM

jmoggr:
This creates: 
- a new Nix package which is responsible for running the `vitest` tests from the `frontend` package.
- a new Nix check, which is a full catcolab host running in a VM that we run the frontend tests against.

This result is end-to-end tests against a host that was setup from scratch.

I needed to bump the `vitest` version and some follow on dependencies to let `vitest` run in a read-only environment.

### DIFF
--- a/packages/frontend/default.nix
+++ b/packages/frontend/default.nix
@@ -61,8 +61,7 @@ let
         ];
       };
       # See README.md
-      # hash = "";
-      hash = "sha256-krMxJvNarhwZK8O08RonIwpzaTlmjT2ToWINe+tm9n8=";
+      hash = "sha256-TQzF+hhRABG7xKlTIThRsYvGHRnvxZs6XAJFbWOyBYk=";
     };
   };
 


infrastructure/modules/catcolab/services.nix
```diff
@@ -135,17 +143,31 @@ with lib;
...
+    # moving migrations out of backend (which causes many more problems), or have the backend process control
+    # the automerge process instead of systemd.
     systemd.services.automerge = {
       enable = true;
       wantedBy = [ "multi-user.target" ];
 
+      after = [
+        "database-setup.service"
+        "network-online.target"
+      ];
```
jmoggr, line 161: This missing block is what was causing the automerge service to not work correctly sometimes after startup. I'm not sure why we have never run into this issue before.

epatters: Good find. I guess we were just getting lucky with the timing of the services starting up.

packages/frontend/.env.production
epatters, line 1: Did you intend to delete this file?

packages/frontend/package.json
```diff
@@ -52,6 +52,7 @@
         "fast-json-patch": "^3.1.1",
         "firebase": "^10.14.0",
         "js-file-download": "^0.4.12",
+        "jsdom": "^27.0.1",
```
epatters: We should not have to add this dependency. I took care to structure the frontend package and frontend-only tests so that no simulated DOM is needed. We might need further tweaks to ensure that the integration tests do not need it either. (In principle it should not since no components are used in the tests.)

.github/workflows/ci.yml
```diff
@@ -131,37 +100,6 @@ jobs:
...
-            cd packages/notebook-types
-            pnpm run build:node
-            cd ../automerge-doc-server
-            pnpm install
-            cd ../backend
-            cp .env.development .env && cp .env.development ../migrator/.env
-            # due to how we are detecting deploy vs dev we need to clear this env variable
-            unset INVOCATION_ID
-            cargo run -p migrator apply
-
```
kasbah, line 164: Did you intend to delete this? This is still testing that what our docs say in `backend/README.md` even after #866 .

packages/frontend/package.json
jmoggr: I removed it.

Looking back at my commits: I fought with this for a while without realizing that I accidentally added a DOM dependency for some, now removed, debugging code.

.github/workflows/ci.yml
jmoggr: Yes.

PR with updated docs for database setup is here: https://github.com/ToposInstitute/CatColab/pull/871

While this does remove the explicit check for developer database setup, basically the exact same steps are being used between setting up the VM database as the steps for setting up a development database, so I think the development database setup is sufficiently covered without the need for a dedicated test.

packages/frontend/.env.production
jmoggr: Yes, but I meant to comment on it.

1. No documentation or code references this file and and it's not used as part of our automated tests.
2. It's ot clear when it would be useful, if changes are being made to the backend then testing against prod is only relevant after we've deployed to prod.
3. I'm somewhat against running tests against prod which create an update state, it could add a lot of noise to a lot of already noisy data.

If we want to keep it as a smoke test after deploys to prod, then we can keep it and add an automated check as part of deploy-rs (I already have code for this).

There is also nothing stopping us from just keeping it if you'd like. I just felt there were a number of small reasons to get rid of it.

epatters: I'm confused: isn't this file used by vite when building the frontend for production? Or, if it's not being used, where is vite getting that information?

jmoggr: :facepalm:

> where is vite getting that information?

From the `.env.production` that I deleted... It is only used in CI workflows. I've restored it.

There were no search results for `.env.production` in the repo. I've added comments with .env filenames to the deploy workflow so the usage of .env files can be found through search.

jmoggr:
I pushed a fix.

Steps: 
- Comment out `hash` in `frontend/default.nix`
- run `nix build .#frontend`, it'll spit out an error like:
```
error: hash mismatch in fixed-output derivation '/nix/store/x9c0r42gkz8kb916l636mvpbci2wf9bs-frontend-pnpm-deps.drv':
         specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
            got:    sha256-FqjdtE/OgmV+aYGh1AijQqvfg9/UQC/dClJRTLyGxCE=
```
- replace the old hash with the `got` hash.

I'll make sure to improve the docs on this

>  I tried reverting the changes to the pnpm lock file

It's possible that there was another instance of something changing and the hash not getting updated. Fortunately this PR should prevent that going forward.

PR #822, opened 2025-11-04
FIX: Ensure `frontend` and `ui-components` are using same Kobalte version

kasbah:


kasbah:
I was hoping this would fix our UI issues from the component refactor but it seems to have had no effect on that.

PR #823, opened 2025-11-04
Add "trash" page and document restore functionality

kasbah:
image

image

image

### DIFF
--- a/packages/frontend/src/page/document_page.tsx
+++ b/packages/frontend/src/page/document_page.tsx
@@ -53,12 +53,12 @@ export default function DocumentPage() {
         }
     });
 
-    const [primaryLiveDocument] = createResource(
+    const [primaryLiveDocument, { refetch: refetchPrimaryDocument }] = createResource(
         () => params.ref,
         (refId) => getLiveDocument(refId, api, models, params.kind as DocumentType),
     );
 
-    const [secondaryLiveDocument] = createResource(
+    const [secondaryLiveDocument, { refetch: refetchSecondaryDocument }] = createResource(
         () => {
             if (!params.subkind || !params.subref) {
                 return;
@@ -124,7 +124,10 @@ export default function DocumentPage() {
                                         initialSize={1}
                                         minSize={0.25}
                                     >
-                                        <DocumentPane document={liveDocument()} />
+                                        <DocumentPane
+                                            document={liveDocument()}
+                                            refetch={refetchPrimaryDocument}
+                                        />
                                     </Resizable.Panel>
                                     <Show when={isSidePanelOpen()}>
                                         <ResizableHandle class="resizeable-handle" />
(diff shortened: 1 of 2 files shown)


packages/backend/src/document.rs
```diff
@@ -335,7 +352,9 @@ pub async fn search_ref_stubs(
                             AND p_searcher.subject = $1
                     )
                 ) AND (
-                    refs.deleted_at IS NULL
+                    -- optionally filter for only deleted refs
+                    ($8::bool IS TRUE AND refs.deleted_at IS NOT NULL)
+                    OR ($8::bool IS NOT TRUE AND refs.deleted_at IS NULL)
```
epatters, line 357: No action to take here, but the fact that this query is now on its eighth numbered query parameter sure makes me wish that `sqlx` supported named query parameters: https://github.com/launchbadge/sqlx/issues/875

packages/frontend/src/page/document_page.tsx
```diff
@@ -186,16 +187,44 @@ function SplitPaneToolbar(props: {
...
+
+    const handleRestore = async () => {
+        const refId = props.document.liveDoc.docRef?.refId;
+        if (!refId) return;
+
+        try {
+            const result = await api.rpc.restore_ref.mutate(refId);
+            if (result.tag === "Ok") {
+                // Refresh the page to show the restored document
+                window.location.reload();
```
epatters: It feels like a hack to manually trigger a page reload in a single-page app. It's more work, but IMO more idiomatic, to grab the refetch callbacks when the resources are created

```typescript
const [primaryLiveDocument, {refetch: refetchPrimaryDocument}] = createResource(...);

const [secondaryLiveDocument, {refetch: refetchSecondaryDocument}] = createResource(...);
```

and then pass the refetchers into the `DocumentPane`s as an optional prop.

There might be other patterns that could be used here, but that's the first thing that comes to mind.

packages/frontend/src/user/trash.tsx
epatters, line 1: This file seems to be copied from `user/documents.tsx` with just a few scattered modifications.

No action needed in this PR, but in future work we should build a more featureful and better looking document browser component that accommodates both of these uses.

epatters (changes requested):
Thanks Kaspar, this is a great complement to the recently added deletion feature. A few comments below, only one of which is immediately actionable.

packages/frontend/src/page/document_page.tsx
kasbah: I tried refetching with https://github.com/ToposInstitute/CatColab/commit/4b60af81eef543bcfb14e49af30730f52cb877da but the docRef cache seems to be the main thing so https://github.com/ToposInstitute/CatColab/pull/823/commits/350d9567d7974e2357b6272e88c2d3d9b4d325a3 works just as well.

kasbah: Never mind, I just noticed if I don't refetch the styling for "not deleted" in the sidebar doesn't get applied. I applied the refetch commit.

epatters (approved):

PR #824, opened 2025-11-05
Replace welcome overlay with a home page

kasbah:


packages/frontend/src/model/model_editor.tsx
```diff
@@ -34,21 +31,8 @@ export function ModelNotebookEditor(props: {
...
-    const auth = firebaseApp && useAuth(getAuth(firebaseApp));
-
-    const [isOverlayOpen, setOverlayOpen] = createSignal(
-        liveDoc().doc.notebook.cellOrder.length === 0 && auth != null && auth.data == null,
-    );
-    const toggleOverlay = () => setOverlayOpen(!isOverlayOpen());
-
     return (
         <LiveModelContext.Provider value={() => props.liveModel}>
-            <WelcomeOverlay isOpen={isOverlayOpen()} onClose={toggleOverlay} />
```
epatters, line 51: Pleased to see this removed from the model editor. Having an app-level widget embedded so deep in the component hierarchy was always a hack.

packages/frontend/src/page/home_page.tsx
```diff
@@ -0,0 +1,188 @@
...
+    const api = useApi();
+
+    const firebaseApp = (() => {
+        try {
+            return useFirebaseApp();
+        } catch {}
+    })();
+    const auth = firebaseApp && useAuth(getAuth(firebaseApp));
+    const isLoggedIn = () => auth?.data != null;
+    const isAuthLoading = () => auth?.loading === true;
```
epatters: Now that we've extracted this functionality from the model editor, we should be able to safely assume that the app-level context is available. So we can simplify all this to:

```typescript
    const firebaseApp = useFirebaseApp();
    const auth = useAuth(getAuth(firebaseApp));
    const isLoggedIn = () => auth.data != null;
    const isAuthLoading = () => auth.loading;
```

packages/frontend/src/page/home_page.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+    const handleCreateModel = async () => {
+        setCreating(true);
+        try {
+            const ref = await createModel(api, stdTheories.defaultTheoryMetadata().id);
+            navigate(`/model/${ref}`);
+        } catch (error) {
+            console.error("Failed to create model:", error);
+            setCreating(false);
+        }
```
epatters, line 48: Seems safer/more idiomatic to reset the creating status regardless of whether an error occurs:

```typescript
catch (error) {
    console.error("Failed to create model:", error);
} finally {
    setCreating(false);
}
```

packages/frontend/src/page/home_page.tsx
```diff
@@ -0,0 +1,188 @@
...
+                                        disabled={creating()}
+                                    >
+                                        <FilePlus />
+                                        <span>{creating() ? "Creating..." : "New model"}</span>
+                                    </button>
+                                </div>
+                                <div class="resources-container">
+                                    <div class="resources-list">
+                                        <a
+                                            href="https://topos.institute/blog/#category=CatColab"
```
epatters, line 104: @tim-at-topos, this link does not have the desired effect for me (Firefox on Linux). In fact, when I go to the Topos blog (<https://topos.institute/blog/>) and click any of the categories on the right side, nothing happens. Perhaps there has been a regression on the blog?

packages/frontend/src/page/home_page.tsx
```diff
@@ -0,0 +1,188 @@
...
+                                            href="/help/guides/example-models"
+                                            class="resource-link"
+                                            target="_blank"
+                                        >
+                                            <span class="resource-icon">
+                                                <Binoculars />
+                                            </span>
+                                            <span>Example models</span>
+                                            <span class="external-link-icon">
+                                                <ExternalLink />
```
epatters: It's a small matter, but this link (and also the one immediately below to `/help`) isn't actually external, so IMO we shouldn't treat it as such by opening it in a new window. However, I'm open to the argument that we should do so for consistency with the other links.

packages/frontend/src/page/home_page.tsx
kasbah, line 48:
```suggestion
        } catch (error) {
            console.error("Failed to create model:", error);
        } finally {
            setCreating(false);
        }
```

kasbah:
Thanks @epatters, all feedback (except blog link) should be resolved.

packages/frontend/src/page/home_page.tsx
epatters: Issue filed upstream.

tim-at-topos:
I'm sure I probably gave a contradictory opinion in the past, but I would suggest swapping "blog post" with "help pages" — our help pages should be good enough that they are the best place to read about anything, and e.g. in particular include links to any specific blog posts

epatters:
Good point, I agree. Feel free to open a new PR making that change.

PR #825, opened 2025-11-06
Fix delete in sidebar

kasbah:
The saga of weird interactions around portals / dialogs continues. Delete in sidebar is currently broken on next. 

Reverting the refactor of moving the menu item fixes it. I could have _sworn_ I tested the refactor but it may have been another case of "this isn't really changing anything" that made me complacent. We really need to watch out for these and also need to fix the root cause.


kasbah:
The way this is popping up makes me think the root cause is a race condition. Reverting changes as I have been doing may be subtly changing the timing and just making it seem like it is fixed.

kasbah:
Superseded by #826

PR #826, opened 2025-11-06
Remove kobalte dependency from frontend

kasbah:
Replaces kobalte menu with corvu popover. This seems to fix the issues I have been encountering with portal / dialog / popovers including fixing delete from sidebar on next.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -42,7 +42,8 @@
                 },
                 "noUselessElse": {
                     "level": "off"
-                }
+                },
+                "useBlockStatements": "error"
             }
         }
     },
--- a/packages/frontend/src/components/delete_document_dialog.tsx
+++ b/packages/frontend/src/components/delete_document_dialog.tsx
@@ -31,7 +31,9 @@ export function useDeleteDocument() {
     const handleConfirmDelete = async () => {
         setShowDeleteConfirm(false);
         const info = currentDocInfo();
-        if (!info) return;
+        if (!info) {
+            return;
+        }
 
         try {
             const result = await api.rpc.delete_ref.mutate(info.refId);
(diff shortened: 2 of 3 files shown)


packages/frontend/src/page/page_container.tsx
```diff
@@ -17,9 +18,14 @@ export function PageContainer(props: {
...
     const [openImport, setImportOpen] = createSignal(false);
 
+    const deleteDocument = useDeleteDocument();
+
     const actions: PageActions = {
         showLoginDialog: () => setLoginOpen(true),
         showImportDialog: () => setImportOpen(true),
+        showDeleteDialog: async (docInfo: DeleteDocumentInfo) => {
+            return await deleteDocument.openDeleteDialog(docInfo);
+        },
```
kasbah:
```suggestion
showDeleteDialog: deleteDocument.openDeleteDialog,
```

packages/frontend/src/page/page_container.tsx
```diff
@@ -17,9 +18,12 @@ export function PageContainer(props: {
     const [loginOpen, setLoginOpen] = createSignal(false);
     const [openImport, setImportOpen] = createSignal(false);
 
+    const deleteDocument = useDeleteDocument();
```
epatters, line 21: Encapsulating everything needed for the delete dialog in the function `useDeleteDocument` is cleaner than how we're handling the login and import dialogs in this component. In the future, we should follow your pattern in those cases too.

epatters (approved):
Thanks for persevering to find a workaround for these strange interactions.

I'll go ahead and merge this since it seems important to get this fixed, but I'm assuming that the new, custom menubar components should eventually go into `ui-components`?

PR #835, opened 2025-11-08
Redesign of Julia interop, avoiding Jupyter kernel

kris-brown:
This PR simplifies a lot of the structure of the CatColabInterop.jl package.

It introduces a new analysis of schema instances by sending the diagram to Catlab, where it is processed by `@acset_colim` to produce a tabular instance.

(This is currently a WIP - once the workflow is working for the tabular view analysis, the existing Decapodes analysis will be brought back into functionality).


packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,148 @@
...
+            const response = await axios.post(
+                "http://127.0.0.1:8080/acsetcolim",
+                {
+                    model: model?.presentation(),
+                    diagram: diagram?.presentation(),
+                },
+                config,
+            );
+            return response.data;
+        },
```
kasbah: You should be able to do this without `axios`, using native `fetch` instead. It's a bit more code but also more explicit about what's going on. Something like:

```ts
async ([model, diagram]) => {
    const response = await fetch(
        "http://127.0.0.1:8080/acsetcolim",
        {
            method: "POST",
            headers: {
                "Content-Type": "application/json",
                ...config.headers,
            },
            body: JSON.stringify({
                model: model?.presentation(),
                diagram: diagram?.presentation(),
            })
        },
    );
    
    if (!response.ok) {
        throw new Error(`HTTP error! status: ${response.status}`);
    }
    
    return response.json();
}
```

packages/algjulia-interop/scripts/endpoint.jl
```diff
@@ -0,0 +1,51 @@
...
+
+function CorsHandler(handle)
+    return function (req::HTTP.Request)
+        # return headers on OPTIONS request
+        if HTTP.method(req) == "OPTIONS"
+            return HTTP.Response(200, CORS_HEADERS)
+        else
+            r = handle(req)
+            append!(r.headers, ["Access-Control-Allow-Origin" => "*"])
+            r
```
kasbah, line 27: It's a small thing but I think it makes more sense to send the same headers for any request.

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,156 @@
...
+    url: "http://127.0.0.1:8080",
+    method: "post",
+    headers: {
+        "Access-Control-Allow-Origin": "*",
+        "Access-Control-Allow-Methods": "POST, PUT, DELETE, GET, OPTIONS",
+        "Access-Control-Allow-Headers": "*",
+        "Access-Control-Allow-Credentials": "false",
+    },
+};
+
```
kasbah: It doesn't make sense to set these CORS headers in client-side requests. You can just leave them off. Additionally the rest of the config is not used anymore since switching to fetch.

```suggestion
```

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,156 @@
...
+            const diagram = props.liveDiagram.elaboratedDiagram();
+            return model && diagram && [model, diagram];
+        },
+
+        async ([model, diagram]) => {
+            const response = await fetch("http://127.0.0.1:8080/acsetcolim", {
+                method: "POST",
+                headers: {
+                    "Content-Type": "application/json",
+                    ...config.headers,
```
kasbah:
```suggestion
```

kris-brown:
Thanks, these are addressed now!

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,139 @@
...
+                throw new Error(`HTTP error! status: ${response.status}`);
+            }
+
+            return response.json();
+        },
+    );
+    const model = props.liveDiagram.liveModel.elaboratedModel();
+    if (model === undefined) {
+        throw "Bad model";
+    }
```
epatters: You don't want to throw an error here and this code looks redundant anyway since you're already extracting the elaborated model in the Solid resource. I think you can just delete the above lines:

```typescript
    const model = props.liveDiagram.liveModel.elaboratedModel();
    if (model === undefined) {
        throw "Bad model";
    }
```

packages/frontend/src/stdlib/analyses.tsx
```diff
@@ -41,6 +41,16 @@ export const diagramGraph = (
 
 const DiagramGraph = lazy(() => import("./analyses/diagram_graph"));
 
+export const tabularView = (
+    options: AnalysisOptions,
+): DiagramAnalysisMeta<GraphLayoutConfig.Config> => ({
```
epatters: Using the graph layout config doesn't make sense. Presumably a holdover from copy-pate.

If I'm not mistaken, this analysis has no content/state of its own, so you can replace

```typescript
DiagramAnalysisMeta<GraphLayoutConfig.Config>
```

with

```typescript
DiagramAnalysisMeta<{}>
```

packages/frontend/src/stdlib/analyses.tsx
```diff
@@ -41,6 +41,16 @@ export const diagramGraph = (
...
+    ...options,
+    component: (props) => <TabularView title={options.name} {...props} />,
+    initialContent: GraphLayoutConfig.defaultConfig,
```
epatters: This then becomes

```typescript
initialContent: {},
```

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,139 @@
...
+                        </tr>
+                    )}
+                </For>
+            </tbody>
+        </table>
+    );
+}
+
+/** Given a schema (DblModel of ThSchema), a JSON output `rawdata` from Catlab,
+    and a particular object `obname` in the schema, create an HTML table with
```
epatters: Docstring seems to end mid-sentence.

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,139 @@
...
+) {
+    const [res] = createResource(
+        () => {
+            const model = props.liveDiagram.liveModel.elaboratedModel();
+            const diagram = props.liveDiagram.elaboratedDiagram();
+            return model && diagram && [model, diagram];
+        },
+
+        async ([model, diagram]) => {
+            const response = await fetch("http://127.0.0.1:8080/acsetcolim", {
```
epatters, line 96: Since this feature currently requires a server running locally, I think we should treat it as "experimental" and thus not show this analysis in prod. I'll suggest elsewhere how to do that.

packages/frontend/src/stdlib/theories/simple-schema.ts
```diff
@@ -118,6 +118,12 @@ export default function createSchemaTheory(theoryMeta: TheoryMeta): Theory {
                 description: "Visualize the instance as a graph",
                 help: "visualization",
             }),
+            analyses.tabularView({
+                id: "tabularview",
+                name: "Tabular Visualization",
+                description: "Visualize the instance as a table",
+                help: "tabularview",
+            }),
```
epatters, line 134: Per my other comment, we should add this analysis only when the variable `import.meta.env.PROD` is `true`.

Note: I haven't tested that, but I think it should work. See https://vite.dev/guide/env-and-mode#built-in-constants

packages/algjulia-interop/README.md
```diff
@@ -1,47 +1,37 @@
...
-   Julia kernel to Jupyter
-   
-At this stage, you should be able to launch a Julia kernel inside a JupyterLab.
-
-Having done that, navigate to this directory and run:
-
-```sh
-julia --project -e 'import Pkg; Pkg.instantiate()'
-```
+[AlgebraicJulia](https://www.algebraicjulia.org/) available to CatColab.At this
```
epatters: Typo: missing space after "CatColab."

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const rows: Array<string> = rawdata[obname as keyof typeof rawdata];
+
+    // Get the homs and attrs with source `obname`
+    const outhoms = model
+        .morGenerators()
+        .filter(
+            (m) =>
+                obname ===
+                model
+                    .obGeneratorLabel(model.morPresentation(m)?.dom.content.toString() || "")
```
epatters: As another rule of thumb, you should never have to call `toString()` when working with presentations. Calling it suggests that you haven't narrowed the type enough. I'd rewrite the above predicate as

```typescript
(id) => {
    const mor = model.morPresentation(id);
    return mor.dom.type === "Basic" && obname === model.obGeneratorLabel(mor.dom.content);
}
```

(But that's assuming that you're still keying by object labels, which I wouldn't recommend. See my other comment.)

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,139 @@
...
+                </For>
+            </tbody>
+        </table>
+    );
+}
+
+/** Given a schema (DblModel of ThSchema), a JSON output `rawdata` from Catlab,
+    and a particular object `obname` in the schema, create an HTML table with
+*/
+function createACSetTable(model: DblModel, rawdata: object, obname: string) {
```
epatters: It's a code smell that `rawdata` has type `object`. You should have a type def of the format of the data expected to be returned from CatColab.

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,139 @@
...
+    // The primary key of this table is given by `rawdata[obname]`
+    const rows: Array<string> = rawdata[obname as keyof typeof rawdata];
```
epatters: As a rule of thumb, when working with models, you should always key mappings by IDs rather than human-readable labels, which (1) are not guaranteed to be unique and (2) are structured objects (currently, lists of strings, rather than strings).

I would suggest that the last argument to this function be `obId: string` rather than `obname` and then

```typescript
const rows: Map<string, string> = new Map();
```

Note that `Map`s in JavaScript (unlike JSON objects) are guaranteed to preserve the insertion order, so you don't need to worry about losing ordering.

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,139 @@
...
+            (m) =>
+                obname ===
+                model
+                    .obGeneratorLabel(model.morPresentation(m)?.dom.content.toString() || "")
+                    ?.toString(),
+        );
+
+    // Convert morgenerators to user-friendly names
+    const headers = [obname].concat(
+        outhoms.map((m) => model.morGeneratorLabel(m)?.toString() || ""),
```
epatters: We should probably have a helper function for this, but the correct way to get a human-friendly name is

```typescript
model.morGeneratorLabel(m)?.join(".") ?? ""
```

kris-brown:
Thanks! I believe your suggested changes have all been incorporated. However, I'm not sure how to test whether the `import.meta.env.PROD` condition is working properly when running it locally.

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    return createTable(headers, data);
+}
+
+/** Stack tables on top of each other in a naive way, one per ob/attrtype */
+function createACSet(model: DblModel, rawdata: Record<string, string[]>) {
+    return (
+        <div class="simulation">
+            <PanelHeader title="Tabular view" />
+            <For each={model?.obGenerators()}>{(ob) => createACSetTable(model, rawdata, ob)}</For>
```
epatters: Looks like we're calling a function to create JSX elements instead of using a component, which is rarely what you want to do as it can cause reactivity problems. @kasbah, can you help Kris restructure this code into components?

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -0,0 +1,134 @@
...
+                <Spinner />
+                <div>⏳ Loading model...</div>
+            </Match>
+            <Match when={res.error}>
+                <div>❌ Error loading model: {res.error?.message || "Unknown error"}</div>
+            </Match>
+            <Match when={res()}>
+                {(data) => {
+                    const result = data();
+                    return <div>{createACSet(result.model, result.data)}</div>;
```
epatters: Ditto here.

epatters (changes requested):
Thanks Kris, this is looking really close to ready!

epatters (changes requested):
One other thing: this PR should remove `@jupyterlab/services` from `frontend`.

packages/frontend/src/stdlib/analyses/tabular_view.tsx
kris-brown: I suspect this is addressed, if I properly understood Kaspar's instructions!

epatters (approved):

PR #841, opened 2025-11-10
Move most components back to `ui-components`

kasbah:
This moves ui-components originally moved with #808. 

The `dedupe` in vite config (extending #834) fixes the error we were seeing before  around the document panes. I confirmed this be re-creating the error and then adding `@corvu/*` dependencies to the dedupe.

It does seem like it needs to be kept in sync with any deps (that use Solid?) that are used in both the packages. I added a script to do that and CI job to make sure that we have done that.

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -14,10 +14,12 @@
         "preview": "vite preview",
         "format": "biome format --write",
         "lint": "biome lint --write && biome check --write",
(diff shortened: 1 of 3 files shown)


epatters (approved):
Glad that there's a path forward here by dedupe-ing. Thanks for adding the consistency check to the CI.

PR #843, opened 2025-11-10
Generate SQL schema definition from CatColab schema

quffaro:


packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,110 @@
+// import type { JSX } from "solid-js";
+
+import { DropdownMenu } from "@kobalte/core/dropdown-menu";
```
kasbah: FYI this has been removed in the meantime so your merge with `main` broke things. Have a look here for replacing this with `@corvu/popover`

https://github.com/ToposInstitute/CatColab/pull/826/files#diff-a078dd4015617a7e02c970ddeceaca6e0798c5cbe1bd0306511532c517eb11e8

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,110 @@
...
+import download from "js-file-download";
+import CircleHelp from "lucide-solid/icons/circle-help";
+import Download from "lucide-solid/icons/download";
+import { createSignal, Show } from "solid-js";
+
+import { ThSchema } from "catlog-wasm";
+import type { ModelAnalysisProps } from "../../analysis";
+import { IconButton } from "../../components";
```
kasbah:
```suggestion
import { IconButton } from "catcolab-ui-components";
```

packages/frontend/src/model/morphism_input.tsx
```diff
@@ -21,6 +21,7 @@ export function BasicMorInput(
 
     const completions = (): QualifiedName[] | undefined =>
         props.morType && liveModel().elaboratedModel()?.morGeneratorsWithType(props.morType);
+    console.log(props.morType);
```
kasbah:
```suggestion
```

packages/frontend/src/model/object_input.tsx
```diff
@@ -107,6 +107,7 @@ function BasicObInput(allProps: ObInputProps & IdInputOptions) {
 
     const completions = (): QualifiedName[] | undefined =>
         liveModel().elaboratedModel()?.obGeneratorsWithType(props.obType);
+    console.log(props.obType);
```
kasbah:
```suggestion
```

kasbah:
I worked around the import issues mentioned above and tried to get a SQL DML from this schema:

image

And I get an error:

```error
TypeError: thSchema.renderSql is not a function
    sqlOutput sql.tsx:25
    get when sql.tsx:57
    conditionValue dev.js:1521
    runComputation dev.js:742
    updateComputation dev.js:724
    readSignal dev.js:660
    runComputation dev.js:742
    updateComputation dev.js:724
    createMemo dev.js:273
    Show dev.js:1524
error_boundary.tsx:8:13
```

If you want to paste the schema in ourself:

<details>
<summary>JSON</summary>
{"name":"New schema for CatColab backend","notebook":{"cellContents":{"019a91c6-a809-76bd-b4bf-449c98d39edc":{"content":{"id":"019a91c6-a809-76bd-b4bf-4224b473450a","name":"ref","obType":{"content":"Entity","tag":"Basic"},"tag":"object"},"id":"019a91c6-a809-76bd-b4bf-449c98d39edc","tag":"formal"},"019a91c6-dc8a-741c-8693-c3ec5946f2de":{"content":{"id":"019a91c6-dc8a-741c-8693-bfb3348ba824","name":"BINARY","obType":{"content":"AttrType","tag":"Basic"},"tag":"object"},"id":"019a91c6-dc8a-741c-8693-c3ec5946f2de","tag":"formal"},"019a91c6-fc18-751e-9ff9-1944556dfe07":{"content":{"cod":{"content":"019a91c6-dc8a-741c-8693-bfb3348ba824","tag":"Basic"},"dom":{"content":"019a91c6-a809-76bd-b4bf-4224b473450a","tag":"Basic"},"id":"019a91c6-fc18-751e-9ff9-16cd16a07454","morType":{"content":"Attr","tag":"Basic"},"name":"automerge-doc","tag":"morphism"},"id":"019a91c6-fc18-751e-9ff9-1944556dfe07","tag":"formal"},"019a91c7-9008-7475-9125-02d3d7fcb4d4":{"content":{"id":"019a91c7-9008-7475-9124-ff9107827cd5","name":"snapshot","obType":{"content":"Entity","tag":"Basic"},"tag":"object"},"id":"019a91c7-9008-7475-9125-02d3d7fcb4d4","tag":"formal"},"019a91c7-b057-769e-bc00-f4a2059fd276":{"content":{"cod":{"content":"019a91c6-a809-76bd-b4bf-4224b473450a","tag":"Basic"},"dom":{"content":"019a91c7-9008-7475-9124-ff9107827cd5","tag":"Basic"},"id":"019a91c7-b057-769e-bc00-f0d81e6b9bb8","morType":{"content":{"content":"Entity","tag":"Basic"},"tag":"Hom"},"name":"of","tag":"morphism"},"id":"019a91c7-b057-769e-bc00-f4a2059fd276","tag":"formal"},"019a91c8-0f48-7081-b0b6-867e2a34965e":{"content":{"id":"019a91c8-0f48-7081-b0b6-83f68fd9f516","name":"TEXT","obType":{"content":"AttrType","tag":"Basic"},"tag":"object"},"id":"019a91c8-0f48-7081-b0b6-867e2a34965e","tag":"formal"},"019a91c8-3fc4-7099-bf34-588502707401":{"content":{"cod":{"content":"019a91c8-0f48-7081-b0b6-83f68fd9f516","tag":"Basic"},"dom":{"content":"019a91c7-9008-7475-9124-ff9107827cd5","tag":"Basic"},"id":"019a91c8-3fc4-7099-bf34-559c42004e58","morType":{"content":"Attr","tag":"Basic"},"name":"automerge-url","tag":"morphism"},"id":"019a91c8-3fc4-7099-bf34-588502707401","tag":"formal"},"019a91c8-9990-7299-add5-663655683eaa":{"content":{"cod":{"content":"019a91c8-0f48-7081-b0b6-83f68fd9f516","tag":"Basic"},"dom":{"content":"019a91c7-9008-7475-9124-ff9107827cd5","tag":"Basic"},"id":"019a91c8-9990-7299-add5-6346f22be765","morType":{"content":"Attr","tag":"Basic"},"name":"tag","tag":"morphism"},"id":"019a91c8-9990-7299-add5-663655683eaa","tag":"formal"},"019a91cc-424b-76ec-95aa-6470907e8fb9":{"content":"￼Each snapshot corresponds to heads for the Automerge doc in its ref. Heads are stored as a full Automerge URL constructed by calling stringifyAutomergeUrl. ￼Don't need to store timestamp since retrievable from Automerge.","id":"019a91cc-424b-76ec-95aa-6470907e8fb9","tag":"rich-text"},"019a91d2-fd80-7526-89c1-18b1b69879c9":{"content":"Document refs store an Automerge doc and also have permissions (not shown).","id":"019a91d2-fd80-7526-89c1-18b1b69879c9","tag":"rich-text"}},"cellOrder":["019a91c6-dc8a-741c-8693-c3ec5946f2de","019a91c8-0f48-7081-b0b6-867e2a34965e","019a91d2-fd80-7526-89c1-18b1b69879c9","019a91c6-a809-76bd-b4bf-449c98d39edc","019a91c6-fc18-751e-9ff9-1944556dfe07","019a91cc-424b-76ec-95aa-6470907e8fb9","019a91c7-9008-7475-9125-02d3d7fcb4d4","019a91c7-b057-769e-bc00-f4a2059fd276","019a91c8-9990-7299-add5-663655683eaa","019a91c8-3fc4-7099-bf34-588502707401"]},"theory":"simple-schema","type":"model","version":"1"}

</details>

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,108 @@
+// import type { JSX } from "solid-js";
+
```
kasbah:
```suggestion
```

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,108 @@
...
+    // TODO SQLite can be an invalid change
+    const [backend, setBackend] = createSignal("MySQL");
+    const sqlOutput = () => {
+        const model = props.liveModel.elaboratedModel();
+        return model ? thSchema.renderSql(model, backend()) : null;
+    };
+    const downloadText = (text: string) => {
+        downloadTextContent(
+            text,
+            // props.filename ??
```
kasbah:
```suggestion
```

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    return (
+        <div>
+            <Show when={sqlOutput()}>
+                {(sql) => (
+                    <div>
+                        <Foldable title={title()} header={header(sql())}>
+                            {BackendConfig(backends)}
+                        </Foldable>
+                        <pre style="white-space: pre-wrap;">{sql()}</pre>
```
kasbah:
```suggestion
                        <pre style="white-space: pre-wrap; font-family:var(--mono-font)">{sql()}</pre>
```
Changes the font to the one we selected in global.css. This is currently `Roboto Mono` but will change to `Source Code Pro` with #865.

kasbah: Maybe it would be more correct to set the default font-family for `pre` tags in global.css

packages/catlog-wasm/src/model.rs
```diff
@@ -344,6 +344,16 @@ impl DblModel {
...
 
+    /// Tries to return the object namespace
+    pub fn ob_namespace(&self) -> Result<Namespace, String> {
+        Ok(self.ob_namespace.clone())
+    }
+
+    /// Tries to returns the morphism namespace
+    pub fn mor_namespace(&self) -> Result<Namespace, String> {
+        Ok(self.mor_namespace.clone())
+    }
```
epatters, line 353: A couple issues with these methods: (1) they shouldn't return a `Result` because they are infallible and (2) they shouldn't clone because they should just be used for lookup, which doesn't require ownership.

In fact, you should be able to avoid exposing this data at all; see below.

packages/catlog-wasm/src/theories.rs
```diff
@@ -87,6 +87,18 @@ impl ThSchema {
         );
         Ok(boxed.replace_box(model.into()))
     }
+
+    /// Renders a model into valid SQL
+    #[wasm_bindgen(js_name = "renderSql")]
+    pub fn render_sql(&self, model: &DblModel, backend: &str) -> Result<String, String> {
+        let sql_string = catlog::stdlib::analyses::sql::SQLAnalysis::new(
```
epatters: The subpackage `analyses` is already imported, so you can save a few characters and just write `sql_string = analyses::sql::[...]`

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,323 @@
...
+use itertools::Itertools;
+use sea_query::SchemaBuilder;
+use sea_query::{
+    ColumnDef, ForeignKey, ForeignKeyCreateStatement, Iden, MysqlQueryBuilder,
+    PostgresQueryBuilder, SqliteQueryBuilder, Table, TableCreateStatement, prepare::Write,
+};
+use sqlformat::{Dialect, format};
+use std::fmt;
+
+impl Namespace {
```
epatters: Remove this `impl`. (Should be a code smell to be adding this here.)

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,323 @@
...
+/// Struct for building a valid SQL DDL
+pub struct SQLAnalysis {
+    obns: Namespace,
+    morns: Namespace,
+    backend: SqlBackend,
+}
+
+impl SQLAnalysis {
+    /// Constructs a new SQLAnalysis instance
+    pub fn new(obns: Namespace, morns: Namespace, backend: &str) -> Result<SQLAnalysis, String> {
```
epatters: Replace parameter `obns: Namespace` with

```rust
ob_label: impl Fn(&QualifiedName) -> QualifiedLabel
```

and similarly for morphism labels.

packages/catlog-wasm/src/theories.rs
```diff
@@ -87,6 +87,18 @@ impl ThSchema {
...
+            model.ob_namespace()?,
```
epatters: Building on comment below, pass something like

```rust
|id| model.ob_generator_label(id).unwrap_or_else(|| QualifiedLabel::single("".into()))
```

and similarly for morphism labels.

packages/catlog-wasm/src/theories.rs
```diff
@@ -87,6 +87,18 @@ impl ThSchema {
...
+            model.mor_namespace()?,
+            backend,
+        )?
+        .render(model.discrete()?);
```
epatters: On second thought, pass the above closures here so that you don't need to store them as fields.

packages/catlog/src/stdlib/analyses/sql.rs
epatters: It's unidiomatic for a constructor in Rust to be fallible (I've never seen it, anyway). I'd make `backend` be of type `SqlBackend` and if you need to fallibly coerce from a string, do that elsewhere, perhaps in a `try_from` method.

Cargo.lock
```diff
@@ -2841,6 +2900,30 @@ dependencies = [
...
+version = "0.32.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a5d1c518eaf5eda38e5773f902b26ab6d5e9e9e2bb2349ca6c64cf96f80448c"
+dependencies = [
+ "inherent",
+ "sea-query-derive",
+]
+
+[[package]]
+name = "sea-query-derive"
```
epatters, line 3036: Do we need the `sea-query-derive` feature? If not, we can disable it to save a little on bundle size.

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,111 @@
...
+import CircleHelp from "lucide-solid/icons/circle-help";
+import Copy from "lucide-solid/icons/copy";
+import Download from "lucide-solid/icons/download";
+import { createSignal, For, Show } from "solid-js";
+
+import { BlockTitle, IconButton } from "catcolab-ui-components";
+import { ThSchema } from "catlog-wasm";
+import type { ModelAnalysisProps } from "../../analysis";
+
+/** Button to download an SVG. */
```
epatters: Stale docstring.

packages/frontend/src/stdlib/analyses.tsx
```diff
@@ -256,3 +257,16 @@ export const stockFlowDiagram = (
 });
 
 const StockFlowDiagram = lazy(() => import("./analyses/stock_flow_diagram"));
+
+export const renderSql = (
+    options: AnalysisOptions,
+): ModelAnalysisMeta<SQLDownloadConfig.DownloadConfig> => ({
+    ...options,
+    component: DownloadTextButton,
```
epatters: `DownloadTextButton` is too generic a name for a file that lists all the analyses.

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,111 @@
...
+        </div>
+    );
+}
+
+export function downloadTextContent(text: string, filename: string) {
+    return download(text, filename, "text/plain");
+}
+
+export type DownloadConfig = {
+    backend: string;
```
epatters: Are there only three valid strings for `backend`? If so, would suggest create a more specific type.

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,111 @@
...
+                    </div>
+                )}
+            </Show>
+        </div>
+    );
+}
+
+export function downloadTextContent(text: string, filename: string) {
+    return download(text, filename, "text/plain");
+}
```
epatters: Why export this function? It's only used once anyway; would suggest just deleting it.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,323 @@
+//! SQL
```
epatters: Please write a proper docstring here.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,323 @@
...
+use crate::{
+    dbl::model::DiscreteDblModel,
+    zero::{QualifiedLabel, QualifiedName},
+};
+use crate::{dbl::model::*, one::FgCategory};
+use crate::{
+    one::Path,
+    zero::{Namespace, name},
+};
```
epatters, line 6: I wish rustfmt did it automatically, but alas: these imports look messy, please clean them up a bit.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,323 @@
...
+                        // if the Id of the name is an entity, it is assumed to be a column
+                        // which references the primary key of another table.
+                        if model.mor_generator_type(mor) == Path::Id(name("Entity")) {
+                            acc.col(ColumnDef::new(mor_name.clone()).integer().not_null())
+                        } else {
+                            let tgt_name = self
+                                .obns
+                                .clone()
+                                .label_name(model.get_cod(mor).unwrap_or(&name("")).clone());
+                            match format!("{}", tgt_name.clone()).as_str() {
```
epatters: I'd suggest moving this deeply nested `match` to a helper function for rendering attribute types.

There's also a certain amount of redundancy in the cases (`ColumnDef::new(...)`) that could likely be eliminated by refactoring the code slightly.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,323 @@
...
+        batch::PARSE_CONFIG,
+        modelgen::generate,
+        text_elab::Elaborator,
+        toplevel::{Theory, Toplevel, std_theories},
+    };
+    use crate::zero::Namespace;
+    use std::rc::Rc;
+    use tattle::Reporter;
+
+    impl DiscreteDblModel {
```
epatters: It's not your fault this feature wasn't properly exposed, but regardless this impl shouldn't be here. I'll see if I can improve that separately.

epatters (changes requested):
Thanks Matt, and sorry for the long delay in returning a review. I'm excited about this feature!

I've left a first round of comments below. I'll likely have more feedback (I haven't tested anything yet :) but I wanted to get you something now.

packages/catlog-wasm/src/theories.rs
```diff
@@ -110,6 +110,30 @@ impl ThSchema {
         );
         Ok(boxed.replace_box(model.into()))
     }
+
+    /// Renders a model into valid SQL
+    #[wasm_bindgen(js_name = "renderSql")]
+    pub fn render_sql(&self, model: &DblModel, backend: &str) -> Result<String, String> {
+        match analyses::sql::SqlBackend::try_from(backend) {
+            Ok(backend) => {
```
epatters: Since you're just carrying the error variant along, the idiomatic way to write this is:

```rust
match analyses::sql::SqlBackend::try_from(backend).and_then(|backend| {
    [...]
})
```

packages/catlog-wasm/src/model.rs
```diff
@@ -707,35 +707,29 @@ pub(crate) mod tests {
     pub(crate) fn sch_walking_attr(th: &DblTheory, ids: [Uuid; 3]) -> DblModel {
         let mut model = DblModel::new(th);
         let [attr, entity, attr_type] = ids;
-        assert!(
```
epatters, line 710: Not sure what happened here, but you need to run the formatter.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,273 @@
...
+use sea_query::{
+    prepare::Write, ColumnDef, ForeignKey, ForeignKeyCreateStatement, Iden, MysqlQueryBuilder,
+    PostgresQueryBuilder, SqliteQueryBuilder, Table, TableCreateStatement,
+};
+use sqlformat::{format, Dialect};
+use std::{collections::HashMap, fmt};
+
+impl Iden for QualifiedName {
+    fn unquoted(&self, s: &mut dyn Write) {
+        ToString::to_string(&name(self.clone())).as_str().unquoted(s)
```
epatters: Why so complicated? If I'm not mistaken, this can just be:

```rust
self.to_string().unquoted(s)
```

Certainly you don't need to clone or call `name`.

Perhaps the issue is that the `to_string` methods are ambiguous for `ToString` vs `Iden` (annoying that SeaQuery did that). In that case, this would be equivalent:

```rust
format!("{self}").unquoted(s)
```

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+}
+
+impl Iden for QualifiedLabel {
+    fn unquoted(&self, s: &mut dyn Write) {
+        ToString::to_string(&self.clone()).as_str().unquoted(s)
```
epatters: Similar comments apply here.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+}
+
+impl From<QualifiedLabel> for TableCreateStatement {
+    fn from(label: QualifiedLabel) -> Self {
```
epatters: The Rust docs have excellent guidelines about [when to implement `From`](https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from). I quote this relevant passage:

> The conversion is obvious: it’s the only reasonable conversion between the two types. Otherwise it’s better to have it be a named method or constructor, like how [str::as_bytes](https://doc.rust-lang.org/std/primitive.str.html#method.as_bytes) is a method and how integers have methods like [u32::from_ne_bytes](https://doc.rust-lang.org/std/primitive.u32.html#method.from_ne_bytes), [u32::from_le_bytes](https://doc.rust-lang.org/std/primitive.u32.html#method.from_le_bytes), and [u32::from_be_bytes](https://doc.rust-lang.org/std/primitive.u32.html#method.from_be_bytes), none of which are From implementations. Whereas there’s only one reasonable way to wrap an [Ipv6Addr](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html) into an [IpAddr](https://doc.rust-lang.org/std/net/enum.IpAddr.html), thus IpAddr: From<Ipv6Addr> exists.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,273 @@
...
+            .iter()
+            .map(|table| match self.backend {
+                SqlBackend::MySQL => table.to_string(MysqlQueryBuilder),
+                SqlBackend::SQLite => table.to_string(SqliteQueryBuilder),
+                SqlBackend::PostgresSQL => table.to_string(PostgresQueryBuilder),
+            })
+            .collect();
+
+        // to ensure tests pass consistently. However, sea_query should probably allow multiple
+        // tables
```
epatters: Confused by this comment. Is the issue that the ordering has become unpredictable due to using a hash map for `morphisms` above? In that case, the better solution is to replace `HashMap` with `IndexMap` like we're now doing for the underlying data structures.

packages/catlog/src/stdlib/analyses/sql.rs
epatters: > I'd suggest moving this deeply nested `match` to a helper function for rendering attribute types.

I think this comment still applies.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,273 @@
...
+                    },
+                );
+
+                mors.iter()
+                    .filter(|mor| model.mor_generator_type(mor) == Path::Id(name("Entity")))
+                    .fold(
+                        // TABLE AND COLUMN DEFS
+                        table_column_defs,
+                        |acc, mor| {
+                            let tgt = model.get_cod(mor).unwrap(); // TODO
```
epatters: Lingering TODO. I agree that the unwrap here is in general not safe.

There are two possibilities for how to approach. You can write an analysis that is safe to be run (will not panic or raise exceptions) even for models that do not validate, say in this case because a codomain is undefined. That's the approach taken by most of the visualization analyses. Or you can write an analysis that should only be run on models that are validated. That's the approach taken for most of the simulation analyses. Then you can just unwrap here or call `model.cod`.

No strong opinion on which approach you take, but you should choose one and document it clearly. In the latter approach, you also then to ensure that the frontend invokes the analysis only when the model validates.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn as_type(&self) -> Box<dyn SchemaBuilder> {
+        match self {
+            SqlBackend::MySQL => Box::new(MysqlQueryBuilder),
+            SqlBackend::SQLite => Box::new(SqliteQueryBuilder),
+            SqlBackend::PostgresSQL => Box::new(PostgresQueryBuilder),
+        }
+    }
+}
+
+impl From<SqlBackend> for Dialect {
```
epatters: By contrast, this an appropriate use of `From`, IMO.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,273 @@
...
+                Person : Entity,
+                Dog : Entity,
+                walks : (Hom Entity)[Person, Dog],
+                Hair : AttrType,
+                has : Attr[Person, Hair],
+            ]",
+            &th.into(),
+        );
+
+        let raw_creates = [
```
epatters: Would strongly recommend using `expect_test` for this test. This is an ideal use case for that crate.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -0,0 +1,273 @@
...
+        let morns = Namespace::new_for_text();
+
+        if let Some(m) = &model.and_then(|m| m.as_discrete()) {
+            let ddl = SQLAnalysis::new(SqlBackend::MySQL).render(
+                m,
+                |id| obns.label(id).unwrap_or(QualifiedLabel::single("".into())),
+                |id| morns.label(id).unwrap_or(QualifiedLabel::single("".into())),
+            );
+
+            // TODO Hash is nondeterministic
```
epatters: Ah yes, you should fix this then. See my comment above.

packages/catlog/Cargo.toml
```diff
@@ -38,6 +38,8 @@ ustr = "1"
 uuid = { version = "1.18" }
 wasm-bindgen = { version = "0.2.100", optional = true }
 notebook-types = { version = "0.1.0", path = "../notebook-types" }
+sea-query = "0.32.7"
+sqlformat = "0.5.0"
```
epatters: For now, these dependencies should be feature-gated, like we do for the simulation modules.

Eventually I'd like to get most of the analysis stuff out of catlog through a more modular design, but that's for another time.

packages/frontend/src/components/sql_ui.tsx
```diff
@@ -0,0 +1,52 @@
...
+            <li>{"Float"}</li>
+            <li>{"Bool"}</li>
+            <li>{"Time"}</li>
+            <li>{"Date"}</li>
+            <li>{"DateTime"}</li>
+        </ul>
+    </>
+);
+
+// TODO this was factored out of sql.tsx since the lucide icons are expecting the browser API to be available.
```
epatters: I see why you did this, but I'd suggest another approach. Merge this file back into `sql.tsx`. Then create a tiny file `stdlib/analyses/sql_types.ts` (no JSX at all) containing the enum `SqlBackend`, which is all you need to import in the `analyses.tsx`.

packages/frontend/src/help/logics/simple-schema.mdx
```diff
@@ -84,3 +84,85 @@ In other words, the set corresponding to an attribute type starts as empty and n
 
 <HelpAnalysisById theory={props.theory} analysisId="diagram"/>
 <HelpAnalysisById theory={props.theory} analysisId="erd"/>
+
+<HelpAnalysisById theory={props.theory} analysisId="sql"/>
```
epatters, line 88: Thanks for writing these docs!

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { SqlHeader } from "../../components";
+
+export enum SqlBackend {
+    MySQL = "MySQL",
+    SQLite = "SQLite",
+    PostgresSQL = "PostgresSQL",
+}
+
+/** Component to interface with SQL analysis. Allows user to download the script and change the backend. */
+export default function SqlSchemaInterface(props: ModelAnalysisProps<DownloadConfig>) {
```
epatters: I'd suggest `SqlSchemaAnalysis` as a more informative name.

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const thSchema = new ThSchema();
+
+    const [backend, setBackend] = createSignal(SqlBackend.MySQL);
+    const sqlOutput = () => {
+        const model = props.liveModel.elaboratedModel();
```
epatters, line 64: If you decide that your analysis should require a validated model, this would be the appropriate place to enforce that.

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,66 @@
...
+                }
+            >
+                <For each={Object.values(SqlBackend)}>
+                    {(bknd) => <option value={bknd}>{bknd}</option>}
+                </For>
+            </select>
+        </div>
+    );
+
+    const title = () => "SQL Schema";
```
epatters: There's not much reason to do this, IMO. I'd just put the string literal directly in the JSX.

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export enum SqlBackend {
+    MySQL = "MySQL",
+    SQLite = "SQLite",
+    PostgresSQL = "PostgresSQL",
+}
+
+/** Component to interface with SQL analysis. Allows user to download the script and change the backend. */
+export default function SqlSchemaInterface(props: ModelAnalysisProps<DownloadConfig>) {
+    const thSchema = new ThSchema();
```
epatters: Rather than instantiating a new theory here, have the analysis take a `renderSql` function as a prop. This is the pattern used by the simulation analyses, see also `simulation_types.ts`.

epatters (changes requested):
Thanks Matt, I've left what should be the final round of comments!

packages/catlog/src/stdlib/analyses/sql.rs
quffaro: Thanks, this makes sense

quffaro: Turns out this code was unused!

PR #845, opened 2025-11-11
Simplify deduplication of deps in vite config

kasbah:
Following on from #841 I realised vite config executes so we can actually compute the list in the config.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
       - name: Setup NodeJS
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 24
           cache: "pnpm"
 
       - name: Install Rust toolchain from file
@@ -119,7 +119,7 @@ jobs:
       - name: Setup NodeJS
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 24
           cache: "pnpm"
 
       - name: Install dependencies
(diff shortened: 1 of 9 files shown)


packages/frontend/vite.config.ts
```diff
@@ -49,3 +41,24 @@ export default defineConfig({
...
+    const uiComponentsPkg = JSON.parse(
+        readFileSync(resolve(__dirname, "../ui-components/package.json"), "utf-8"),
+    );
+
+    const frontendDeps = new Set(Object.keys(frontendPkg.dependencies || {}));
+    const uiComponentsDeps = new Set(Object.keys(uiComponentsPkg.dependencies || {}));
+
+    // @ts-expect-error: intersection method does exist on Set in our
+    // vite.config target i.e. NodeJS
+    const commonDeps = frontendDeps.intersection(uiComponentsDeps);
```
epatters, line 61: Looks like this method is supported in Node v22+: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection#browser_compatibility

We should bump the Node version used in CI to the latest LTS, which is apparently v24. Currently the CI is pegged to v20.

epatters:
It's marked as a draft, but this PR looks about ready. Anything left to do here?

kasbah:
It's just that there will be a merge conflict with #846 which needs to be manually resolved through a bit of an annoying (albeit fairly straight-forward) process: https://github.com/ToposInstitute/CatColab/pull/846#issuecomment-3516740954

kasbah:
Merge done (well, rebase actually). All good now.

PR #846, opened 2025-11-11
Update Biome linter and enable "domain" for Solid linting

kasbah:
Mainly I am doing this to enable the [Solidjs domain](https://biomejs.dev/linter/domains/#solid) in Biome. Some churn due to import ordering and auto-formatting changing slightly, unfortunately it seems.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -19,7 +19,18 @@
         "enabled": true,
         "actions": {
             "source": {
-                "organizeImports": "on"
+                "organizeImports": {
+                    "level": "on",
+                    "options": {
+                        "groups": [
+                            ":NODE:",
+                            [":PACKAGE:", "!catcolab-ui-components", "!catcolab-api", "!catlog-wasm"],
+                            ":BLANK_LINE:",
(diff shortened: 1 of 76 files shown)


kasbah:
Any ideas @jmoggr ? https://github.com/ToposInstitute/CatColab/actions/runs/19265781014/job/55081142138?pr=846#step:5:2376 

EDIT: Never mind, found [the README](https://github.com/ToposInstitute/CatColab/tree/8928d66eb65d57248c97ec72adff0dc485851f3d/packages/automerge-doc-server#updating-pnpm-dependencies-with-nix). This is really how we are going to do this, huh. :thinking:

jmoggr:
> This is really how we are going to do this, huh. 🤔

Yeah... it's not great. I'm open to ideas, but within the constraints of Nix I think this is the best we can hope for for now. This will hopefully be improved in the future, we no longer have this problem with rust for example. The fragmentation of the npm ecosystem has been carried over into tooling for Nix, slowing the development of good tooling.

The sadder thing here is that we need a separate service/package for this at all, ideally this would all be handled by the backend. It's just sadness all the way down...

PR #849, opened 2025-11-11
Make netlify add "deployment" info to PR rather than comment

kasbah:
This way we don't get a notification from the github bot comment and it's generally less intrusive. We do lose the ui-components link but we can remember to go to `/dev/ui-components`.

I overengineered it a bit so we get live status updates and previous ones are marked as destroyed when the new one succeeds.

image

---

image

---

image

---

image


epatters (approved):
Thanks for this deluxe CI experience :) And it will be nice to avoid getting an extra notification.

PR #851, opened 2025-11-12
Add a `LiveDocWithRef` type and name `Doc` types consistently

kasbah:
Removes the optional `docRef` from `LiveDoc` and instead we use a `LiveDocWithRef = {liveDoc: LiveDoc, docRef: DocRef}`  where appropriate.

### DIFF
--- a/packages/frontend/src/analysis/document.ts
+++ b/packages/frontend/src/analysis/document.ts
@@ -3,13 +3,13 @@ import type { AnyDocumentId, Repo } from "@automerge/automerge-repo";
 import {
     type Analysis,
     type AnalysisType,
+    currentVersion,
     type Document,
     type StableRef,
     type Uuid,
-    currentVersion,
 } from "catlog-wasm";
-import { type Api, type DocRef, type LiveDoc, findAndMigrate, makeLiveDoc } from "../api";
-import { type LiveDiagramDoc, getLiveDiagram, getLiveDiagramFromRepo } from "../diagram";
+import { type Api, type DocRef, findAndMigrate, type LiveDoc, makeLiveDoc } from "../api";
+import { getLiveDiagram, getLiveDiagramFromRepo, type LiveDiagramDoc } from "../diagram";
 import type { LiveModelDoc, ModelLibrary } from "../model";
 import { newNotebook } from "../notebook";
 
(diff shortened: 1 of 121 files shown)


packages/frontend/src/analysis/document.ts
```diff
@@ -77,34 +77,42 @@ export async function createAnalysis(api: Api, analysisType: AnalysisType, analy
     return api.createDoc(init);
 }
 
+export type LiveAnalysisLiveDocWithRef = {
```
epatters, line 80: Would shorten this name to `LiveAnalysisDocWithRef`, both to avoid saying "live" twice and for consistency with `LiveAnalysisDoc` above.

packages/frontend/src/diagram/document.ts
```diff
@@ -130,17 +130,23 @@ export function createDiagram(api: Api, inModel: StableRef): Promise<string> {
     return api.createDoc(init);
 }
 
+export type LiveDiagramLiveDocWithRef = {
```
epatters, line 133: Ditto: would shorten this name to `LiveDiagramDocWithRef`.

packages/frontend/src/page/document_page.tsx
```diff
@@ -302,14 +333,21 @@ async function getLiveDocument(
...
     switch (documentType) {
-        case "model":
-            return models.getLiveModel(refId);
-        case "diagram":
-            return getLiveDiagram(refId, api, models);
-        case "analysis":
-            return getLiveAnalysis(refId, api, models);
+        case "model": {
+            const { liveDoc: _, docRef } = await api.getLiveDoc(refId, "model");
+            const liveDocument = await models.getLiveModel(refId);
```
epatters: Seems a bit wasteful to construct the live document twice, discarding the first one.

There might be other ways to refactor this, but the first that comes to mind is:

- delete the `getPermissions` and `isDocumentDeleted` methods of `Api`
- replace them both with a single method `getDocRef`

Then this code can be replaced with

```typescript
const docRef = await api.getDocRef(refId);
const liveDocument = await models.getLiveModel(refId);
```

epatters (changes requested):
Thanks for cleaning this situation up. I agree that it was a bit messy. Thanks also for paying close attention to consistency in naming.

A few small comments below.

kasbah:
Thanks @epatters, comments are all addressed.

epatters:
@kasbah, it looks like the changes you made in response to my comments were somehow lost before this PR was merged. I know you made the changes because I saw them with my own eyes :) Do you still have the relevant commits lying around?

kasbah:
I messed up merging main branch into this somehow in a way that got rid of these changes: #855

PR #858, opened 2025-11-16
Prevent wrapping in document title in sidebar

perrybleiberg:
Resolves #819

PR #861, opened 2025-11-18
Split build/deploy allowing deploy from external PRs

kasbah:
Deploy is now triggered by `workflow_run` of the build and thus will be run even on PRs from external contributors.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,190 @@
+name: build
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+
+jobs:
+  save_pr_metadata:
+    name: Save PR metadata
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    outputs:
+      pr_number: ${{ steps.metadata.outputs.pr_number }}
+      pr_branch: ${{ steps.metadata.outputs.pr_branch }}
+      pr_sha: ${{ steps.metadata.outputs.pr_sha }}
+      pr_repo: ${{ steps.metadata.outputs.pr_repo }}
+      base_ref: ${{ steps.metadata.outputs.base_ref }}
+    steps:
+      - name: Save PR metadata
+        id: metadata
+        run: |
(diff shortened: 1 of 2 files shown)


jmoggr:
1. I'm very nervous about having other peoples code run in our CI, how certain are we that they can't exfiltrate secrets? For example could they open a PR with changes to our github actions which then run?

2. This looks rather complicated and it seems that there is more going on then the PR title indicates. Github actions are a nightmare to debug, so I would lean towards keeping them as simple as possible.

kasbah:
Those are fair concerns @jmoggr. 

1. 
    - As before, the running of any of our actions is dependant on manual approval for first-time PR creators. The new "deploy" workflow is triggered by running of the "build" workflow which needs the manual approval. 
    - The "deploy" workflow is run in context of our main branch rather than the PR branch. A PRs modifying the "deploy" workflow does not affect any actions run until it is merged.
    - Coming to think of it, we can improve our security posture by making a "production" GH environment that always requires maintainer approval for access, no matter what.
    
2.  I don't disagree with keeping our CI as simple as possible but I do want these quality of life improvements such as the nicer inline deploy notifications (a big part of the complexity -- but this is already on main) and getting deploy previews for external contributions. Happy for any suggestions to simplify any of this.

epatters:
What's the status of this PR, @kasbah?

kasbah:
The question really is: how much do we care about external PRs getting useful CI runs?

epatters:
The current situation is that when we get a PR from a new contributor, the CI won't run until I manually press a button that says "Approve workflows." That seems OK to me since it doesn't happen very often. So, since there are security concerns with running the CI on untrusted code, I'll close this for now. Happy to revisit later if there is demand.

kasbah:
I am pretty sure you don't get the option to do a preview deploy to Netlify for an external PR. E.g on https://github.com/ToposInstitute/CatColab/pull/858, that was the reason for this. 

This PR would allow you to build and deploy external PRs _only with manual approval_.

epatters:
Ok, sorry, I had forgotten that issue. I agree that it would be useful to get preview deploys of external PRs that have been manually approved. You can decide how much you'd like to prioritize getting this through, @kasbah. I think @jmoggr should be the primary reviewer (consistent with our new code owners listing).

PR #864, opened 2025-11-21
Fix extra vertical space for long document titles

### LINKED ISSUE #818: Long document titles cause extra vertical space
In the new paned interface (#749), when a document title is too long to fit, it truncates (not unreasonable) but also causes extra vertical space to appear above it, which is unnecessary and distracting:

Image

perrybleiberg:
Resolves #818. Even when the text in an InlineInput does not wrap, line break opportunities in the input force the hidden inline-input-filler span to account for wrapping, causing the extra space. The change prevents this from happening in document-head.

### DIFF
--- a/packages/frontend/src/page/document_page.css
+++ b/packages/frontend/src/page/document_page.css
@@ -47,6 +47,10 @@
     & .lucide {
         color: var(--link-color);
     }
+
+    & .inline-input-filler {
+        text-wrap-mode: nowrap;
+    }
 }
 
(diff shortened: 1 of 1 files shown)


perrybleiberg:
Thanks for taking a look Kaspar, I'll try again and/or follow your lead here.

epatters:
Closing due to staleness. Feel free to reopen if you'd like to address Kaspar's feedback.

PR #865, opened 2025-11-21
Theory-specific icons with two letters for models

kasbah:
image

- https://branch-865-merge--catcolab.netlify.app/dev/ui-components/?path=/docs/icons-documenttypeicon--docs 
- https://branch-865-merge--catcolab.netlify.app/dev/ui-components/?path=/docs/icons-internal-modelfileicon--docs

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -49,20 +49,48 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          # Need full history for `git log` for cache key
+          fetch-depth: 0
+
+      - name: Compute cache key
+        id: cache-key
+        run: |
+          hash=$(git log -1 --format="%H" -- dev-docs)
+          echo "git-hash=${hash}" >> $GITHUB_OUTPUT
+          echo "Dev docs hash: ${hash}"
+
+      - name: Cache dev docs build
+        id: cache-dev-docs-build
+        uses: actions/cache@v4
+        with:
+          path: dev-docs/output
+          key: dev-docs-build-${{ steps.cache-key.outputs.git-hash }}
 
       - name: Setup pnpm
+        if: steps.cache-dev-docs-build.outputs.cache-hit != 'true'
         uses: pnpm/action-setup@v4
 
       - name: Setup NodeJS
+        if: steps.cache-dev-docs-build.outputs.cache-hit != 'true'
         uses: actions/setup-node@v4
         with:
           node-version: 24
           cache: "pnpm"
 
       - name: Build developer docs
+        if: steps.cache-dev-docs-build.outputs.cache-hit != 'true'
         run: |
           pnpm install
           pnpm --filter ./dev-docs run doc
+
+      - name: Report cache status
+        run: |
+          if [ "${{ steps.cache-dev-docs-build.outputs.cache-hit }}" = "true" ]; then
+            echo "✅ Dev docs restored from cache"
+          else
+            echo "🔨 Dev docs built and cached"
+          fi
 
       - name: Upload developer docs
         uses: actions/upload-artifact@v4
@@ -76,24 +104,48 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          # Need full history for `git log` for cache key
+          fetch-depth: 0 
+
+      - name: Compute cache key
+        id: cache-key
+        run: |
+          # Get git hash of the most recent commit affecting frontend or any of its dependencies
+          hash=$(git log -1 --format="%H" -- packages/frontend packages/backend/pkg packages/catlog packages/catlog-wasm packages/notebook-types packages/ui-components Cargo.toml Cargo.lock)
+          echo "git-hash=${hash}" >> $GITHUB_OUTPUT
+          echo "Frontend git hash: ${git_hash}"
+
+      - name: Cache frontend build
+        id: cache-frontend-build
+        # Skip caching for production builds
+        if: github.event_name != 'release'
+        uses: actions/cache@v4
+        with:
+          path: |
+            packages/frontend/dist
+          key: frontend-build-${{ steps.cache-key.outputs.git-hash }}-${{ github.event_name == 'release' && 'production' || 'staging' }}
 
       - name: Setup pnpm
+        if: github.event_name == 'release' || steps.cache-frontend-build.outputs.cache-hit != 'true'
         uses: pnpm/action-setup@v4
 
       - name: Setup NodeJS
+        if: github.event_name == 'release' || steps.cache-frontend-build.outputs.cache-hit != 'true'
         uses: actions/setup-node@v4
         with:
           node-version: 24
           cache: "pnpm"
 
       - name: Install Rust toolchain from file
+        if: github.event_name == 'release' || steps.cache-frontend-build.outputs.cache-hit != 'true'
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           # Don't override flags in cargo config files.
           rustflags: ""
 
       - name: Build for Staging
-        if: github.event_name == 'push' || github.event_name == 'pull_request'
+        if: (github.event_name == 'push' || github.event_name == 'pull_request') && steps.cache-frontend-build.outputs.cache-hit != 'true'
         run: |
           pnpm install
           # Uses env from packages/frontend/.env.staging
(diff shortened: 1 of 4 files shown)


packages/ui-components/src/global.css
```diff
@@ -1,3 +1,5 @@
+@import url("https://fonts.googleapis.com/css2?family=Inter:ital,opsz,wght@0,14..32,100..900;1,14..32,100..900&family=Roboto+Mono:ital,wght@0,100..700;1,100..700&family=Source+Code+Pro:ital,wght@0,200..900;1,200..900&display=swap");
```
kasbah: This also loads the other fonts we use. I guess we have been falling back when not available?

packages/ui-components/src/model_file_icon.tsx
```diff
@@ -0,0 +1,95 @@
...
+            width={localProps.size ?? defaultAttributes.width}
+            height={localProps.size ?? defaultAttributes.height}
+            stroke={localProps.color ?? defaultAttributes.stroke}
+            stroke-width={
+                localProps.absoluteStrokeWidth
+                    ? (Number(localProps.strokeWidth ?? defaultAttributes["stroke-width"]) * 24) /
+                      Number(localProps.size)
+                    : Number(localProps.strokeWidth ?? defaultAttributes["stroke-width"])
+            }
+            class={mergeClasses(
```
epatters: Should be possible to simplify this using Solid's builtin [`classList`](https://docs.solidjs.com/reference/jsx-attributes/classlist):

```typescript
class="lucide lucide-icon lucide-file-catcolab"
classList={{ [localProps.class as string]: localProps.class != null }}
```

packages/ui-components/src/util/css.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Merges classes into a single string
+ *
+ * @param {array} classes
+ * @returns {string} A string of classes
+ */
+export const mergeClasses = <ClassType = string | undefined | null>(
```
epatters: It's not a very big deal, but I'm skeptical that we need this helper function. (It also seems a bit overcomplicated in having the type parameter `ClassType`.) See my suggested alternative above.

packages/ui-components/src/document_type_icon.tsx
```diff
@@ -4,14 +4,23 @@ import FileX from "lucide-solid/icons/file-x";
 import Network from "lucide-solid/icons/network";
 import { Match, Switch } from "solid-js";
 
-import type { DocumentType } from "../api";
+import { ModelFileIcon } from "./model_file_icon";
 
-export function DocumentTypeIcon(props: { documentType: DocumentType; isDeleted?: boolean }) {
+export type DocumentType = "model" | "diagram" | "analysis";
```
epatters, line 9: I'm worried about having this enum hard-coded here, as the list of document types *will* grow over time. The source of truth for this type is:

```typescript
import type { Document } from "catlog-wasm";

type DocumentType = Document["type"];
```

packages/ui-components/src/global.css
epatters: Thanks for noticing this. TBH, I don't recall how that font got there nor do I even recognize the font name. Likely it got copied from somewhere. All of that is to say that our font choices are fair game to be revised at any time :)

packages/ui-components/src/model_file_icon.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+                    <ModelFileIcon theory={modelType} size={48} />
+                </div>
+            )}
+        </For>
+    ),
+};
+
+export const DifferentColors: Story = {
+    render: () => (
+        <For each={["#000", "#fff", "#f00", "#0f0", "#00f"]}>
```
epatters, line 71: I can't help but notice that the red, green, and blue colors here are very ugly. Not a priority right now but I figure at some point we should have a standard palette that we're using for the app? E.g., the alerts have more palatable choices of colors.

packages/ui-components/src/model_file_icon.tsx
```diff
@@ -0,0 +1,95 @@
...
+    "simple-schema": ["S", "c"],
+    "petri-net": ["P", "n"],
+    "causal-loop": ["C", "l"],
+    "causal-loop-delays": ["C", "d"],
+    "indeterminate-causal-loop": ["C", "i"],
+    "primitive-stock-flow": ["S", "f"],
+    "reg-net": ["R", "n"],
+    "unary-dec": ["D", "c"],
+    "power-system": ["P", "s"],
+};
```
epatters: I'm also worried about having to update this list. It's already harder than I'd like for people to add new theories. To keep the data in one place, we could add a `code` field to the [theory metadata](https://github.com/ToposInstitute/CatColab/blob/814bfc72ac957e7e1d0972c5b2b7d4eded072fef/packages/frontend/src/theory/theory_library.ts#L8) defined in the frontend, rather than defining the table here.

epatters (changes requested):
Thanks Kaspar, this is a fun feature!

packages/ui-components/src/global.css
kasbah: In that case I'll keep `Inter` but reduce this to just load *one* monospace font: `Source Code Pro` since that works well in the icons. I'll get rid of `Roboto Mono` which we had been using (but not loading explicitly).

packages/ui-components/src/model_file_icon.stories.tsx
kasbah: Yes, this was mainly to test that the color prop is being applied just like with regular Lucide icons. `ModelFileIcon` and these stories are conceptually internal to ui-components and the color option isn't exposed via `DocumentTypeIcon` which is what we use in the frontend.

We should indeed make a standard palette though.

packages/ui-components/src/util/css.ts
kasbah: Thx, I copied this from lucide-solid and didn't know about `classList`. Fixed.

packages/ui-components/src/document_type_icon.tsx
kasbah: I tried that with [076fc47](https://github.com/ToposInstitute/CatColab/pull/865/commits/076fc47e46de76b0527e642f329b376873c5175c) but [reverted it](https://github.com/ToposInstitute/CatColab/pull/865/commits/ae86612a082e79579fd0414b83e9a3bab693850b) because having a `ui-components` depend on `catlog-wasm` is too annoying in our current setup.

Additionally, I may be rationalizing but it's actually more correct to re-define the type here. The `<Switch>` only covers the cases we have in the type so when it updates both need to be updated.

https://github.com/ToposInstitute/CatColab/pull/865/files#diff-acdbbbcd75e959c9498d8b107213a2288e5856dead5ba6751a4bf249d8893caeR18-R32

packages/ui-components/src/model_file_icon.tsx
kasbah: I've made these part of theory metadata.

kasbah:
I've addressed all feedback.

packages/ui-components/src/model_file_icon.stories.tsx
epatters: It might make sense to change these colors now that #888 is merged, but I leave that decision to you.

packages/ui-components/src/document_type_icon.stories.tsx
```diff
@@ -0,0 +1,163 @@
+import { For } from "solid-js";
+import type { Meta, StoryObj } from "storybook-solidjs-vite";
+
+import { type DocumentType, DocumentTypeIcon } from "./document_type_icon";
+
+const theoryToLetterMap: Record<string, [string, string]> = {
```
epatters, line 6: Shouldn't this mapping now be distributed across the theories in `stdlib/theories`?

epatters (changes requested):
Sorry for letting this one linger! I'm afraid you'll have to fix a merge conflict. I also had one bit of substantive feedback.

packages/ui-components/src/model_file_icon.tsx
epatters: Thanks! I'm confused though about how far this integration goes, see my [inline comment](https://github.com/ToposInstitute/CatColab/pull/865#discussion_r2624517475).

packages/ui-components/src/document_type_icon.stories.tsx
kasbah: It is there but we don't want to depend on the frontend from here. `DocumentTypeIcon` is just a component that can display any two letters now. For the stories we make it display the letters we know will be used with it. It's not important if the story is out of date and not using exactly the same letters as new theories get added. It's just an example of usage.

epatters: Ah yes, sorry. I misread which file this mapping is in. That makes sense.

epatters (approved):

PR #866, opened 2025-11-21
Add scripts and documentation for tests and common development flows

jmoggr:


dev-docs/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+### [cc-utils db](docs/references/cc-utils-db.md)
+Reference for `cc-utils db` commands used to manage databases across environments.
+
+## Quick Start
+
+**Preview remote branch:**
+```bash
+cc-utils vm start my-feature-branch --db cached
+```
```
kasbah, line 26: It might just be me but I was confused by the combination of the words "remote branch" and the lack of an `origin/` prefix. When I saw a comment later calling it a "Github branch" it cleared it up for me. I think we should use the words "Github branch" everywhere. Maybe we should even explicitly say "CatColab Github repo branch"

The other thing this has without any explanation is the `--db cached` thing. Maybe that shouldn't be in the quickstart?

dev-docs/README.md
```diff
@@ -0,0 +1,40 @@
+# Developer Documentation
```
kasbah, line 1: Is there a reason this is a separate document from CONTRIBUTING.md? It should at least be linked from it.

dev-docs/docs/references/cc-utils-db.md
```diff
@@ -0,0 +1,52 @@
+# cc-utils DB Commands Reference
+
+The `cc-utils db` commands provide tools for managing databases across environments (local, staging, production, vm).
```
kasbah, line 3: As mentioned at dev meeting we need to document how to get `cc-utils` onto your path i.e. mention `nix develop` and any alternatives.

dev-docs/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+This directory contains documentation for CatColab developers working on testing, deployment, and local development workflows.
+
+## Guides
+
+### [Testing Guide](docs/testing-guide.md)
+Learn how to run integration tests, test deployments, and debug test failures with a progressive isolation strategy.
+
+### [Development Workflows](docs/workflows.md)
+Example workflows demonstrating how to use `cc-utils` script for common development tasks.
```
kasbah, line 11: Small thing but running just `cc-utils`  on my system prints:

```
/home/kaspar/projects/topos/CatColab/infrastructure/scripts/cc-utils: line 29: $1: unbound variable
```

but exits with 0 i.e. success status while doing nothing. `cc-utils db` etc seems to be working. Would be nice to exit with `-1` and print "Error: no command given" and then show the help.

dev-docs/docs/references/cc-utils-vm.md
```diff
@@ -0,0 +1,82 @@
...
+**Database options:**
+- No `--db` flag (default) - Load fresh staging database dump
+- `--db skip` - Don't load any database
+- `--db cached` - Use most recent staging dump in dumps/, or create new if missing
+- `--db <file>` - Load from specific dump file path
+
+**Examples:**
+```bash
+# Start local VM with fresh staging DB
+cc-utils vm start
```
kasbah, line 27: My experience running this so far is it building a lot of things and timing out.

```
Starting local VM...
Logs: cc-utils vm logs
Waiting for VM to start...
Error: Timeout waiting for VM to be available
Check logs: /home/kaspar/.cache/catcolab/vm.log
```

In the background it's still building and I can tail that log 


1. I think this should log to stdout so I can see what's going on
2. The timeout needs to be adjusted or something, maybe there should be separate `build` and `start` commands?

kasbah (changes requested):
I've started seeing how I can use these scripts in my own development workflow so thanks for developing and documenting them. 

In general I am a bit apprehensive at the prospect of maintaining this much bash. We should run `shellcheck` on it in CI at the very least. In the future we should think about starting to use a more robust scripting language (or even non-scripting language?) for things like this.

As you will see I got to the point just short of actually running the VM in a useful way. I will continue to try and use these tools and do a second pass at review for sure.

dev-docs/docs/references/cc-utils-vm.md
kasbah: When I run it today it seems to use the built VM and completes instantly and I can do `cc-utils vm connect`.

dev-docs/README.md
kasbah: Looking at the docs for `--db cached` it seems like it is a confusing name. Shouldn't it be `--db staging`?

jmoggr:
We are taking a simpler approach to scripts and docs for now

PR #867, opened 2025-11-21
Upgrade to match biome versions between NPM and Nix

jmoggr:
Biome installed from NPM does not play nicely with NixOS, so I use the version of biome provided by the Nix development shell. The NPM version was recently upgraded and the Nix version was not, which was causing failures to parse the biome config files.

I needed to upgrade the unstable Nix packages to get a recent enough version of biome (2.3.6). I upgraded the NPM biome to match (from 2.3.4), since that was somewhat easier than finding the nixpkgs commit with the correct version of biome.

### DIFF
--- a/packages/automerge-doc-server/src/server.ts
+++ b/packages/automerge-doc-server/src/server.ts
@@ -32,7 +32,6 @@ function migrateDocument(doc: unknown): { Ok: any } | { Err: string } {
 export class AutomergeServer implements SocketIOHandlers {
     private docMap: Map<string, DocHandle<unknown>>;
 
-    private app: express.Express;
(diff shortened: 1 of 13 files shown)


packages/automerge-doc-server/src/server.ts
```diff
@@ -43,8 +42,8 @@ export class AutomergeServer implements SocketIOHandlers {
     constructor(port: number | string) {
         this.docMap = new Map();
 
-        this.app = express();
-        this.server = this.app.listen(port);
+        const app = express();
+        this.server = app.listen(port);
```
jmoggr, line 46: Fixes new lint error

PR #868, opened 2025-11-21
Enable rich text in readonly documents

### LINKED ISSUE #732: Rich text is broken in read-only documents
When a notebook is loaded from the backend in read-only mode, rich text annotations don't appear and the plain text that remains looks corrupted. On inspection, the reason is clear: as a vestige from our JSON-first, rather than Automerge-first, days, loading a document in read-only mode returns only the JSON data, in which the Automerge-specific rich text annotations cannot be represented.

To fix this, when a read-only document is requested, the backend should return a copy of the Automerge document in binary format, which can then be [`import`-ed](https://automerge.org/automerge-repo/classes/_automerge_automerge-repo.Repo.html#import) into the dummy Automerge repo on localhost.

jmoggr:
Closes #732

This is stacked on top of #867

### DIFF
--- a/packages/automerge-doc-server/src/server.ts
+++ b/packages/automerge-doc-server/src/server.ts
@@ -1,5 +1,4 @@
 import type * as http from "node:http";
-import { save } from "@automerge/automerge";
 import { type DocHandle, type DocumentId, Repo, type RepoConfig } from "@automerge/automerge-repo";
 import { NodeWSServerAdapter } from "@automerge/automerge-repo-network-websocket";
 import dotenv from "dotenv";
@@ -169,18 +168,10 @@ export class AutomergeServer implements SocketIOHandlers {
     }
 
     async exportDoc(docId: string): Promise<ExportDocSocketResponse> {
-        const handle = await this.repo.find(docId as DocumentId);
-        if (!handle) {
-            return { Err: `Failed to find doc handle in repo for doc_id '${docId}'` };
+        const binaryData = await this.repo.export(docId as DocumentId);
+        if (!binaryData) {
+            return { Err: `Failed to find document in repo with ID '${docId}'` };
         }
-
-        await handle.whenReady();
-        const doc = handle.doc();
-        if (!doc) {
-            return { Err: `Document ${docId} is not ready` };
-        }
-
-        const binaryData = save(doc);
         return {
             Ok: {
                 binaryData: Array.from(binaryData),
--- a/packages/automerge-doc-server/package.json
+++ b/packages/automerge-doc-server/package.json
@@ -13,7 +13,6 @@
     "author": "CatColab developer team",
     "license": "MIT",
     "dependencies": {
-        "@automerge/automerge": "^3.1.1",
         "@automerge/automerge-repo": "^2.2.0",
         "@automerge/automerge-repo-network-websocket": "^2.2.0",
         "dotenv": "^16.5.0",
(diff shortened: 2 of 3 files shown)


packages/automerge-doc-server/src/server.ts
```diff
@@ -163,6 +168,26 @@ export class AutomergeServer implements SocketIOHandlers {
...
+        }
+
+        const binaryData = save(doc);
+        return {
+            Ok: {
+                binaryData: Array.from(binaryData),
+            },
+        };
+    }
+
```
epatters, line 181: I believe this method could be simplified to:

```typescript
async exportDoc(docId: AnyDocumentId): Promise<ExportDocSocketResponse> {
    const binaryData = await this.repo.export(docId);
    if (!binaryData) {
        return { Err: `Failed to find document in repo with ID '${docId}'` };
    }
    return {
        Ok: {
            binaryData: Array.from(binaryData),
        },
    };
}
```

packages/frontend/src/api/types.ts
```diff
@@ -119,7 +119,14 @@ export class Api {
         if (isLive) {
             docId = refDoc.docId as DocumentId;
         } else {
-            const docHandle = this.localRepo.create(refDoc.content);
+            // Decode base64 string to Uint8Array
```
epatters: Apparently there is a new method that does this as a one-liner: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64

kasbah: Why don't we send the data as an array of numbers instead of base64 encoding it?

jmoggr: > Why don't we send the data as an array of numbers instead of base64 encoding it?

Because Base64 is dramatically more space efficient when compared to a stringified JSON array. 255, the largest 8bit value, would be sent as the string "255," taking 4 bytes, the same value encoded in Base64 would take ~1.3 bytes. The space savings (~30-60%?) are large enough compared to the complexity (very low) that this optimization is worth doing preemptively. Also it is fairly standard to encode binary data as Base64.

You could stream the data over http, however that would break our nice rpc interface.

I noticed that we are not using gzip compression on data sent from our backend, and we might want to be. I'll open an issue for that.

jmoggr: > Apparently there is a new method that does this as a one-liner: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64

That is very cool and very new! Unfortunately the types do not seem to have made it into typescript. The changes were introduced in [this proposal](https://github.com/tc39/proposal-arraybuffer-base64?utm_source=chatgpt.com), however it seems that only some of the types from that proposal have made it into typescript. [ArrayBufferLike](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-7.html#typedarrays-are-now-generic-over-arraybufferlike), which was part of that proposal, is available, but not the new `fromBase64` prototype property, [as you can see in this TS nightly playground](https://www.typescriptlang.org/play/?ts=6.0.0-dev.20251125#code/KoSwdgLgHAggTnAhgTwHQDM4HsC2AhRAZwFMA2AFgAoBKIA).

I'm somewhat confused by this. I've dug around in the commit history and my best explanation so far is that someone dropped the ball on adding the types. I've added a comment documenting the possible simplification in the future. In the mean time I'm going to see if I can get myself added to the Typescript contributors list :)

packages/automerge-doc-server/src/server.ts
jmoggr: Yes, I made the change. I had tunnel vision on the automerge library itself.

jmoggr:
> Also, you can rebase now that https://github.com/ToposInstitute/CatColab/pull/867 is merged.

Done!

PR #870, opened 2025-11-25
Minor fixes to dev docs

kasbah:


jmoggr:
Base branch changed considerably and is closed. We're taking a simpler approach to docs and scripts for now.

PR #874, opened 2025-12-01
Upgrade Biome and fix method of extending Biome config files

kasbah:


jmoggr:
Biome fails to load the relative paths for extending files when run in directories other than the one a `biome.json` is in. This was causing auto formatting in my editor to fail, and is likely a bug in biome. This PR fixes that issue.

Thanks Kaspar for actually looking at the docs! That syntax is interesting...

jmoggr:
Do we need to upgrade biome? All things being equal I'd prefer to not do that because it makes life a bit harder in Nix. See: https://github.com/ToposInstitute/CatColab/pull/867

If we need to upgrade often it would not be unreasonable for me to fix the problem on the Nix side.

kasbah:
We don't necessarily need to update it but should be using the same version everywhere and the `$schema` version should be the specified version. That's the main idea of the commit but using the latest everywhere makes sense if we are changing the version in some places anyway. Also, this is a patch-level increment in sem-ver i.e. +0.0.1 so we generally want to pull those fixes when we can.

I am guessing there is really no better workflow for these deps in nix?

epatters:
Where are we landing on this one? Should we do a fix to the biome config file without bumping the biome version?

jmoggr:
> Should we do a fix to the biome config file without bumping the biome version?

I've bumped the biome version to 2.3.8 so that the nix and npm versions agree again. This is ready to merge.

---

Regarding the biome/nix problem:

Getting the biome npm package to run on NixOS is possible, but non-trivial. I'm willing to just deal with the mismatched versions on my end instead of letting my usage of NixOS affect the repo. 

At the same time I don't see the value in frequently updating biome. Even without my NixOS problem it would still be painful: https://github.com/ToposInstitute/CatColab/pull/874/commits/dc67430e11fd57af1ae26ab746beac508d0a1b15. Even without anything nix related it is still moderately annoying.

biome.json
```diff
@@ -24,9 +24,18 @@
...
+                                "!catcolab-api",
+                                "!catlog-wasm"
+                            ],
                             ":BLANK_LINE:",
-                            ["catcolab-ui-components", "catcolab-api", "catlog-wasm"],
+                            [
+                                "catcolab-ui-components",
+                                "catcolab-api",
+                                "catlog-wasm"
+                            ],
```
kasbah: I don't think that's how biome itself would auto-format this file.

PR #875, opened 2025-12-01
Remove `automerge-doc-server` and sync Automerge docs from `backend` using `samod`

jmoggr:


packages/backend/src/app.rs
```diff
@@ -1,9 +1,11 @@
 use firebase_auth::FirebaseUser;
 use serde::Serialize;
-use socketioxide::SocketIo;
+// use socketioxide::SocketIo;
```
kasbah:
```suggestion
```

Needs to be removed from Cargo.toml too.

jmoggr: Fixed!

packages/backend/src/automerge_json.rs
```diff
@@ -0,0 +1,202 @@
...
+use crate::document::{RefContent, autosave};
+use automerge::hydrate;
+use automerge::transaction::Transactable;
+use futures_util::stream::StreamExt;
+use samod::DocHandle;
+use serde_json::Value;
+use uuid::Uuid;
+
+/// Insert a JSON value into a map property
+fn insert_value_into_map<'a>(
```
epatters, line 13: You might have mentioned these helper functions to me before, Jason, but I'm surprised that we have to write them. Is this functionality that you would expect to be upstreamed?

packages/backend/src/document.rs
```diff
@@ -17,15 +18,35 @@ pub async fn new_ref(ctx: AppCtx, content: Value) -> Result<Uuid, AppError> {
...
+        })
+        .map_err(|e| AppError::Invalid(format!("Failed to populate document: {:?}", e)))?;
+
+    // Insert the new automerge document into automerge-repo
+    let doc_handle = ctx
+        .state
+        .repo
+        .create(automerge_doc)
+        .await
+        .map_err(|e| AppError::Invalid(format!("Failed to create document: {:?}", e)))?;
```
epatters: The docstring for the error variant `Invalid` says "Client made request with invalid data". I'm not sure under what conditions adding a doc to the repo can fail, but I'd be surprised if it fit the above description. Perhaps we should have a new error variant for unexpected errors from Automerge like we do for unexpected errors from the SQL DB.

packages/backend/src/document.rs
```diff
@@ -17,15 +18,35 @@ pub async fn new_ref(ctx: AppCtx, content: Value) -> Result<Uuid, AppError> {
...
+
+    let doc_id = doc_handle.document_id().to_string();
 
-    let mut transaction = ctx.state.db.begin().await?;
+    // let doc_content = doc_handle.with_document(|doc| doc.clone());
```
epatters: Lingering commented code.

packages/backend/src/document.rs
```diff
@@ -65,25 +92,25 @@ pub async fn head_snapshot(state: AppState, ref_id: Uuid) -> Result<Value, AppEr
...
-    let doc_id = query.fetch_one(&state.db).await?.doc_id;
+    let doc_id = get_doc_id(state.clone(), ref_id).await?;
 
-    let export_response = export_automerge_doc(&state.automerge_io, doc_id).await?;
+    let doc_handle = state
+        .repo
+        .find(doc_id)
+        .await
+        .map_err(|e| AppError::Invalid(format!("Failed to find document: {:?}", e)))?
+        .ok_or_else(|| AppError::Invalid("Document not found".to_string()))?;
```
epatters, line 99: Ditto about these errors and others below. Is this the appropriate variant?

packages/migrator/src/migrations/m20250805230408_fix_automerge_storage.rs
epatters, line 1: Any reason to not just delete this file/migration entirely?

packages/backend/Cargo.toml
```diff
@@ -14,17 +15,21 @@ migrator = { version = "0.1.0", path = "../migrator" }
...
   "rustls",
 ] }
+futures-util = "0.3"
 http = "1.1.0"
 hyper = { version = "1.4.1", features = ["server"] }
 jsonrpsee = "0.24.6"
 jsonrpsee-server = "0.24.6"
 notebook-types = { version = "0.1.0", path = "../notebook-types" }
-qubit = "0.10.2"
+qubit = { version = "1.0.0-beta.0", features = ["ts-serde-json", "ts-uuid"] }
```
epatters, line 24: Thanks for taking the opportunity to upgrade this and other crates. Looking at the [release notes](https://github.com/andogq/qubit/releases/tag/qubit-v1.0.0-beta.0) for qubit, should we replace `#[derive(TS)]` with `#[qubit::ts]`? See: https://github.com/andogq/qubit/commit/bf34c4c7f99edf8d0fc7d6d9c0347ff95b7e488d

packages/backend/src/automerge_json.rs
jmoggr: I was only a bit surprised. It seems like not many people are needing to manipulate Automerge docs both from Javascript and Rust, and practically doing that has only recently been made possible with Samod.

We could be the first people to run into this, I'll open an issue on Samod. It seems reasonable to have this functionality upstreamed.

packages/backend/src/document.rs
jmoggr: Removed

packages/migrator/src/migrations/m20250805230408_fix_automerge_storage.rs
jmoggr: We should absolutely not be doing this as a standard practice, but in this case it would be harmless, not unreasonable (in this case), and easy to delete the file and drop the relevant line from the migrations table.

I would personally keep it because I view the cost of maintaining it to be lower than the cost of modifying history. However this is mostly a matter of taste so it's ultimately your call.

---

Deleting it would mean that migrations listed in the migrations table (`_sqlx_migrator_migrations`) won't match the migrations in code, which I think is bad since the two migration lists matching is generally expected. We could delete the entry from the migrations table manually, but I typically take a principled stance against modifying the migrations table ourselves. Also the maintenance overhead of the empty migration is effectively zero.

In theory the migrations table should be treated as an append only/immutable log, and even modifying any migration code after applying it breaks that contract. In this case the burden of maintaining the migration code was unreasonably high and we can be highly confident that all existing databases that matter already have this migration applied. But **restoring from a backup that was made before this migration was applied will fail**. I think this is a reasonable price (we briefly discussed this in a meeting, but I should have made that more clear in the PR).

Other migration tools will make a huge fuss when the migration histories don't match, however sqlx does not (which is not what I expected when I started writing this reply!).

IMO the migrations should be viewed similarly to the the versions in the `notebook-types` package, and we wouldn't delete a legacy version.

packages/backend/Cargo.toml
jmoggr: I feel conflicted about this. I have a preference towards being explicit that we are using `ts-rs`, since those are the docs that I would look at for serialization issues. On the other hand the docs do say we should upgrade, and if I were to start using qubit today I would probably be using `#[quit::ts]`.

The change is easy and I suspect that we will do it, but I'll wait for your yay/nay on this.

---

Note: The qubit upgrade was necessary because a tower upgrade was necessary because a axum upgrade was necessary to get some element of the websocket handler types to work. The ~100 lines of trivial looking changes to `main.rs` took about half the dev time...

packages/backend/src/document.rs
jmoggr: That is correct, I made the change.

packages/backend/src/document.rs
jmoggr: nope, changed.

packages/automerge-doc-server/README.md
```diff
@@ -1,67 +0,0 @@
...
-
-#### Update the hash
-Replace pkgs.lib.fakeHash with the actual hash shown in the error message:
-```
-hash = "sha256-tIgtzlslvm2A1UpwfVsYk3E9HkKJntu36gEtsFjswgo=";
-```
-
-Now the dependencies are correctly tracked, and the build should proceed normally. Make sure to include the updated hash in a commit.
-
-Yes, this is the official way of solving this in nix. No, there is no way to automate this with nix. Nix: where developer experience goes to die.
```
kasbah, line 58: We need to document this somewhere else because this is the doc I currently refer to and send people to for this.

jmoggr: I've rewritten the docs and included them here: https://github.com/ToposInstitute/CatColab/pull/943

packages/migrator/src/migrations/m20250805230408_fix_automerge_storage.rs
epatters: Thanks for the detailed explanation. I agree with you: we should just keep the migration.

packages/backend/Cargo.toml
epatters: Yeah, I confess I don't understand why upstream made the change, but since that's now recommendation and the usage in their examples, I think we should just go with it.

packages/backend/src/document.rs
epatters: If I'm not mistaken, many/most of the `AppError::Invalid` uses have not yet been converted to the new `AppError::AutomergeRepo`.

jmoggr: It is still used for errors that result from invalid user input. To me `AppError::AutomergeRepo` is a relatively scary error: it indicates that something went wrong with AutomergeRepo/Samod. We could use it in cases where invalid input is meant to interact with AutomergeRepo, but that seems less precise.

I'm happy to make the change, but I'd like to make sure that I'm not misunderstanding something.

I believe this can only fail if `populate_automerge_from_json(tx, automerge::ROOT, &content)?` fails, and that can only fail if there is invalid input.
1 use
```
   automerge_doc
        .transact(|tx| {
            populate_automerge_from_json(tx, automerge::ROOT, &content)?;
            Ok::<_, automerge::AutomergeError>(())
        })
        .map_err(|e| AppError::Invalid(format!("Failed to populate document: {:?}", e)))?;
```

If there is an samod specific error `.await?` will return it. If the document was not found then the input `doc_id` was invalid, since it doesn't exist.
3 uses
```
    let doc_handle = state
        .repo
        .find(doc_id)
        .await?
        .ok_or_else(|| AppError::Invalid("Document not found".to_string()))?;
```

Failure to parse user input to `samod::DocumentId`, I think that can only happen if the input is invalid.
1 use
```
    let doc_id: samod::DocumentId = doc_id
        .parse()
        .map_err(|_| AppError::Invalid("Invalid document ID".to_string()))?;
```

packages/backend/Cargo.toml
jmoggr: Okay. Since this is a bulk change I'll do this as a separate PR. I created a tracking issue here: https://github.com/ToposInstitute/CatColab/issues/948

packages/backend/src/document.rs
epatters: Thanks for clarifying, that makes sense!

epatters (approved):

PR #880, opened 2025-12-03
Keep parent document on the left in paned view

kasbah:
When trying out some things it just annoying me too much that the viz would unexpectedly open on the left. I finally snapped and fixed it.

### DIFF
--- a/packages/frontend/src/page/document_page_sidebar.tsx
+++ b/packages/frontend/src/page/document_page_sidebar.tsx
@@ -2,7 +2,7 @@ import { useNavigate } from "@solidjs/router";
 import { createMemo, createResource, For, Show } from "solid-js";
 import invariant from "tiny-invariant";
 
-import type { Link } from "catlog-wasm";
+import type { Document, Link } from "catlog-wasm";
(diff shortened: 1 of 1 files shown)


epatters (changes requested):
Thanks Kaspar, I agree that this was very annoying!

I think we should generalize this logic so that when two documents in a parent-child relation are being displayed, the parent document always shows up on the left. To do that, we could factor out a new function

```typescript
function getDocParent(doc: Document): StableRef | undefined
```

from [`getLiveDocRoot`](https://github.com/ToposInstitute/CatColab/blob/7de4854e8206ecb77cd65ec32cfdcc9a7a048b39/packages/frontend/src/page/document_page_sidebar.tsx#L30) and then replace your current conditional with a check of whether the clicked doc is a parent of the primary doc.

kasbah:
Thanks, that's more correct (though it does add an API call). I've changed it.

epatters (approved):
Thanks! I think the extra document load is fine because the parent document is going to get loaded anyway and the `Api` object will cache the document handle.

PR #883, opened 2025-12-04
Fix 'DocHandle is not ready' error on readonly documents

### LINKED ISSUE #882: Example models on production not loading
None of these are loading: https://catcolab.org/help/guides/example-models

On console:

```
Error: Failed to retrieve document: Automerge server error: DocHandle is not ready
    fetchDocCacheEntry types.ts:112
error_boundary.tsx:8:13
```

jmoggr:
closes #882 

see code comment for explanation of fix.

There was a path dependency issue with automerge repo that we missed: we had never tested viewing a read-only document that we had not previously been connected to.

The chance of us catching this with better testing was quite low, and the automerge-repo docs do not indicate that this could be an issue with `repo.export`. More "correct" support for readonly documents in samod should prevent similar errors in the future (all documents would have the same access pattern, reducing the possibility of path-dependent bugs).

### DIFF
--- a/packages/automerge-doc-server/src/server.ts
+++ b/packages/automerge-doc-server/src/server.ts
@@ -168,6 +168,14 @@ export class AutomergeServer implements SocketIOHandlers {
     }
 
     async exportDoc(docId: string): Promise<ExportDocSocketResponse> {
+        // repo.find ensures that the doc handle is available when repo.export is called below. This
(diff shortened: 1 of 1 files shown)


epatters (approved):
Thanks for the fix and for clearly documenting it!

epatters:
Issue reported upstream too: https://github.com/automerge/automerge-repo/issues/536

PR #887, opened 2025-12-08
Move warning banner into `ui-components`

kasbah:
Also started using css modules for the components.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -80,6 +80,9 @@
         },
         "linter": {
             "enabled": true
+        },
+        "parser": {
+            "cssModules": true
         }
     }
 }
--- a/packages/frontend/src/page/document_page.css
+++ b/packages/frontend/src/page/document_page.css
@@ -1,34 +1,3 @@
-.warning-banner {
-    display: flex;
-    align-items: center;
-    flex-wrap: wrap;
-    gap: 0.75rem;
-    background-color: var(--warning-bg, #fff3cd);
-    border: 1px solid var(--warning-border, #ffc107);
-    color: var(--warning-text, #856404);
-    padding: 1rem 1.25rem;
-    margin-bottom: 1rem;
-    border-radius: 0.375rem;
-    font-size: 0.9375rem;
-    line-height: 1.5;
-}
-
-.warning-banner > .lucide-icon {
-    flex-shrink: 0;
-    min-width: 20px;
-}
-
-.warning-banner-content {
-    flex: 1;
-    min-width: 200px;
-}
-
-.warning-banner .button {
-    margin: 0;
-    flex-shrink: 0;
-    white-space: nowrap;
-}
-
 .document-head {
     display: flex;
     flex-direction: row;
--- a/packages/frontend/src/page/document_page.tsx
+++ b/packages/frontend/src/page/document_page.tsx
@@ -3,7 +3,6 @@ import { useNavigate, useParams } from "@solidjs/router";
 import ChevronsRight from "lucide-solid/icons/chevrons-right";
 import Maximize2 from "lucide-solid/icons/maximize-2";
 import RotateCcw from "lucide-solid/icons/rotate-ccw";
-import TriangleAlert from "lucide-solid/icons/triangle-alert";
 import {
     createEffect,
     createResource,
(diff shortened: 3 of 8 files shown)


epatters (commented):
Thanks Kaspar!

I like the use of CSS modules. In the `frontend` package, I was experimenting with using CSS modules for the styling arrows and visualizations and using [`typed-css-modules`](https://github.com/Quramy/typed-css-modules) to generate TypeScript type defs for them. See the `compile` command in `package.json` and e.g. [`svg_styles.module.css.d.ts`](https://github.com/ToposInstitute/CatColab/blob/1c175dfe5c8ccd05cc6827759bea14daacd02cf2/packages/frontend/src/stdlib/svg_styles.module.css.d.ts).

Two questions:

1. Should we be generating such type defs throughout the codebase?
2. If so, should this be checked and enforced in CI?

My guesses to both questions are "yes" but I'm opening to hearing other opinions.

kasbah:
I didn't think about the possibility of generating the types actually. I noticed there is a [Vite plugin for CSS modules](https://github.com/privatenumber/vite-css-modules), did you ever encounter any of the issues it patches? It also generates the types and looking at the source, generating the types is not very complicated, I extracted it into it's own plugin:

<details>
<summary>vite-plugin-css-module-types.ts</summary>

```ts
import { readFile, writeFile } from 'node:fs/promises';
import type { Plugin } from 'vite';

/**
 * Minimal Vite plugin that generates .d.ts files for CSS modulestha
 * 
 * Usage:
 * ```ts
 * import { cssModuleTypesPlugin } from './minimal-types-plugin';
 * 
 * export default {
 *   plugins: [cssModuleTypesPlugin()]
 * }
 * ```
 */
export function cssModuleTypesPlugin(): Plugin {
  const cssModuleRE = /\.module\.(css|scss|sass|less|styl|stylus)$/;

  return {
    name: 'css-module-types',
    
    // Run after vite:css to access the processed CSS modules
    enforce: 'post',
    
    async transform(code, id) {
      // Only process CSS module files
      if (!cssModuleRE.test(id)) {
        return null;
      }

      // Remove query params (like ?used)
      const filePath = id.split('?')[0];
      
      try {
        // Extract exported class names from the generated code
        // The code will look like: export { foo, bar }; export default { foo, bar };
        const exportMatches = code.match(/export\s*\{([^}]+)\}/);
        
        if (!exportMatches) {
          return null;
        }

        // Parse exported names
        const exports = exportMatches[1]
          .split(',')
          .map(s => s.trim())
          .filter(s => s && s !== 'default')
          .map(s => {
            // Handle "name as alias"
            const parts = s.split(/\s+as\s+/);
            return parts[parts.length - 1].replace(/['"]/g, '');
          });

        // Generate .d.ts content
        const dtsContent = generateDtsContent(exports);
        
        // Write .d.ts file
        const dtsPath = `${filePath}.d.ts`;
        
        // Only write if content changed
        const existingContent = await readFile(dtsPath, 'utf8').catch(() => null);
        if (existingContent !== dtsContent) {
          await writeFile(dtsPath, dtsContent);
          console.log(`[css-module-types] Generated ${dtsPath}`);
        }
      } catch (error) {
        console.error(`[css-module-types] Error generating types for ${filePath}:`, error);
      }

      return null; // Don't modify the code
    }
  };
}

function generateDtsContent(classNames: string[]): string {
  if (classNames.length === 0) {
    return `/* eslint-disable */
// Generated by css-module-types plugin
declare const styles: {};
export default styles;
`;
  }

  const declarations = classNames
    .map(name => `  readonly ${name}: string;`)
    .join('\n');

  return `/* eslint-disable */
// Generated by css-module-types plugin
declare const styles: {
${declarations}
};
export default styles;
export const ${classNames.join(', ')}: string;
`;
}
```

</details>

The workflow around it is still a little awkward because the dev server will only generate the types once you request the associated .module.css through vite, but at least `vite build` will always generate the types. However we tend to `tsc` before `vite build` anyway.

I'm not sure what to go for here. Maybe using the `typed-css-modules` cli as a separate step is clearer.

epatters:
Ah, thanks for noticing that. I didn't realize that Vite had a CSS modules plugin that would generate type defs. Since we're already using Vite, it seem worth trying out the Vite plugin, but no strong opinion on which to use from me.

kasbah:
I'll work on that on another PR. I think this can be merged as is.

PR #888, opened 2025-12-08
Name existing colors as CSS custom properties 

kasbah:
This rationalizes our color use through CSS custom properties defined globally, with some slight cheating here and there where our colors really didn't make sense or were not what was intended. I also renamed "primary" to "positive" for our buttons so I could name the colors that right away.

I created a `colors.stories.tsx` file to list and study the existing usage of these colors: a palette that tries to make some sort of sense of the colors. I'm sure I failed in some respects -- I also gave up on the finding better names for the remaining uses of `--color-gray-*`. 

I searched for / generated a bunch of examples of actual usage of these colors from our code to try and better understand where and how they are used. These will go out of date at some point. We can delete them when that happens.

https://branch-888-merge--catcolab.netlify.app/dev/ui-components/?path=/docs/colors--docs

### DIFF
--- a/packages/ui-components/src/colors.stories.tsx
+++ b/packages/ui-components/src/colors.stories.tsx
@@ -1,0 +1,4283 @@
+import Resizable from "@corvu/resizable";
+import EllipsisVertical from "lucide-solid/icons/ellipsis-vertical";
+import FilePlus from "lucide-solid/icons/file-plus";
+import GripVertical from "lucide-solid/icons/grip-vertical";
+import Menu from "lucide-solid/icons/menu";
+import Plus from "lucide-solid/icons/plus";
+import RotateCcw from "lucide-solid/icons/rotate-ccw";
+import X from "lucide-solid/icons/x";
+import { createSignal, For, onMount } from "solid-js";
+import type { StoryObj } from "storybook-solidjs-vite";
+
+import { ErrorAlert, Note, Question, Warning } from "./alert";
+import { Button } from "./button";
+import {
+    type BooleanColumnSchema,
+    FixedTableEditor,
+    type TextColumnSchema,
+} from "./fixed_table_editor";
+import { FormGroup, SelectField, TextInputField } from "./form";
+import { IconButton } from "./icon_button";
+import { InlineInput } from "./inline_input";
+import { ResizableHandle } from "./resizable";
+import { Spinner } from "./spinner";
+import { WarningBanner } from "./warning_banner";
+import "./colors.css";
+
+interface ColorSwatchProps {
+    value: string;
+}
+
+function ColorSwatch(props: ColorSwatchProps) {
+    let swatchRef: HTMLDivElement | undefined;
+    const [computedColor, setComputedColor] = createSignal("");
+
+    onMount(() => {
+        if (swatchRef) {
+            const computedStyle = window.getComputedStyle(swatchRef);
+            const color = computedStyle.getPropertyValue("background-color");
+            setComputedColor(color);
+        }
+    });
+
+    return (
+        <div
+            style={{
+                display: "flex",
+                "flex-direction": "column",
+                gap: "0.5rem",
+                "min-width": "250px",
+            }}
+        >
+            <div
+                style={{
+                    "font-family": "var(--mono-font)",
+                    "font-size": "0.70rem",
+                    "font-weight": "600",
+                    color: "var(--color-gray-800)",
+                }}
+            >
+                {props.value}
+            </div>
+            <div
+                ref={swatchRef}
+                style={{
+                    width: "100%",
+                    height: "80px",
+                    background: `var(${props.value})`,
+                    border: "1px solid var(--color-gray-550)",
+                    "border-radius": "8px",
+                    "box-shadow": "0 2px 4px var(--color-overlay-light)",
+                }}
+            />
+            <div
+                style={{
+                    "font-family": "var(--mono-font)",
+                    "font-size": "0.75rem",
+                    color: "var(--color-gray-800)",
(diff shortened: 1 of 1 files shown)


packages/frontend/src/help/help_layout.css
```diff
@@ -84,12 +84,12 @@
...
     & a {
-        color: var(--topos-secondary);
+        color: var(--color-topos-secondary);
         text-decoration: underline;
 
         &:hover {
-            /*color: color-mix( in oklab, var(--topos-secondary) 80%, black);*/
-            color: var(--topos-primary);
+            /*color: color-mix( in oklab, var(--color-topos-secondary) 80%, black);*/
+            color: var(--color-topos-primary);
```
kasbah:
```suggestion
            color: var(--color-topos-primary);
```

packages/frontend/src/notebook/notebook_cell.css
```diff
@@ -38,15 +38,15 @@
...
 }
 
 .drop-indicator-with-dots::before {
     content: "";
     position: absolute;
     width: 6px;
     height: 6px;
-    background-color: #6b7280;
+    background-color: var(--color-dnd-insert);
     /* border-radius: 50%; */
```
kasbah:
```suggestion
    background-color: var(--color-dnd-insert);
```

packages/frontend/src/help/help_layout.css
```diff
@@ -29,15 +29,15 @@
 
             & > li.help-navigation-link a {
                 text-decoration: none;
-                color: var(--topos-secondary);
+                color: var(--color-topos-secondary);
                 font-weight: 600;
                 padding: 0.3em 0.5em;
-                /*border: 1px solid var(--topos-secondary);*/
+                /*border: 1px solid var(--color-topos-secondary);*/
```
kasbah:
```suggestion
```

packages/frontend/src/stdlib/styles.module.css
```diff
@@ -1,16 +1,20 @@
...
-        conic-gradient(at var(--corner) var(--corner), #0000 75%, #000 0) 0 0 /
-        calc(100% - var(--corner)) calc(100% - var(--corner)),
-        linear-gradient(#000 0 0) content-box;
+        conic-gradient(
+            at var(--corner) var(--corner),
+            var(--color-svg-mask-gradient-transparent) 75%,
+            var(--color-svg-mask-gradient-start) 0
+        )
+        0 0 / calc(100% - var(--corner)) calc(100% - var(--corner)),
+        linear-gradient(var(--color-svg-mask-gradient-start) 0 0) content-box;
```
kasbah:
```suggestion
        conic-gradient(at var(--corner) var(--corner), #0000 75%, #000 0) 0 0 /
        calc(100% - var(--corner)) calc(100% - var(--corner)),
        linear-gradient(#000 0 0) content-box;
```

packages/ui-components/src/colors.css
```diff
@@ -0,0 +1,160 @@
...
+    /* ========================================
+     * Brand Colors
+     * ======================================== */
+    --color-topos-primary: rgb(83, 182, 178);
+    --color-topos-primary-hover: #45a19d;
+    --color-topos-secondary: rgb(128, 49, 78);
+
+    /* ========================================
+     * Neutral Colors - Grays
+     * From lightest to darkest
```
epatters, line 23: What's the thinking behind having this big list of shades of gray?

- Is it temporary?
- If not, are these variables intended to be used directly in components, or just in this CSS file to create more meaningful names, e.g., `color-text-secondary`?

packages/ui-components/src/colors.css
```diff
@@ -0,0 +1,160 @@
...
+
+    /* ========================================
+     * Hover Colors
+     * Background colors for hover states
+     * ======================================== */
+    --color-hover-bg: var(--color-gray-150);
+    --color-hover-bg-light: var(--color-gray-100);
+    --color-hover-bg-medium: var(--color-gray-250);
+    --color-hover-bg-dark: var(--color-gray-300);
+    --color-hover-bg-strong: var(--color-gray-500);
```
epatters, line 55: Related question: do we intend to have so many options for hover backgrounds or is this just a first step capturing our existing usage?

epatters (approved):
Thanks Kaspar for doing this patient work to start rationalizing our CSS and color schemes!

I'm going to merge this as-is since it's a big diff and already an improvement on our current situations, but I do have a few questions below about where you're heading with this work.

packages/ui-components/src/colors.css
kasbah: These are good questions and I am not sure right now. The main reason I made them is that I was too lazy to give every single use of gray a unique name (but I did do that for some). Having a graded palette like this is not an unreasonable thing to have in a color system (though we should have _less_ variants of gray). 

It's a trade-off between rigidity and flexibility. Maybe the clearest solution is to have a grading like this for _all_ the colors we want to use, so for red, blue green etc. and then a semantic layer with where they are aliased with their purpose.

packages/ui-components/src/colors.css
kasbah: This was just capturing usage and we should reduce it.

packages/ui-components/src/colors.css
epatters: Thanks for the context. In view of the discussion here and in today's devs meeting, and since it sounds like this will be a long-running project, perhaps you can create an issue to track any planned follow-on to this PR?

kasbah: https://github.com/ToposInstitute/CatColab/issues/912

PR #895, opened 2025-12-10
Change object/morphism types in model notebooks

### LINKED ISSUE #184: Popup menu to change object/morphism type

quffaro:
Closes #184.

### DIFF
--- a/packages/frontend/src/model/model_editor.tsx
+++ b/packages/frontend/src/model/model_editor.tsx
@@ -27,6 +27,9 @@ export function ModelNotebookEditor(props: { liveModel: LiveModelDoc }) {
     const liveDoc = () => props.liveModel.liveDoc;
 
     const cellConstructors = () => {
+	    // const cells = liveDoc().doc.notebook.cellOrder;
+		// console.log(cells);
+
         const theory = props.liveModel.theory();
         return theory ? modelCellConstructors(theory) : [];
     };
@@ -54,6 +57,7 @@ export function ModelCellEditor(props: FormalCellEditorProps<ModelJudgment>) {
     const liveModel = useContext(LiveModelContext);
     invariant(liveModel, "Live model should be provided as context");
 
+    console.log(props);
     return (
         <Switch>
             <Match when={props.content.tag === "object" && liveModel().theory()}>
@@ -106,6 +110,7 @@ function modelCellConstructors(theory: Theory): CellConstructor<ModelJudgment>[]
             },
         });
     }
+
     for (const meta of theory.modelTypes ?? []) {
         constructors.push(modelCellConstructor(meta));
     }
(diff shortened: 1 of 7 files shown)


packages/frontend/src/model/model_editor.tsx
```diff
@@ -94,9 +94,13 @@ export function ModelCellEditor(props: FormalCellEditorProps<ModelJudgment>) {
     );
 }
 
-function modelCellConstructors(theory: Theory): CellConstructor<ModelJudgment>[] {
+function modelCellConstructors(
```
epatters, line 97: I don't think it makes sense to implement this feature by changing `modelCellConstructors`. As the name suggests, "cell constructors" are functions that construct new notebook cells. This is quite different in scoping to mutating existing cells of a given type.

packages/frontend/src/notebook/notebook_cell.tsx
```diff
@@ -108,6 +109,8 @@ export function NotebookCell(props: {
     tag?: string;
     currentDropTarget: string | null;
     setCurrentDropTarget: (cellId: string | null) => void;
+    isActive: boolean;
```
epatters, line 112: Why is this prop added? I don't see it being used in any new places.

packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -139,14 +139,56 @@ export function NotebookEditor<T>(props: {
...
+    };
+
+    const isMorType = (content: unknown): content is { tag: "morphism"; morType: MorType } => {
+        return (
+            typeof content === "object" &&
+            content !== null &&
+            "tag" in content &&
+            content.tag === "morphism"
+        );
+    };
```
epatters, line 162: A hint that this design is off is that notebook editor is not supposed to know *anything* about theories, models, etc. It's a completely generic component.

epatters (changes requested):
Thanks Matt for taking this on!

I have one high-level comment about the architecture, reflected in several of the comments below. Rather than using the cell constructor mechanism to construct a whole new cell and then extract its object/morphism type, we should, in each object declaration cell, create a cell switcher by looking up the available object types directly from the theory, and similarly for the morphism declaration cells. No code in the `notebook` directory should make reference to theories or models.

epatters:
Also wondering about this PR, @quffaro. If you're still planning to work on this but need help with the design, Kaspar could likely help.

quffaro:
I'll reach out to @kasbah tomorrow to help me discuss the design so I can finish it. I want the feature but it fell off my radar

epatters:
Ok, cool. It's been a while since I thought about this, but IIRC, the strategy I had in mind was to generalize the notebook to allow the cell "labels" to be arbitrary components, rather than just strings, then have the model editor pass in object/morphism type selectors as appropriate.

PR #897, opened 2025-12-11
Purge kobalte by re-implementing alert

### LINKED ISSUE #848: Remove Kobalte from frontend dependencies
As of #847, the only remaining use of `kobalte` is in the alert components, so this should be a small job and will eliminate a substantial dependency.

FWIW, I think we should keep using `corvu` for the time being since it's more pick-and-choose and the components seem high quality.

kasbah:
closes #848

### DIFF
--- a/packages/ui-components/src/alert.tsx
+++ b/packages/ui-components/src/alert.tsx
@@ -16,7 +16,7 @@ export type AlertProps = {
 
 /** A warning alert. */
 export const Warning = (props: AlertProps) => (
-    <div role="alert" class="alert alert-warning">
+    <div class="alert alert-warning">
         <div class="alert-heading">
             <TriangleAlert />
             {props.title ?? "Warning"}
@@ -29,7 +29,7 @@ export const Warning = (props: AlertProps) => (
 Not called `Error` to avoid shadowing that name in JavaScript.
  */
 export const ErrorAlert = (props: AlertProps) => (
-    <div role="alert" class="alert alert-error">
+    <div class="alert alert-error">
         <div class="alert-heading">
             <OctagonX />
             {props.title ?? "Error"}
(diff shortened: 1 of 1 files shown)


packages/ui-components/src/alert.tsx
```diff
@@ -17,46 +16,46 @@ export type AlertProps = {
 
 /** A warning alert. */
 export const Warning = (props: AlertProps) => (
-    <Alert class="alert alert-warning">
+    <div class="alert alert-warning">
```
kasbah, line 19: Adding `role="alert"` would strictly re-implement what we had but I think we should get rid of it.

> The alert role is used to communicate an important and usually time-sensitive message to the user. When this role is added to an element, the browser will send out an accessible alert event to assistive technology products which can then notify the user.
> 
> The alert role should only be used for information that requires the user's immediate attention, for example:
> 
>     An invalid value was entered into a form field
>     The user's login session is about to expire
>     The connection to the server was lost so local changes will not be saved
> 
> [...]
> 
> Warning: Because of its intrusive nature, the alert role must be used sparingly and only in situations where the user's immediate attention is required.
> 


https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role

epatters: I agree, our alerts are not so important as to require the user's "immediate attention."

kasbah: I noticed one use of it in the permissions dialog that did seem like it needed immediate attention. We should use something else there. Possibly, confusingly our `WarningBanner` should have `role=alert` and should be used there.

epatters: In that case, should our alert components have an optional prop for whether to set `role=alert`, with the default being to not set it?

PR #899, opened 2025-12-11
Cache builds in GitHub actions

kasbah:
This reduces time to deploy our previews by only building what is needed. These builds tend to take 1-2 minutes but I have also  noticed frequent network related intermittent failures with e.g. Math docs that this will avoid on unrelated PRs.

Downside is we need to keep the cache key computation for frontend updated when changing any in-repo dependencies but I think it's worth it.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -112,7 +112,7 @@ jobs:
         id: cache-key
         run: |
           # Get git hash of the most recent commit affecting frontend or any of its dependencies
-          hash=$(git log -1 --format="%H" -- packages/frontend packages/backend/pkg packages/catlog packages/catlog-wasm packages/notebook-types packages/ui-components Cargo.toml)
+          hash=$(git log -1 --format="%H" -- packages/frontend packages/backend/pkg packages/catlog packages/catlog-wasm packages/notebook-types packages/ui-components Cargo.toml Cargo.lock)
           echo "git-hash=${hash}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


kasbah:
image
:arrow_right: 
image

when cached

.github/workflows/deploy.yml
```diff
@@ -76,24 +104,48 @@ jobs:
...
         uses: actions/checkout@v4
+        with:
+          # Need full history for `git log` for cache key
+          fetch-depth: 0 
+
+      - name: Compute cache key
+        id: cache-key
+        run: |
+          # Get git hash of the most recent commit affecting frontend or any of its dependencies
+          hash=$(git log -1 --format="%H" -- packages/frontend packages/catlog-wasm packages/catlog packages/backend/pkg packages/ui-components)
```
kasbah: Notebook types?

epatters: Yes, `notebook-types` too: `frontend` doesn't depend on it directly but does indirectly through `catlog-wasm`.

epatters: Should this depend on the top-level `Cargo.lock` too?

epatters (changes requested):
Thanks Kaspar! I'm a little nervous about the cache invalidation but I've also been annoyed by the intermittent failures in GitHub actions, especially when building unrelated artifacts, so I'm happy to give this a try.

.github/workflows/deploy.yml
kasbah: I added both.

PR #900, opened 2025-12-11
Convert form styles to CSS module

kasbah:


kasbah:
This causes missing styles on the permissions form.

image

PR #905, opened 2025-12-12
Stock-flow diagrams with signed links

tim-at-topos:
Enriching the theory of stock-flow diagrams to have positive and negative links, instead of merely links. Includes a mass-action simulation analysis.


tim-at-topos:
@epatters I'm trying to get + and - to appear on positive / negative links in the visualisation, and it's proving very frustrating. What's currently confusing me even more is the file `svg_styles.module.css`, which seems to tell me that links should be blue and rectangles should be a delightful ube colour, but this isn't what's happening in reality. Any suggestions?

packages/backend/pkg/src/index.ts
```diff
@@ -21,8 +21,8 @@ import type { NewPermissions } from "./NewPermissions.ts";
 import type { UserSummary } from "./UserSummary.ts";
 import type { UsernameStatus } from "./UsernameStatus.ts";
 import type { UserProfile } from "./UserProfile.ts";
-import type { Paginated } from "./Paginated.ts";
 import type { RefStub } from "./RefStub.ts";
+import type { Paginated } from "./Paginated.ts";
```
epatters, line 24: It's annoying that (1) the order of these lines seems unstable across builds and (2) our formatter isn't normalizing the order, but those are problems for another time. For now, can you just avoid causing a diff to this unrelated file in this PR?

packages/catlog/src/stdlib/theories.rs
```diff
@@ -119,6 +119,27 @@ pub fn th_category_links() -> DiscreteTabTheory {
...
+/// The theory of categories with signed links.
+///
+/// It can be useful to consider a version of stock and flow diagrams where the
+/// links are labelled with a sign: positive or negative
+pub fn th_category_signed_links() -> DiscreteTabTheory {
+    let mut th = DiscreteTabTheory::new();
+    th.add_ob_type(name("Object"));
+    let ob_type = TabObType::Basic(name("Object"));
+    th.add_mor_type(
+        name("PositiveLink"),
```
epatters: I'd consider naming this `Link` instead of the more descriptive `PositiveLink`. These choices *shouldn't* actually matter (and I like your instinct to be descriptive), but since we don't yet have migrations between models of tabulator theories except the trivial ones (inclusions), my suggestion has the advantage of allowing a trivial migration from models of `th_category_links` with the right interpretation.

epatters (changes requested):
Thanks Tim, this looks great so far!

My only major request is that we create a *new* theory in the frontend rather than replacing the old one. There are at least three reasons for this:

1. It won't break any existing models that use the old theory
2. More importantly, it's consistent with our philosophy of having a family of languages of increasing expressiveness and complexity, with migrations between them, like we've done with CLDs. (As we've discussed, the UX for these migrations currently isn't great, but that's a separate problem.)
3. More importantly still, in my perusal of stock-flow diagrams in the wild, the links very often *don't* have signs, and thus keeping the old theory around is a way of respecting the tradition while still being able to go beyond it

epatters:
> @epatters I'm trying to get + and - to appear on positive / negative links in the visualisation, and it's proving very frustrating. What's currently confusing me even more is the file `svg_styles.module.css`, which seems to tell me that links should be blue and rectangles should be a delightful ube colour, but this isn't what's happening in reality. Any suggestions?

Argh, there appears to be a regression on next in 

- coloring places in a Petri net
- coloring links in a stock-flow diagram on next.

On prod they look fine. @kasbah, I wonder if this is an unintended consequence of #888? I don't have time to investigate right now.

But note @tim-at-topos that the fun ube color is specific to places in Petri nets: the CSS selector is `.place rect`. However, making the links be blue *should* work and, until recently, did.

kasbah:
> On prod they look fine. @kasbah, I wonder if this is an unintended consequence of https://github.com/ToposInstitute/CatColab/pull/888? I don't have time to investigate right now.

My observations differ. Importantly I don't see any regression from prod to next.

- Petri net place *borders* are colored everywhere (prod, next, loaclhost/main)
- Petri net fill is set to transparent on localhost but colored on main 

<details>
<summary>screenshots</summary>
image

image
image

</details>

There is a conflict between `.node rect { fill: transparent; }` set in `visualization/graph_svg.css` and `fill: #e6e6ff` set in `svg_styles.module.css`. Presumably development and production servers result in different resolution order.
	
- Stock and flow links are not colored anywhere (prod, next, localhost/main)

image

EDIT: This is also due to a similar conflict. It *can* be worked around with `!important`. 

I made an issue for it: https://github.com/ToposInstitute/CatColab/issues/913

epatters:
> Presumably development and production servers result in different resolution order.

That must be it, although I *do* see different behavior in prod and next/local for Petri nets. In any case, I've got a fix in #915.

PR #914, opened 2025-12-16
Migration from signed stock-flow to CLD

tim-at-topos:
To-do:

- [x] write some tests
- [ ] make those tests pass

### DIFF
--- a/packages/catlog/src/stdlib/analyses/signed_links_to_signed_cat.rs
+++ b/packages/catlog/src/stdlib/analyses/signed_links_to_signed_cat.rs
@@ -112,6 +112,7 @@ mod tests {
     use super::*;
     use crate::dbl::model::MutDblModel;
     use crate::stdlib::negative_backward_link;
+    use crate::zero::QualifiedName;
     use std::rc::Rc;
 
     #[test]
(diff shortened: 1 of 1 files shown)


tim-at-topos:
@epatters This test fails, but for "bad reasons" — the models are indeed actually the same, but they are ordered differently. Would you prefer me to

1. Change the test to simply iterate through all objects and all morphisms and compare them one by one, removing any matches from each model as we go and ensuring that we end up with two empty models?
2. Actually implement some general method for checking models are equal "up to reordering of `HashIndex`"?
3. Something else entirely?

I suppose that one thing that (3) could be is building a visualisation analysis that displays the CLD, but this is roughly as non-trivial as (2) imho (for somebody like me), so I'd rather save that specific suggestion for a future PR

epatters:
@tim-at-topos, is this one still blocked on me? If you want to pick this back up, I can try to debug why your test is failing.

tim-at-topos:
From memory (and my unhelpful commit messages), I made the tests pass in the end by checking for equality of models by just iterating over all `ob_generators` and `mor_generators`, which **does** pass, even though checking for equality of models directly **does not**. Maybe you could just check that you can reproduce the same behaviour?

Aside from that, I guess I'm happy to ask for this PR to be considered for merging, but I must admit that I haven't yet figured out the next step. The original plan was to just preview the migrated model in an analysis, which seems like something that would be generally useful (i.e. making analyses work with a derived model of a model rather than just the model itself). To actually allow users to migrate along this migration was blocked by us figuring out how we wanted derived models to behave (maybe something that deserves an RFC?).

PR #918, opened 2025-12-18
Visualize schemas as entity-relationship diagrams

kasbah:
Adds entity-relationship diagram analysis for simple schemas that uses Graphviz HTML-like tables and crows-foot edges.

image

https://branch-918-merge--catcolab.netlify.app/analysis/019b3206-c98c-7f40-a3df-027faf6ee5a6

### DIFF
--- a/packages/frontend/src/stdlib/analyses/schema_erd.tsx
+++ b/packages/frontend/src/stdlib/analyses/schema_erd.tsx
@@ -1,0 +1,268 @@
+import type * as Viz from "@viz-js/viz";
+import { escape as escapeHtml } from "html-escaper";
+import download from "js-file-download";
+import Download from "lucide-solid/icons/download";
+import { createResource, Show } from "solid-js";
+
+import { Foldable, FormGroup, IconButton, SelectField } from "catcolab-ui-components";
+import type { DblModel, MorType, ObType } from "catlog-wasm";
+import type { ModelAnalysisProps } from "../../analysis";
+import { loadViz } from "../../visualization";
+import { Direction, type SchemaERDConfig } from "./schema_erd_config";
+
+/** Visualize a schema as an Entity-Relationship Diagram.
+
+This visualization is specifically designed for schemas (models of the simple-schema
+theory) and displays entities as tables with their attributes listed inside.
+*/
+export default function SchemaERD(props: ModelAnalysisProps<SchemaERDConfig>) {
+    const graph = () => {
+        const model = props.liveModel.elaboratedModel();
+        if (model) {
+            return schemaToERD(model);
+        }
+    };
+
+    const [vizResource] = createResource(loadViz);
+
+    const svgString = () => {
+        const viz = vizResource();
+        const g = graph();
+        if (!viz || !g) {
+            return undefined;
+        }
+
+        const direction = props.content.direction ?? Direction.Vertical;
+        return viz.renderString(g, {
+            format: "svg",
+            graphAttributes: {
+                rankdir: direction === Direction.Horizontal ? "LR" : "TB",
+            },
+        });
+    };
+
+    const schemaName = () => props.liveModel.liveDoc.doc.name || "Untitled";
+    const header = () => (
+        <>
+            <IconButton
+                onClick={() => {
+                    const svg = svgString();
+                    if (svg) {
+                        download(svg, `${schemaName()} - ERD.svg`, "image/svg+xml");
+                    }
+                }}
+                disabled={!svgString()}
+                tooltip={`Export the entity-relationship diagram as SVG`}
+            >
+                <Download size={16} />
+            </IconButton>
+        </>
+    );
+
+    return (
+        <div class="graph-visualization-container">
+            <Foldable title="Entity-relationship diagram" header={header()}>
+                <FormGroup compact>
+                    <SelectField
+                        label="Direction"
+                        value={props.content.direction ?? Direction.Vertical}
+                        onChange={(evt) => {
+                            props.changeContent((content) => {
+                                content.direction = evt.currentTarget.value as Direction;
+                            });
+                        }}
+                    >
+                        <option value={Direction.Horizontal}>{"Horizontal"}</option>
+                        <option value={Direction.Vertical}>{"Vertical"}</option>
+                    </SelectField>
+                </FormGroup>
+            </Foldable>
+            <div class="graph-visualization">
+                <Show when={svgString()}>{(svg) => <div innerHTML={svg()} />}</Show>
+            </div>
+        </div>
+    );
+}
+
+/** Convert a schema model into an ERD-style Graphviz graph using HTML-like labels. */
+export function schemaToERD(model: DblModel): Viz.Graph {
+    const entityType: ObType = { tag: "Basic", content: "Entity" };
+    const attrType: MorType = { tag: "Basic", content: "Attr" };
+
+    const entities = model.obGeneratorsWithType(entityType);
+    const nodes: Required<Viz.Graph>["nodes"] = [];
+
+    // Collect all mappings to know which entities they point to
+    const mappingsByEntity = new Map<
+        string,
+        Array<{ id: string; name: string; targetEntity: string }>
+    >();
+    const mappingType: MorType = { tag: "Hom", content: entityType };
+    for (const morId of model.morGeneratorsWithType(mappingType)) {
+        const mor = model.morPresentation(morId);
+        if (
+            mor &&
+            mor.dom.tag === "Basic" &&
+            mor.cod.tag === "Basic" &&
+            entities.includes(mor.dom.content) &&
+            entities.includes(mor.cod.content)
+        ) {
+            const mappingName = mor.label?.join(".") ?? "";
+            const sourceEntity = mor.dom.content;
+            const targetEntity = mor.cod.content;
+
+            if (!mappingsByEntity.has(sourceEntity)) {
+                mappingsByEntity.set(sourceEntity, []);
+            }
+            mappingsByEntity.get(sourceEntity)?.push({
+                id: morId,
+                name: mappingName,
+                targetEntity,
+            });
+        }
+    }
+
+    // Build entity tables
+    for (const entityId of entities) {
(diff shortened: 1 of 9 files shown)


kasbah:
- Possibly this should be called a "Entity relationship graph" to avoid confusion with diagrams as we currently have them?
- Does it make sense to have this visualization as a `diagramAnalysis`? I am not familiar enough with diagrams to explore that.

packages/frontend/src/stdlib/analyses/schema_erd.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+    const entities = model.obGeneratorsWithType(entityType);
+    const nodes: Required<Viz.Graph>["nodes"] = [];
+
+    // Collect all mappings to know which entities they point to
+    const mappingsByEntity = new Map<
+        string,
+        Array<{ id: string; name: string; targetEntity: string }>
+    >();
+    for (const morId of model.morGenerators()) {
```
epatters: Rather than iterating over all morphism generators and then filtering manually, you can do:

```typescript
const mappingType: MorType = { tag: "Hom", content: entityType };

for (const morId of model.morGeneratorsWithType(mappingType)) {
   ...
}
```

packages/frontend/src/stdlib/analyses/schema_erd.tsx
```diff
@@ -0,0 +1,263 @@
...
+            mor.morType.content.content === "Entity";
+
+        if (
+            isMappingType &&
+            mor.dom.tag === "Basic" &&
+            mor.cod.tag === "Basic" &&
+            entities.includes(mor.dom.content) &&
+            entities.includes(mor.cod.content)
+        ) {
+            const mappingName = mor.label?.every((seg) => seg !== "") ? mor.label.join(".") : morId;
```
epatters: The `else` clause doesn't seem right: `morId` is a UUID and thus should never be displayed. In this case, you could return `""` or something like `"unnamed"`.

packages/frontend/src/stdlib/analyses/schema_erd_config.ts
```diff
@@ -0,0 +1,13 @@
+/** Layout direction for graph layouts with a primary/preferred direction. */
+export enum Direction {
+    Horizontal = "horizontal",
+    Vertical = "vertical",
+}
```
epatters, line 5: Instead of redefining this enum, shouldn't we import [`Direction`](https://github.com/ToposInstitute/CatColab/blob/419a5495032a01a76ecec13dfdf011e78247d15e/packages/frontend/src/visualization/graph_layout_config.ts#L22) from `graph_layout_config`?

packages/frontend/src/stdlib/analyses/schema_erd.tsx
```diff
@@ -0,0 +1,263 @@
...
+                attributes.push({ name: attrName, type: attrTypeName });
+            }
+        }
+        const mappings = mappingsByEntity.get(entityId) || [];
+
+        const entityLabel = escapeHtml(entity.label?.join(".") || entityId);
+        const paddingLeft = computePaddingCenteredLeft(entityLabel);
+        const paddingRight = computePaddingCenteredRight(entityLabel);
+        // We cannot use CSS custom properties for the color here
+        let tableRows = `<tr><td port="${entityId}" bgcolor="#a6f2f2" align="center" colspan="2"><b><font point-size="12">${paddingLeft}${entityLabel}${paddingRight}</font></b></td></tr>`;
```
epatters: Can we split these long HTML-like strings into multiline strings for readability?

packages/frontend/src/stdlib/analyses/schema_erd.tsx
```diff
@@ -0,0 +1,263 @@
...
+                </FormGroup>
+            </Foldable>
+            <div class="graph-visualization">
+                <Show when={svgString()}>
+                    {(svg) => (
+                        <div
+                            ref={(el) => {
+                                const svgEl = el.querySelector("svg");
+                                if (svgEl) {
+                                    setSvgRef(svgEl);
```
epatters: Do you need this ref to the SVG element? You're setting the inner HTML directly, and `svgEl` is used only in the download button, but the SVG string is already available. Perhaps skip `DownloadSVGButton` and just directly download Graphviz's own SVG output?

packages/frontend/src/stdlib/analyses/schema_erd.tsx
```diff
@@ -0,0 +1,263 @@
...
+            });
+        }
+    }
+
+    // Build entity tables
+    for (const entityId of entities) {
+        const entity = model.obPresentation(entityId);
+
+        const attributes: Array<{ name: string; type: string }> = [];
+        for (const morId of model.morGenerators()) {
```
epatters: You can simplify this loop similarly:

```typescript
for (const morId of model.morGeneratorsWithType(attrType)) {
   ...
}
```

epatters:
> Possibly this should be called a "Entity relationship graph" to avoid confusion with diagrams as we currently have them?

Yeah, I dunno. "ER diagram" is the standard term, so I'm inclined to go with it despite the possibility of confusion. In a different direction, @tim-at-topos and I have discussed renaming the "diagrams" in CatColab to "diagrammatic instances," which is longer but less confusing and suggests the connection with instances in general. At the root of all this is the fact that [diagram](https://ncatlab.org/nlab/show/diagram) is a technical term in category theory that is inspired by, but also clashes with, the colloquial meaning of "diagram." 

>  Does it make sense to have this visualization as a `diagramAnalysis`? I am not familiar enough with diagrams to explore that.

I think not. You can think of diagrams in CatColab as more like knowledge graphs, displaying *particular* entities and their relationships, rather than the general schema of database/knowledge graph as in an ER diagram.

packages/frontend/src/stdlib/analyses/schema_erd_config.ts
kasbah: I'd like to decouple them because if we decide other graphs can now go from e.g. "bottom to top" it hasn't been tested for this fiddly HTML-like layout and might not make sense for these graphs anyway.

packages/frontend/src/stdlib/analyses/schema_erd.tsx
kasbah: Unfortunately the white space is significant so everything inside the `<td>` needs to be on one line. I slightly improved on what was there though.

packages/frontend/src/stdlib/analyses/schema_erd.tsx
kasbah: I removed it and gave the downloaded SVG a better name along with it.

packages/frontend/src/stdlib/analyses/schema_erd.tsx
kasbah: Thanks, I adjusted it. I added a bunch of `invariant`s that amount to type assertions to me. Am I right in thinking we could assert this on the type level?

https://github.com/ToposInstitute/CatColab/pull/918/files#diff-2a683752a86d4793e5188b438028a1dbe62bee64d2482da744d19683590675e6R104-R114

packages/frontend/src/stdlib/analyses/schema_erd.tsx
```diff
@@ -0,0 +1,280 @@
...
+        invariant(
+            mor.dom.tag === "Basic",
+            `Unexpected morphism domain. Expecting 'Basic' got: ${mor.dom.tag}`,
+        );
+        invariant(
+            mor.cod.tag === "Basic",
+            `Unexpected morphism co-domain. Expecting 'Basic' got: ${mor.dom.tag}`,
+        );
+        invariant(entities.includes(mor.dom.content), "Domain content not in entities");
+        invariant(entities.includes(mor.cod.content), "Codomain content not in entities");
```
epatters: Since this code is using the elaborated model, but regardless of whether it validates, the model *can* be ill-typed if the user has made a mistake. Thus, the safe thing is to replace these invariants with conditionals of the form

```typescript
if (!(...)) {
    continue;
}
```

Alternatively, you could only show the visualization when the model validates, in which case it would be correct to use invariants, but IMO it's better to make visualization analyses degrade gracefully in the face of user error when possible.

packages/frontend/src/stdlib/analyses/schema_erd.tsx
```diff
@@ -0,0 +1,280 @@
...
+            const mor = model.morPresentation(morId);
+            invariant(mor, `Could not find attribute morphism presentation for id: ${morId}`);
+            invariant(
+                mor.dom.tag === "Basic",
+                `Unexpected morphism domain. Expecting 'Basic' got: ${mor.dom.tag}`,
+            );
+            invariant(
+                mor.cod.tag === "Basic",
+                `Unexpected morphism co-domain. Expecting 'Basic' got: ${mor.dom.tag}`,
+            );
```
epatters: Ditto for these invariants.

packages/frontend/src/stdlib/analyses/schema_erd.tsx
epatters: The question of whether to use `invariant`s is a bit subtle. See my re-review below.

packages/frontend/src/stdlib/analyses/schema_erd.tsx
kasbah: Ok, thanks. I briefly tried to create an invalid model but failed to do so, so I jumped to the conclusion that it wasn't possible. I've gone back to using an `if` block.

epatters: Thanks. Yeah, I actually thought this code would trigger errors and was surprised when it didn't during my testing. I think this has to do with recent changes to the elaborator for models of discrete theories (now using DoubleTT rather than the previous, more direct method) which is causing invalid arrows to be filtered out during elaboration when they weren't before. However, this behavior should be considered accidental and isn't guaranteed to persist. The general rule is that you can't assume a model to be valid/well-typed unless it passes validation.

epatters (approved):

PR #920, opened 2025-12-19
Improve notebook cell drag and drop

kasbah:
- Fix fiddly-ness around where the mouse is when dropped
- Use the notebook cell itself as the preview
- Allow dropping a cell back where it was (most intuitive way to cancel the drag and drop)


epatters (approved):
Thanks Kaspar, this is much appreciated! @quffaro and I have both spent time on drag and drop and it's amazing how fiddly this stuff is.

PR #922, opened 2026-01-05
Mathematical presentation of DoubleTT: first steps

epatters:
What's written here currently only partially overlaps with what is implemented (neither strictly contains the other) but the hope is that over time this will converge.

### DIFF
--- a/math-docs/trees/tt-0001.tree
+++ b/math-docs/trees/tt-0001.tree
@@ -15,13 +15,15 @@ describing models of \dbl{T}.}
 \transclude{tt-0005}
 
 \subtree{
-\title{Modalities of DoubleTT}
+\title{Modalities in DoubleTT}
 
 \p{Of the following modalities on the double theory, the [codiscrete
-modality](tt-0007) is particularly important as it makes admissable a modality
-on the type theory itself. By contrast, the [discrete modality](tt-0006) is work
-in progress and not included below.}
+modality](tt-0007) is particularly important as it makes admissible a
+[truncation modality](tt-0009) on the type theory itself. By contrast, the
+[discrete modality](tt-0006) is work in progress and not included below.}
 
 \transclude{tt-0007}
 \transclude{tt-0008}
 }
+
+\transclude{tt-0009}
--- a/math-docs/trees/tt-0005.tree
+++ b/math-docs/trees/tt-0005.tree
@@ -14,7 +14,7 @@
   \prooftree{
   \hypo{\dbl{T} \vDash X\ \ob}
   \hypo{\vdash \Gamma\ \context}
-  \infer2{\vdash \Ob_X\ \type}
+  \infer2{\Gamma \vdash \Ob_X\ \type}
   }
   }
 
--- a/math-docs/trees/tt-0007.tree
+++ b/math-docs/trees/tt-0007.tree
@@ -48,7 +48,7 @@ the double theory.}
   \infer[dashed]3{\Gamma \vdash (\coDisc F)(f) = *: 1}
   }
 
-  Note that the first judgment makes the other two judgments admissable.
+  Note that the first judgment makes the other two judgments admissible.
   }
 
   \li{Computation rules for morphism types:
@@ -67,7 +67,7 @@ the double theory.}
   \infer[dashed]4{\Gamma \vdash \eta_P(p) = *: 1}
   }
 
-  \strong{TODO}: Admissable rule for cell applications
+  \strong{TODO}: Admissible rule for cell applications
   }
 
 }
--- a/math-docs/trees/tt-0009.tree
+++ b/math-docs/trees/tt-0009.tree
@@ -1,0 +1,81 @@
+\title{Truncation modality on DoubleTT}
+\import{macros}
+
+\p{The codiscrete modality \em{in} DoubleTT induces an admissible modality
+\em{on} DoubleTT: a meta-level operation that truncates a type generically
+involving both object and morphism types to one involving only object types.
+This is useful for type checking because the object part of DoubleTT is a simple
+type theory, hence types and terms can be normalized using standard
+normalization-by-evaluation algorithms.}
+
+\p{The idea behind this meta-level modality, denoted #{\bigcirc}, is to apply
+the codiscrete functor to types, apply the unit of the codiscrete monad to
+terms, and then apply the computation rules for the codiscrete modality to
+normalize the result.}
+
+\p{To be more precise, we define #{\bigcirc} by structural induction on
+contexts, types, and terms, with the following definitions as base cases:
+
+\ul{
+  \li{on object types and terms:
+
+  \p{\centerfigs{
+  \prooftreeinline{
+  \hypo{\dbl{T} \vDash X\ \ob}
+  \hypo{\vdash \Gamma\ \context}
+  \infer2{\Gamma \vdash \bigcirc \Ob_X := \Ob_{\coDisc(X)}\ \type}
+  }
+  #{\qquad\qquad}
+  \prooftreeinline{
+  \hypo{\dbl{T} \vDash X\ \ob}
+  \hypo{\Gamma \vdash x: \Ob_X}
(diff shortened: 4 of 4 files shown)


math-docs/trees/tt-0004.tree
```diff
@@ -0,0 +1,54 @@
...
+\p{A \define{simple double theory} is an augmented virtual double category with
+all restrictions, i.e., having both unary and nullary restrictions.}
+
+}
+
+\subtree{
+\title{Model of simple double theory}
+\taxon{definition}
+
+\p{Let \dbl{S} be a virtual double category with restrictions. A \define{model}
```
KevinDCarlson, line 31: Is Mod: VDC -> aVDC a right adjoint? Just to the forgetful functor?

math-docs/trees/tt-0002.tree
```diff
@@ -0,0 +1,53 @@
+\title{Judgments of DoubleTT}
+\import{macros}
+
+\p{
+
+DoubleTT has the four standard sorts (meta types) of dependent type theory:
```
KevinDCarlson, line 6: (with explicit substitutions, right?)

math-docs/trees/tt-0002.tree
```diff
@@ -0,0 +1,53 @@
...
+\prooftree{
+\hypo{\Gamma\ \context}
+\hypo{\vdash A\ \type_0}
+\infer2{\Gamma \vdash A\ \type}
+}
+
+This approach has the virtue of making explicit that object types in DoubleTT do
+not depend on contexts, i.e., object types are the types of a \em{simple type
+theory}. However, we find that we can achieve much the same benefit by working
+in a standard dependent type theory that includes a \em{codiscrete modality}.
```
KevinDCarlson, line 49: interrobang

math-docs/trees/tt-0003.tree
```diff
@@ -0,0 +1,239 @@
...
+}
+
+\subtree{
+  \title{Literature on dependent records}
+  \taxon{remark}
+
+  \p{Apart from cosmetic differences, our presentation of record types mostly
+  follows the standard reference ([Pollack 2002](pollack-2002), Section 3.1).
+  The only meaningful difference is that our presentation, unlike Pollack's,
+  includes the uniqueness rule for records, which seems important inasmuch as
```
KevinDCarlson, line 187: This smells to me like it could be a confusion about how type theories are always free, and maybe uniqueness is admissible. Do Owen and DJM have opinions on adding this rule?

math-docs/trees/tt-0005.tree
```diff
@@ -0,0 +1,124 @@
...
+\import{macros}
+
+\p{The following rules are specific to DoubleTT.}
+
+\subtree{
+\title{Object types}
+
+\ul{
+
+  \li{\em{Object type former}: For each object in the double theory, there is an
```
KevinDCarlson, line 11: It seems very strange to have a context in the premise that's not in the conclusion. Is this a mix of intended rules for the absolute type and for its relativization to other contexts?

math-docs/trees/tt-0007.tree
```diff
@@ -0,0 +1,73 @@
+\title{Codiscrete modality}
+\import{macros}
+
+\p{The codiscrete modality is an idempotent monad #{(\dbl{T}, \coDisc, \eta)} on
```
KevinDCarlson, line 4: This was a surprise, are we going to have to assume double theories all come equipped with a codiscrete monad?

math-docs/trees/tt-0001.tree
```diff
@@ -0,0 +1,27 @@
...
+
+\transclude{tt-0003}
+
+\transclude{tt-0005}
+
+\subtree{
+\title{Modalities of DoubleTT}
+
+\p{Of the following modalities on the double theory, the [codiscrete
+modality](tt-0007) is particularly important as it makes admissable a modality
```
KevinDCarlson: Should be "admissible."

math-docs/trees/tt-0007.tree
```diff
@@ -0,0 +1,73 @@
...
+  \li{Computation rules for object types:
+
+  \p{\centerfigs{
+  \prooftreeinline{
+  \hypo{\dbl{T} \vDash X\ \ob}
+  \hypo{\vdash \Gamma\ \context}
+  \infer2{\Gamma \vdash \Ob_{\coDisc(X)} = \Ob_X\ \type}
+  }
+  #{\qquad\qquad}
+  \prooftreeinline{
```
KevinDCarlson, line 18: Why are these two dashed?

math-docs/trees/tt-0004.tree
epatters: Good question. Cruttwell-Shulman have the result that Mod: VDC -> uVDC (unital VDCs and normal functors) is right pseudo-adjoint to the forgetful 2-functor (Prop 5.14). Not sure what happens when you replace the modifier "u" with "a".

math-docs/trees/tt-0002.tree
epatters: I've fudged this point: Angiuli & Gratzer take the approach of having explicit substitutions, which is reflected in my including substitution as a kind of judgment, but I haven't written any of the *rules* for such substitutions. However, I got the impression that DJM and Owen were imagining the traditional, purely meta-level substitutions. As to what we *should* be doing, I don't know.

math-docs/trees/tt-0003.tree
epatters: I agree, it could be related to that point. I don't know their opinions on this matter. I've followed Angiuli and Gratzer's presentation of Sigma types, which includes the uniqueness rule.

math-docs/trees/tt-0005.tree
epatters: Thanks, it was a typo to omit the context in the conclusion. Fixed.

math-docs/trees/tt-0007.tree
epatters: The dashed line is apparently a conventional notation in type theory for [admissible rules](https://ncatlab.org/nlab/show/admissible+rule).

math-docs/trees/tt-0002.tree
epatters: I just pushed a commit expanding on this point.

math-docs/trees/tt-0007.tree
epatters: In a sense, yes, but my current hypothesis is that we should avoid trying to create a separate DoubleTT implementation for each double doctrine, instead making an uber DoubleTT that encompasses all the features we might want.

math-docs/trees/tt-0005.tree
```diff
@@ -0,0 +1,124 @@
...
+}
+}
+
+\subtree{
+\title{Hom types}
+
+\ul{
+  
+  \li{\em{Hom type former}: For each object in the double theory, there is a
+  family of Hom types, parameterized by pairs of object terms:
```
KevinDCarlson, line 39: Isn't it a bit more general to have Hom_X live in context $\Gamma, Ob_X, Ob_X$, and then get these particular types via substitution?

math-docs/trees/tt-0005.tree
```diff
@@ -0,0 +1,124 @@
...
+}
+}
+
+\subtree{
+\title{Morphism types}
+
+\ul{
+
+  \li{\em{Morphism type former}: For each proarrow in the double theory, there
+  is a family of morphism types, parameterized by pairs of object terms:
```
KevinDCarlson, line 89: x,y are in Ob_X, Ob_Y

math-docs/trees/tt-0003.tree
KevinDCarlson: I spent some time trying to understand this and it seems like the idea is that $\eta$ will be admissible if your $\Sigma$-types are inductive, i.e. defined positively, but will need to be imposed if your $\Sigma$-types are negative, as in Angiuli and Gratzer. Pollack's presentation of Sigma with no eta nor eliminator in his figure 1 is misleading, as his implementation just below is indeed inductivev.

Issue #766, opened 2025-10-21
Display equations generated in analyses as LaTeX math


epatters:
Most of the quantitative analyses first generate an ODE system and then simulate that system for some choice of parameters. It should be possible to display the equations as LaTeX math, rendered by KaTeX in the frontend.

All of the existing ODE systems are polynomial (though not all currently pass through the `PolynomialSystem` type). Given that, my suggested implementation strategy is:

1. On the Rust side, write a LaTeX pretty-printer for `PolynomialSystem`
2. Run the pretty-printer on a `PolynomialSystem` with *symbolic* coefficients
3. Display the results in the frontend using KaTeX

kasbah:
Can you give me some example documents where this feature would be useful please?

epatters:
Sure. You can find an example for this feature in the Introduction to [Baez-Pollard](https://arxiv.org/abs/1704.02051). The Petri net for the SIR model *with waning immunity*

<img width="305" height="249" alt="Image" src="https://github.com/user-attachments/assets/02833a99-4c73-4119-b784-29953ddf5b68" />

generates, under the deterministic mass-action semantics, the ODE system

<img width="229" height="195" alt="Image" src="https://github.com/user-attachments/assets/9de99eb7-da80-474f-9a89-d5386908bfb6" />

The system of equations should be displayed much as above. In CatColab, we tend to use descriptive names rather than Greek letters for transitions, so in reproducing this example, we'd likely have $r_{\mathrm{infect}}$, $r_{\mathrm{recover}}$, etc instead of $r_\iota$, $r_\rho$, etc for the rate coefficients.

kasbah:
@tim-at-topos and I discussed and came up with a UI like this where only "Equations" at the top would be shown by default. "Simulation" can then be expanded but the second "Equations" with substitutions would be collapsed by default when expanding that.

- This clashes with current use as this dropdown `Foldable` for settings in e.g. visualization. I think we should eventually replace that usage with some sort of settings widget.

- We can also optimize the simulation to not run until it's expanded. 

- We should change the tables for the rates to use the same name e.g. $R_{\mathrm{recover}}$ (should be lowercase `r` but ignore that for now)

- In most cases of actual usage there will be too many equations so we will adopt a similar pattern to what I sketched out in the documents list mockups: a click to expand.

<img width="346" height="88" alt="Image" src="https://github.com/user-attachments/assets/cb821b03-e799-4072-a4aa-73eacd982b8a" />

---

[...]

PR #923, opened 2026-01-05
Add `ExpandableList` and `KatexDisplay` components

kasbah:
This is some ground work for #766 adding two components: a collapsed/expandable list and a Katex display. 

- [ExpandableList](https://branch-923-merge--catcolab.netlify.app/dev/ui-components/?path=/docs/layout-expandablelist--docs)
- [KatexDisplay](https://branch-923-merge--catcolab.netlify.app/dev/ui-components/?path=/docs/misc-katexdisplay--docs)
- [KatexDisplay in ExpandableList](https://branch-923-merge--catcolab.netlify.app/dev/ui-components/?path=/story/layout-expandablelist--with-katex)

### DIFF
--- a/packages/ui-components/src/expandable_list.module.css
+++ b/packages/ui-components/src/expandable_list.module.css
@@ -1,0 +1,50 @@
+.expandableList {
+    display: flex;
+    flex-direction: column;
+    gap: 0.5rem;
+}
+
+.expandableListTitle {
+    cursor: pointer;
+}
+
+.expandableListTitle:hover {
+    text-decoration: underline;
+}
+
+.expandableListTitle:focus-visible {
+    outline: 2px solid var(--color-button-positive-base);
+    outline-offset: 2px;
+    border-radius: 2px;
+}
+
+.expandableListItems {
+    display: flex;
+    flex-direction: column;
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/katex_display.tsx
```diff
@@ -0,0 +1,30 @@
...
+        } catch (error) {
+            console.error("KaTeX rendering error:", error);
+            container.textContent = `Error rendering math: ${error}`;
+        }
+    };
+
+    createEffect(() => {
+        // Track the math prop for reactivity
+        props.math;
+        renderMath();
```
epatters: A small point, but IMO it would be more idiomatic to have `renderMath` be a function `renderMath(math: string)` and then replace the above with

```typescript
createEffect(() => renderMath(props.math));
```

epatters (changes requested):
Thanks, these components look great! Just one minor comment.

packages/ui-components/src/katex_display.tsx
kasbah: Thanks, agreed, fixed.

epatters (approved):

PR #924, opened 2026-01-05
Upgrade to NixOS 25.11

jmoggr:
The upgrade went very smoothly and took ~15min outside of unearthed issues with `wasm-pack`.

---

We are using `wasm-pack` to build the `notebook-types` and `catlog` packages, and `wasm-pack` expects to find version of `wasm-bindgen-cli` in the environment that matches the Cargo version of `wasm-bindgen` used by the package being built. It appears that coincidentally this has always been true, up until now. Note: outside of Nix `wasm-pack` will download the necessary `wasm-bindgen`.

This is very similar to the issue of the npm biome version not matching the nix version. The workaround employed there was to just use a biome version that matches whatever version of biome is used by nixpkgs-unstable. This is not really feasible for a critical library like `wasm-bindgen`.

It's pretty straight forward to build the build `wasm-bindgen-cli` from source using a helper function from `nixpkgs`. The downside of doing this is that we now need to update multiple Nix hashes whenever we upgrade `wasm-bindgen`, and that the version needs to be manually kept in sync between Nix and Cargo. I think that this is acceptable given that we don't frequently upgrade `wasm-bindgen`, and that this pattern should hopefully not apply to much more of our tooling.

I'm using an overlay in Nix to set the specific version of `wasm-bindgen-cli`, this is basically just inheritance for nixpkgs. The alternative is passing around our own copy of `wasm-bindgen-cli` in Nix, but in my experience I've found that to be more error prone and tedious.

If we like this pattern, it might make sense to apply it to biome as well.

### DIFF
--- a/packages/catlog-wasm/default.nix
+++ b/packages/catlog-wasm/default.nix
@@ -36,12 +36,18 @@ craneLib.buildPackage {
   # run wasm-pack instead of the default cargo
   buildPhase = ''
     cd packages/catlog-wasm
-    # WTF: engage maximum cargo cult. I have no idea wasm-pack needs $HOME set, that is wild.
-    # https://github.com/NixOS/nixpkgs/blob/b5d0681604d2acd74818561bd2f5585bfad7087d/pkgs/by-name/te/tetrio-desktop/tetrio-plus.nix#L66C7-L66C24
-    # https://discourse.nixos.org/t/help-packaging-mipsy-wasm-pack-error/51876
+    # Run the wasm-pack command. wasm-pack will expect to find version of wasm-bindgen-cli in the
+    # environment that must matches the version wasm-bindgen used in the Cargo.toml. The wasm-bindgen-cli
+    # in the nix environment is defined an overlay in flake.nix.
     #
-    # This just runs the wasm-pack command, it's a bit abstracted but it guarantees that we use the same
-    # call to wasm-pack in dev and prod
+    # If the versions do not match there will be a build error when building with nix:
+    # Error: Not able to find or install a local wasm-bindgen.
+    #
+    # With RUST_LOG=debug set there should be a log like indicating the exact problem:
+    # Checking installed `wasm-bindgen` version == expected version: 0.2.105 == 0.2.106
+    # 
+    # wasm-pack needs a writeable $HOME
+    # https://github.com/ipetkov/crane/issues/362
     HOME=$(mktemp -d) npm run build:browser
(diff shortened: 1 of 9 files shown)


packages/notebook-types/default.nix
```diff
@@ -34,12 +34,18 @@ craneLib.buildPackage {
...
+    # If the versions do not match there will be a build error when building with nix:
+    # Error: Not able to find or install a local wasm-bindgen.
+    #
+    # With RUST_LOG=debug set there should be a log like indicating the exact problem:
+    # Checking installed `wasm-bindgen` version == expected version: 0.2.105 == 0.2.106
+    # 
+    # wasm-pack needs a writeable $HOME
+    # https://github.com/ipetkov/crane/issues/362
     HOME=$(mktemp -d) npm run build:node
   '';
```
jmoggr, line 50: I don't love how this is duplicated, however #875 should remove the need for a Nix package for `notebook-types`.

packages/catlog-wasm/default.nix
```diff
@@ -36,12 +36,18 @@ craneLib.buildPackage {
...
+    # environment that must matches the version wasm-bindgen used in the Cargo.toml. The wasm-bindgen-cli
+    # in the nix environment is defined an overlay in flake.nix.
     #
-    # This just runs the wasm-pack command, it's a bit abstracted but it guarantees that we use the same
-    # call to wasm-pack in dev and prod
+    # If the versions do not match there will be a build error when building with nix:
+    # Error: Not able to find or install a local wasm-bindgen.
+    #
+    # With RUST_LOG=debug set there should be a log like indicating the exact problem:
+    # Checking installed `wasm-bindgen` version == expected version: 0.2.105 == 0.2.106
```
jmoggr, line 47: Is this the correct/best place for this kind of documentation? It seems kind of awkward here, but it can't go in the package.json file. It might be better if put in a README.md file, probably a new one in for this package?

packages/notebook-types/default.nix
epatters: Agreed. Since we've got a near-term plan that will eliminate the duplication, we don't have to sweat it right now.

packages/catlog-wasm/default.nix
epatters: Thanks for writing a detailed explanation of the issue. Putting it here seems right to me.

epatters (approved):

PR #929, opened 2026-01-05
Use a gear icon for analysis settings/params panel dropdown

kasbah:
This makes it less confusing. It would have gotten worse when we introduce actual hiding of whole sections in #766  

image
image

### DIFF
--- a/packages/frontend/src/stdlib/analyses/simulation.css
+++ b/packages/frontend/src/stdlib/analyses/simulation.css
@@ -10,6 +10,7 @@
     align-items: flex-start;
     justify-content: space-around;
     padding-bottom: 1em;
+    width: 100%;
 }
 
 .simulation .plot {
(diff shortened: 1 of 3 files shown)


epatters (commented):
I like the use of the gear icon---it is idiomatic for a settings button and will hopefully be more intuitive than the current design---but it seems like a wasteful use of precious vertical real estate to put the button below the title.

We already have an informal pattern of placing buttons (like the export SVG button) vertically aligned with the title but on the right. What do you think about making the settings toggle consistently be the right-most vertically-aligned button?

kasbah:
Good idea, I've moved it to the right. I removed the chevron because I think it's clearer with just the gear there.

I also right-aligned the contents of the settingsPanel, for e.g. the mass action parameters I added 100% width to them to make them look like before. I hope I didn't miss anything affected by the new right-alignment.

kasbah:
I just put it on the right because that's where the button is. In the long run it should probably be a drop-down kind of popup rather than something that shifts the contents of the document down.

PR #933, opened 2026-01-07
Set browser title for all pages

### LINKED ISSUE #927: Show document title in HTML page title
If you've been to a bunch of CatColab pages recently, the browser history is not so useful: 

Image

Probably we could modify these to display the title and/or type of the document?

kasbah:
closes #927

### DIFF
--- a/packages/automerge-doc-server/default.nix
+++ b/packages/automerge-doc-server/default.nix
@@ -69,7 +69,7 @@ pkgs.stdenv.mkDerivation {
 
     # See README.md
     # hash = "";
-    hash = "sha256-8k7dLMkdFn9/fhRUJURrROqFjHwW76pQWokko5yhvJY=";
+    hash = "sha256-/ND19AuSHsXnkrxpoLOARx7GNEg6SQj362UxdWm83UA=";
   };
 
   meta.mainProgram = name;
--- a/packages/backend/pkg/package.json
+++ b/packages/backend/pkg/package.json
@@ -8,7 +8,7 @@
         "build": "tsc -b"
     },
     "dependencies": {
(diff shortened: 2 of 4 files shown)


packages/frontend/src/page/document_page.tsx
```diff
@@ -121,159 +123,146 @@ export default function DocumentPage() {
...
+                                                                }
+                                                                docRef={
+                                                                    secondaryLiveDocWithRef().docRef
+                                                                }
+                                                                refetchPrimaryDoc={
+                                                                    refetchPrimaryDoc
+                                                                }
+                                                                refetchSecondaryDoc={
+                                                                    refetchSecondaryDoc
+                                                                }
```
epatters: The deep nesting is making this increasingly unpleasant to read. We should probably break this component down into subcomponents if it's not too inconvenient.

epatters (commented):
Thanks Kaspar, this will be a nice improvement! Looks like there's a merge conflict, plus one small comment below.

packages/frontend/src/page/document_page.tsx
kasbah: I split out two components which helps with the nesting but this is all still quite messy due to prop-drilling and our state management.

kasbah:
I updated the hash. I'm not sure what else to do regarding [CI failure](https://github.com/ToposInstitute/CatColab/actions/runs/20802583624/job/59750331777?pr=933)  @jmoggr

packages/frontend/src/page/document_page.tsx
epatters: Yeah, I was afraid of that but I still think it's an improvement. Thanks!

PR #943, opened 2026-01-12
Add dev docs for Nix hash mismatches

jmoggr:
I'm not sure why the dev-docs folder was set up as a npm package. I've taken the liberty of repurposing it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,7 @@
+---
+title: "Developer documentation: overview"
+---
+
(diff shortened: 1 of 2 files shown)


kasbah (changes requested):
Because it [gets built](https://github.com/ToposInstitute/CatColab/blob/cc6fccb1741c408763ff75d0558258d75ee054d1/.github/workflows/deploy.yml#L46-L99) and deployed to https://next.catcolab.org/dev ! From before my time but I don't think we should change that right now like this.

epatters (commented):
Thanks for noticing that, Kaspar. I do think it's useful to have a landing page for developer docs since they're spread over several languages and packages.

jmoggr:
I restored the npm package and updated the build to include the markdown file I added.

PR #945, opened 2026-01-13
Consistent name and icon for links to help pages

tim-at-topos:
Just making the names of the "help" page a bit more consistent along the UI.

@kasbah I'd quite like to actually include the icon for the permissions button after the words "clicking the permissions button"  in the below

https://github.com/ToposInstitute/CatColab/blob/b92843f2c3b6d7190e6ebb1208a15352a56c1dc3/packages/frontend/src/help/guide/example-models.mdx?plain=1#L4

Could you please help me with this?


kasbah:
I pushed a commit that adds the icon. It is slightly awkward with the line-height being increased to fit the icon for that line.

I was thinking about the title wording last week actually but didn't get around to doing anything about it. The title and icon should be the same on the home page as well. I pushed another commit for that.

PR #946, opened 2026-01-14
Build rust docs with Nix in CI

jmoggr:
A possible solution to flaky caches in github actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,19 +61,25 @@ jobs:
         run: |
           cargo clippy --all-targets --all-features -- --deny warnings
 
-  rust_docs:
-    name: rust docs
+  check_rust_docs:
+    name: Build Rust docs
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout Repository
+      - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Install Rust toolchain from file
-        uses: actions-rust-lang/setup-rust-toolchain@v1
+      - name: Install Nix
+        uses: cachix/install-nix-action@v25
 
-      - name: Check rustdoc warnings
+      - name: Configure Cachix
+        uses: cachix/cachix-action@v14
+        with:
+          name: catcolab-jmoggr
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
+      - name: Check Rust docs
         run: |
-          RUSTDOCFLAGS='--deny warnings' cargo doc --no-deps
+          nix build .#rust-docs-check
 
   ui_components_tests:
     name: ui-components tests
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -61,20 +61,6 @@ jobs:
...
-    steps:
-      - name: Checkout Repository
-        uses: actions/checkout@v4
-
-      - name: Install Rust toolchain from file
-        uses: actions-rust-lang/setup-rust-toolchain@v1
-
-      - name: Check rustdoc warnings
-        run: |
-          RUSTDOCFLAGS='--deny warnings' cargo doc --no-deps
```
jmoggr, line 76: This check is now done when we build the rust-docs nix package. I'm not sure why this needed to be a separate check run in ci.yml, we're basically running the exact same code in deploy.yml, but without checking.

.github/workflows/deploy.yml
```diff
@@ -206,47 +206,28 @@ jobs:
...
-
-      - name: Cache Rust docs build
-        id: cache-rust-docs
-        uses: actions/cache@v4
+      - name: Install Nix
+        uses: cachix/install-nix-action@v25
         with:
-          path: target/doc
-          key: rust-docs-${{ steps.cache-key.outputs.git-hash }}
+          nix_path: nixpkgs=channel:25.05
```
kasbah: Should we starting with 25.11 or stay with 25.05 for cachix for now?

jmoggr: We don't actually need the `nix_path` argument. It was removed in #947 but this PR was stale. I've removed it here as well.

.github/workflows/ci.yml
epatters: We want the CI to fail when rustdoc emits warnings! So, despite the redundancy, I think it makes sense to run rustdoc in both ci.yml and deploy.yml because the purposes of building are different, namely checking/linting versus deploying.

jmoggr: :facepalm:

I added a `check_rust_docs` job to `ci.yml`, which build the nix rust-docs package (the build will fail if there are warnings).

If we uploaded the rust-docs artifact from `check_rust_docs` we could remove `build_rust_docs` from `deploy.yml`, since the artifact would already been built and uploaded. However that seems possibly confusing so I'll keep the duplication.

epatters: Isn't it now the case that if the rustdoc emits a warn, both `check_rust_docs` and `build_rust_docs` will fail? That seems confusing in its own way. Can we restore the current behavior that only `check_rust_docs` fails on warnings?

jmoggr: done! I added a `checkMode` flag to `rust-docs.nix` and created a `rust-docs-check` packages that uses it.

epatters: Great, thanks!

PR #949, opened 2026-01-16
Draft: type-checking literals in schema instances

KevinDCarlson:
This is David's Claude collaboration on seeing what it would be like to get data types into schema-land. It's not intended for review for merging, but I told him I wanted to see the code, and thus here it is.


epatters:
Thanks for sharing the results of this experiment. Closing as we've had further discussion elsewhere.

PR #954, opened 2026-01-22
Display equations for mass action ODE analysis

### LINKED ISSUE #766: Display equations generated in analyses as LaTeX math

kasbah:
closes #766 (short of porting others to `PolynomialSystem`)

- Implements a separate analysis block with symbolic equations (🫣 take it up with Evan @tim-at-topos)
- Adds an equation foldable, with parameters substituted, to the mass action simulation block

Both of these stop at 20 equations and then say "N more..." which the user can expand. 

Examples:
- [Baez-Pollard](https://branch-954-merge--catcolab.netlify.app/model/019be2c0-d8ad-7dd2-a5ba-8cb2311cbce5/analysis/019be2c0-f2e5-7f11-a032-957c9169a906)
- [Long example](https://branch-954-merge--catcolab.netlify.app/model/019be6a9-8eee-7bb1-8535-18c38845806d/analysis/019be6a9-d913-7160-938b-32de40d2ddd0)

### DIFF
--- a/packages/catlog/src/simulate/ode/polynomial.rs
+++ b/packages/catlog/src/simulate/ode/polynomial.rs
@@ -1,10 +1,11 @@
 //! Polynomial differential equations.
 
-use std::collections::BTreeMap;
 use std::fmt::Display;
+use std::hash::Hash;
 use std::ops::{Add, Neg};
 
 use derivative::Derivative;
+use indexmap::IndexMap;
 use nalgebra::DVector;
 use num_traits::{One, Pow, Zero};
 
@@ -23,12 +24,12 @@ use crate::zero::alg::Polynomial;
 #[derivative(Default(bound = ""))]
 pub struct PolynomialSystem<Var, Coef, Exp> {
     /// Components of the vector field.
-    pub components: BTreeMap<Var, Polynomial<Var, Coef, Exp>>,
+    pub components: IndexMap<Var, Polynomial<Var, Coef, Exp>>,
 }
 
 impl<Var, Coef, Exp> PolynomialSystem<Var, Coef, Exp>
 where
-    Var: Ord,
+    Var: Hash + Ord,
     Exp: Ord,
 {
     /// Constructs a new polynomial system, with no equations.
@@ -59,7 +60,7 @@ where
     /// Maps the variables of the polynomials comprising the system.
     pub fn map_variables<NewVar, F>(&self, mut f: F) -> PolynomialSystem<NewVar, Coef, Exp>
     where
-        NewVar: Clone + Ord,
+        NewVar: Clone + Hash + Ord,
         Coef: Clone + Add<Output = Coef>,
         Exp: Clone + Add<Output = Exp>,
         F: FnMut(&Var) -> NewVar,
@@ -124,7 +125,7 @@ pub struct LatexEquation {
 
 impl<Var, Exp> PolynomialSystem<Var, f32, Exp>
 where
-    Var: Clone + Ord,
+    Var: Clone + Hash + Ord,
     Exp: Clone + Ord + Add<Output = Exp>,
 {
     /// Converts the polynomial system to a numerical one.
@@ -132,7 +133,7 @@ where
     /// The order of the components in the new system is given by the order of the
     /// variables in the old one.
     pub fn to_numerical(&self) -> NumericalPolynomialSystem<Exp> {
-        let indices: BTreeMap<Var, usize> =
+        let indices: IndexMap<Var, usize> =
             self.components.keys().enumerate().map(|(i, var)| (var.clone(), i)).collect();
         let components = self
             .components
@@ -160,7 +161,7 @@ where
 impl<Var, Coef, Exp> FromIterator<(Var, Polynomial<Var, Coef, Exp>)>
     for PolynomialSystem<Var, Coef, Exp>
 where
-    Var: Ord,
+    Var: Hash + Ord,
     Coef: Add<Output = Coef>,
     Exp: Ord,
 {
@@ -215,21 +216,21 @@ mod tests {
         ];
         let sys: PolynomialSystem<_, _, _> = terms.into_iter().collect();
         let expected = expect![[r#"
+            dS = (-β) I S
             dI = (-γ) I + β I S
             dR = γ I
-            dS = (-β) I S
         "#]];
         expected.assert_eq(&sys.to_string());
 
         let sys = sys.extend_scalars(|p| p.eval(|_| 1.0));
         let expected = expect![[r#"
+            dS = -I S
             dI = -I + I S
             dR = I
-            dS = -I S
         "#]];
         expected.assert_eq(&sys.to_string());
 
-        let initial = DVector::from_column_slice(&[1.0, 0.0, 4.0]);
+        let initial = DVector::from_column_slice(&[4.0, 1.0, 0.0]);
         let problem = ODEProblem::new(sys.to_numerical(), initial).end_time(5.0);
         let result = problem.solve_rk4(0.1).unwrap();
         let expected = expect![[r#"
--- a/packages/catlog/src/stdlib/analyses/ode/mass_action.rs
+++ b/packages/catlog/src/stdlib/analyses/ode/mass_action.rs
@@ -366,9 +366,9 @@ mod tests {
         let model = catalyzed_reaction(th);
         let sys = PetriNetMassActionAnalysis::default().build_system(&model);
         let expected = expect!([r#"
-            dc = 0
             dx = (-f) c x
             dy = f c x
+            dc = 0
         "#]);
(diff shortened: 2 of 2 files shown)


tim-at-topos:
I'm so excited for this, thank you!

> * Implements a separate analysis block with symbolic equations (🫣 take it up with Evan @tim-at-topos)

As long as he has a plan for how to make this manageable when there are multiple ODE semantics to pick from without cluttering up the menu with lots of items with similar names, I'll let it slide 😉

epatters:
> I'm so excited for this, thank you!
> 
> > * Implements a separate analysis block with symbolic equations (🫣 take it up with Evan @tim-at-topos)
> 
> As long as he has a plan for how to make this manageable when there are multiple ODE semantics to pick from without cluttering up the menu with lots of items with similar names, I'll let it slide 😉

Haha, fair point. Overall I think we should prefer the relative flexibility offered by the notebook component over making increasingly elaborate, but rigid and unmodifiable, layouts within a single analysis component. In this design, you could, for example, display the equations for the derived ODE system and then have several ODE simulations using different parameters, without repeating the equations each time.

To keep it manageable, we might consider organizing the available analyses with more structure than a list. E.g., we could group related analyses in the analysis menu:

- Linear ODE dynamics
  - Equations
  - Simulation
- Lotka-Volterra ODE dynamics
  - Equations
  - Simulation

packages/catlog/src/zero/rig.rs
```diff
@@ -214,19 +214,36 @@ where
...
-                let coef = coef.to_string();
-                if coef.chars().all(|c| c.is_alphabetic())
-                    || coef.chars().all(|c| c.is_ascii_digit() || c == '.')
-                {
-                    write!(f, "{coef} ")?;
+            if coef.is_one() {
+                write!(f, "{var}")
+            } else {
+                let coef_str = coef.to_string();
+                if coef_str == "-1" {
```
epatters: I think that a cleaner/safer way to do this check is

```rust
if coef == One::one().neg() {
  [...]
}
```

This will require adding `Neg<Output = Coef>` to the trait bound for `Coef` in this method, but that seems fine.

packages/catlog/src/simulate/ode/polynomial.rs
```diff
@@ -163,17 +189,17 @@ mod tests {
         ];
         let sys: PolynomialSystem<_, _, _> = terms.into_iter().collect();
         let expected = expect![[r#"
-            dI = ((-1) γ) I + β I S
+            dI = (-γ) I + β I S
```
epatters, line 220: Yay, thanks for making this look decent!

packages/catlog/src/simulate/ode/polynomial.rs
```diff
@@ -91,6 +91,13 @@ where
     }
 }
 
+/// Trait for converting to LaTeX strings. We are using 2D vectors for laying them out as tables
+/// down the line.
+pub trait ToLatex {
+    /// Converts to a 2D vector of LaTeX strings.
+    fn to_latex(&self) -> Vec<Vec<String>>;
+}
```
epatters: IMO, the `ToLatex` trait is a premature generalization because (1) it's only got one implementation so far and (2) it's not obvious to me that returning a matrix of strings is always what you'd want for a generic "convert to LaTeX" trait. I'd suggest just removing this.

packages/catlog/src/simulate/ode/polynomial.rs
```diff
@@ -105,6 +112,25 @@ where
     }
 }
 
+impl<Var, Coef, Exp> ToLatex for PolynomialSystem<Var, Coef, Exp>
+where
+    Var: Display,
+    Coef: Display + PartialEq + One,
+    Exp: Display + PartialEq + One,
+{
+    fn to_latex(&self) -> Vec<Vec<String>> {
```
epatters: I'd call this `to_latex_equations` and have it return `Vec<LatexEquation>`, where

```rust
struct LatexEquation {
   lhs: String,
   rhs: String,
}
```

is a new auxiliary struct.

packages/catlog/src/zero/rig.rs
epatters: I guess we should delete this line since it's now getting "serious use" :)

More seriously, this docstring is now outdated/misleading, since it says that it will pretty-print as ASCII but more specifically it should be valid LaTeX. At minimum this requirement should be documented anywhere it is expected. Arguably it should be made more explicit: introduce a new method `to_latex` and then call that in `Display` as a "default" pretty-print.

(Julia actually has a nice approach for this: the display method can dispatch on MIME type, so you can be explicit about which format you're target and target multiple formats, but AFAICT there's nothing like that codified in Rust.)

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
```diff
@@ -1,9 +1,56 @@
...
+    plotData: JsResult<ODEPlotData, string>;
+    latexEquations: string[][];
+};
+
+/** Replace braced UUIDs in equation strings with human-readable labels from the model. */
+function replaceUuidsWithLabels(equation: string, model: DblModel): string {
+    const uuidPattern =
+        /\{([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})\}/g;
+
+    return equation.replace(uuidPattern, (_match, uuid) => {
```
epatters: As we discussed this afternoon, we should case on whether the variable is a single letter to decide whether to wrap in `\text{...}`.

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
```diff
@@ -1,9 +1,56 @@
...
 import type { ODEPlotData, StateVarData } from "../../visualization";
 
+/** Result of simulating an ODE with equations, containing both plot data and LaTeX equations. */
+export type ODEPlotDataWithEquations = {
+    plotData: JsResult<ODEPlotData, string>;
+    latexEquations: string[][];
+};
+
+/** Replace braced UUIDs in equation strings with human-readable labels from the model. */
+function replaceUuidsWithLabels(equation: string, model: DblModel): string {
```
epatters: I think it's unfortunate to make recourse to a find-replace on the LaTeX string at this stage. The mappings from object/morphism UUIDs to names are available in the `DblModel` data structure in catlog-wasm. Thus, I'd suggest doing the transform earlier, specifically in `theories.rs` in catlog-wasm by calling `map_variables` (to replace object UUIDs) and `extend_scalars` (to replace morphism UUIDs). Then this function can be deleted.

I can help with this if it's unclear.

kasbah:
Thanks Evan, I believe I have addressed all feedback.

packages/catlog-wasm/src/theories.rs
```diff
@@ -9,13 +9,43 @@ use wasm_bindgen::prelude::*;
...
+    let ob_name = |id: &QualifiedName| {
+        let name = model
+            .ob_generator_label(id)
+            .map_or_else(|| id.to_string(), |label| label.to_string());
+        let name = if name.chars().count() > 1 {
+            format!("\\text{{{name}}}")
+        } else {
+            name
+        };
+        QualifiedName::single(NameSegment::Text(name.into()))
```
epatters: These are human-readable labels, whereas `QualifedName` is for IDs. (`QualifiedName` should perhaps be called `QualifiedId`, but is in any case supposed to contrast with `QualifiedLabel`.)

I'd just return a `String` here and in `mor_name` below.

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
```diff
@@ -19,25 +55,55 @@ export function createModelODEPlot(
...
+        { equals: false },
+    );
+}
 
-            const simulationResult = simulate(model);
-            if (simulationResult?.tag !== "Ok") {
-                return simulationResult;
+/** Reactively simulate an ODE with equations derived from a model.
+
+Returns both plot data and LaTeX equations with UUIDs replaced by human-readable labels.
```
epatters: I guess this docstring is now slightly misleading since the replacement is no longer being done here.

kasbah:
I fixed equations with params substituted not showing up. I had broken that when switching to wasm uuid replacment.

PR #964, opened 2026-01-27
Add file size limit to `new_ref` backend request

### LINKED ISSUE #408: Enforce filesize limit on the backend
In the JSON i/o PR #332 we're limiting users to uploading up to 5MB of JSON. Jason observed that we ought to have the same constant limiting storage per row in the backend.

patrickaldis:
Fixes #408 . This adds a check in the backend that ensures that uploaded JSON doesn't exceed 5MB

This can be verified by running:
```bash
curl -X POST http://localhost:xxxx/rpc \  
                                           -H "Content-Type: application/json" 
\
                                           -d @/xxxx/payload.json
```

Where `payload.json` is the following 10MB json file:
[payload.json](https://github.com/user-attachments/files/24887513/payload.json)

### DIFF
--- a/packages/frontend/src/api/document_rpc.test.ts
+++ b/packages/frontend/src/api/document_rpc.test.ts
@@ -62,6 +62,29 @@ describe("RPC for Automerge documents", async () => {
         assert.strictEqual(invalidResult.code, 400);
     });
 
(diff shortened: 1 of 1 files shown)


jmoggr:
This looks good, thanks!

I added an integration test to check that the limit is obeyed.

PR #970, opened 2026-01-29
Add sudo password to catcolab account

jmoggr:
In CI we deploy using the `root` user, when deploying from local machines we deploy using the `catcolab` user with an interactive prompt for the new sudo password.


kasbah:
LGTM pending successful merge.

PR #971, opened 2026-01-29
Consolidate ssh key management into single file

jmoggr:
Previously we frequently misunderstood and made mistakes about which public keys needed to be defined in which places. This new declarative structure for the public keys should make it much easier to understand what keys are needed where and make adding/editing keys/hosts less error prone.

Stacked on top of #970 due to skill issues with git.

### DIFF
--- a/infrastructure/hosts/catcolab-next/default.nix
+++ b/infrastructure/hosts/catcolab-next/default.nix
@@ -5,11 +5,7 @@
   ...
 }:
 let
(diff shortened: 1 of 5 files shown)


epatters:
Thanks! I'll let Kaspar review but I have indeed previously been tripped up by this.

kasbah:
Oops maybe shouldn't have merged due to base not being main but we can still merge #970 all together. Just makes it more annoying to review.

Issue #881, opened 2025-11-24
Synchronizing client state beyond single documents


jmoggr:
**EDIT**: "application state" should be read as "meta-document state". The label "application state" was easily misinterpreted (what isn't application state?)

### Background

Recent features have expanded the focus of the client from one document to multiple related documents. As a result there is now a bigger divide between the two different kinds of state that need to by synchronized between the client and the server. I will call these "document state" and "application state". Document state is the contents of individual CatColab documents, synchronization of document state is handled for us by Automerge. Application state is the meta information about documents and how they are related, this includes things like deletion status, permissions, and related documents (parent, descendants). Part of this application state is stored as a relational model in the database, other information, like document relationships, requires scanning the contents of all documents.

### Problem Statement

There currently is no formal strategy for keeping the application state on the client in sync with the server. The current approach is for the client to reload the relevant application state whenever it requests a change to the application state. Without a systematic approach for tracking which operations affect which parts of the state, the code handling application state on the client will devolve into reactive spaghetti.

tl;dr: synchronizing some authorized subset of a CRUD app with a client is hard.

#### 1: Centralize client access and updates of application state

There is a common pattern in web development, popularized by [Redux](https://redux.js.org/) in ReactJS, that centralizes state management and decouples events from how they are handled.

LLM summary of the pattern: A unidirectional data-flow architecture where a single centralized state store is updated through immutable, pure-function reducers in response to dispatched actions.

key terms: "Unidirectional Data Flow", "Flux Pattern"

I'm not precisely sure how we'd implement this with SolidJS in CatColab, but the pattern is flexible and not very complicated.

#### 2: Use an authoritative synchronization engine for application state

CRDTs (the synchronization method used by automerge) lend themselves to local-first "decentralized" applications, these properties difficult to reconcile with permissions and restricting access to data. There is an alternative to CRDTs which trade off the "decentralized" aspects of CRDTs for an authoritative central server. 

There is only serious project I know of which provides this functionality for web applications: https://zero.rocicorp.dev/ It is usable, however it is in Beta. I have not used it enough to vouch for it (I am cursed with the ability to write my own poor imitations), but I've followed it's development for a number of years and am somewhat hyped about it.

It would essentially provides an access layer in front of the DB on the client, performing Automerge-like magic to sync whichever subset of the application data the client is currently using, while enabling authorization on the server.

I would not recommend rolling our own "authoritative synchronization engine for application state".

#### Singleton automerge doc for application state

[...]

PR #976, opened 2026-01-30
Use Automerge doc for user state

kasbah:
This is ready for review but I suggest we wait for "Performance regression on main" (https://github.com/automerge/automerge/issues/1159) to be resolved before deploying to production. As it stands it takes 100ms to create a new user state when the user first visits the site which manifest in a delay of seeing any documents if cold visiting `/documents`.

Closes the "server side state" part of #881 by implementing Automerge docs, derived from the db data, for every user state. Mutations are still handled through regular RPC calls. The `UserState` mainly holds the document lists and some user information and is put into a `Context` on the frontend. 

I couldn't resist slightly improving the UI of "My documents" and "Trash" since I had to replace them anyway. The overall feeling of reactivity of the document titles and the user names is very nice.

image

---

I am not sure how useful it is, but inspired by Benjamin Pierce's Colloquium talk, I made a commutative diagram of the paths I used property based testing on. 

image

|| | |
|--| --|--|
| $$t_1$$ | `write_user_state_to_db`| Writes a `UserState` to the database (testing only). |
| $$f_1$$ | `read_user_state_from_db` | Reads a `UserState` from the database. |
| $$f_2$$ | `run_user_state_subscription` | Listens for Postgres notifications related to user state changes. |
| $$g$$ | `user_state_to_automerge`| Converts a `UserState` into an Automerge document. |
| $$t_2$$ | `automerge_to_user_state` | Converts an Automerge document to a `UserState` (testing only). |

Note that `run_user_state_subscription` will be fine in normal usage but if you e.g. use `ALTER TABLE` by loading a dump or something the automerge doc will be inconsistent with the DB. `read_user_state_from_db` is run on request at the start so re-starting the server will fix the inconsistencies.

Here is a very imperfect schema  for the `UserState`. (Imperfect since we can't really represent `Vec<T>`  or `Option<T>` where `T` is an entity in the schema on CatColab -- also papering over how things are referred to, directly or via ID.)

https://catcolab.org/model/019cae54-2940-7971-9060-dcd2b8657ff5/analysis/019cae83-2b0a-7062-85a8-ceb4d1665c46

[UserState - ERD(4)](https://catcolab.org/model/019cae54-2940-7971-9060-dcd2b8657ff5/analysis/019cae83-2b0a-7062-85a8-ceb4d1665c46)

### DIFF
--- a/packages/frontend/src/App.tsx
+++ b/packages/frontend/src/App.tsx
@@ -50,15 +50,15 @@ const Root = (props: RouteSectionProps) => {
     const models = createModelLibraryWithApi(api, theories);
 
     return (
-        <MultiProvider
-            values={[
-                [ApiContext, api],
-                [TheoryLibraryContext, theories],
-                [ModelLibraryContext, models],
-                UserStateProvider,
-            ]}
-        >
-            <FirebaseProvider app={firebaseApp}>
+        <FirebaseProvider app={firebaseApp}>
+            <MultiProvider
+                values={[
+                    [ApiContext, api],
+                    [TheoryLibraryContext, theories],
+                    [ModelLibraryContext, models],
+                    UserStateProvider,
+                ]}
+            >
                 <MetaProvider>
                     <Title>{import.meta.env.VITE_APP_TITLE}</Title>
                     <ErrorBoundary fallback={(err) => <ErrorBoundaryDialog error={err} />}>
@@ -68,8 +68,8 @@ const Root = (props: RouteSectionProps) => {
                         <SessionExpiredModal />
                     </Show>
                 </MetaProvider>
-            </FirebaseProvider>
-        </MultiProvider>
+            </MultiProvider>
+        </FirebaseProvider>
     );
 };
 
--- a/packages/frontend/src/user/user_state_provider.tsx
+++ b/packages/frontend/src/user/user_state_provider.tsx
@@ -1,6 +1,8 @@
 import type { DocHandle, DocumentId } from "@automerge/automerge-repo";
 import type { UserState } from "catcolab-api/src/user_state";
-import { type JSX, onMount } from "solid-js";
+import { getAuth, onAuthStateChanged } from "firebase/auth";
+import { useFirebaseApp } from "solid-firebase";
+import { type JSX, onCleanup } from "solid-js";
 import { createStore, reconcile } from "solid-js/store";
 
 import { useApi } from "../api";
@@ -10,16 +12,54 @@ import { INITIAL_USER_STATE, UserStateContext } from "./user_state_context";
 
 export function UserStateProvider(props: { children: JSX.Element }) {
     const api = useApi();
+    const firebaseApp = useFirebaseApp();
     const [userState, setUserState] = createStore<UserState>(INITIAL_USER_STATE);
 
-    onMount(async () => {
+    let currentDocHandle: DocHandle<UserState> | null = null;
+    let currentChangeHandler: ((arg: { doc: UserState }) => void) | null = null;
+    let currentUserId: string | null = null;
+
+    const teardownDocHandle = () => {
+        if (currentDocHandle && currentChangeHandler) {
+            currentDocHandle.off("change", currentChangeHandler);
+        }
+        currentDocHandle = null;
+        currentChangeHandler = null;
+    };
+
+    // This will initialize on first load and re-initialize on logout/login
+    const unsubscribeAuth = onAuthStateChanged(getAuth(firebaseApp), async (user) => {
+        const userId = user?.uid ?? null;
+        currentUserId = userId;
+
+        teardownDocHandle();
+        setUserState(INITIAL_USER_STATE);
+
         const userStateDocId = unwrap(await api.rpc.get_user_state_doc_id.query());
+        if (currentUserId !== userId) {
+            return;
+        }
+
         const docHandle: DocHandle<UserState> = await api.repo.find(userStateDocId as DocumentId);
-        setUserState(reconcile(normalizeImmutableStrings(docHandle.doc())));
-        docHandle.on("change", ({ doc }) => {
+        if (currentUserId !== userId) {
+            return;
+        }
+
+        currentDocHandle = docHandle;
+        const onChange = ({ doc }: { doc: UserState }) => {
             setUserState(reconcile(normalizeImmutableStrings(doc)));
-        });
+        };
+        currentChangeHandler = onChange;
+
+        setUserState(reconcile(normalizeImmutableStrings(docHandle.doc())));
+        docHandle.on("change", onChange);
     });
+
+    onCleanup(() => {
+        unsubscribeAuth();
+        teardownDocHandle();
+    });
+
     return (
         <UserStateContext.Provider value={userState}>{props.children}</UserStateContext.Provider>
     );


kasbah:
I just thought of something that is not handled properly yet: when we restart the server all the user state docs ~are lost and Samod / automerge will respond with an error to any requests for previous user state docs~ still exist in the Samod `Repo` but we won't be updating them. The client will keep the doc URL for the non-existing doc until the user refreshes the page manually. Not sure how best to handle this.

packages/backend/src/main.rs
```diff
@@ -133,7 +134,17 @@ async fn main() {
             // Notify systemd we're ready
             sd_notify::notify(false, &[sd_notify::NotifyState::Ready]).ok();
 
-            run_web_server(state.clone(), firebase_auth.clone()).await.unwrap();
+            let web_server = run_web_server(state.clone(), firebase_auth.clone());
+            let subscription = user_state_subscription::run_user_state_subscription(state.clone());
+
```
jmoggr: we should move 

```
// Notify systemd we're ready
sd_notify::notify(false, &[sd_notify::NotifyState::Ready]).ok();
```

to be right before we start the tokio polling loop. It should always be the last thing we do just in case something goes wrong. It's technically fine where it is, but it is a bit of code smell.

kasbah: The "tokio polling loop" is `tokio::select!` in this case?

kasbah:
I decided to persist to a `user_state_urls` table in the DB ([updated schema ERD](https://raw.githubusercontent.com/ToposInstitute/CatColab/42d977a8ce73bcef37ab0553565eb5b76ed23c77/packages/backend/src/schema.svg)) but still  completely  repopulate the doc on startup while keeping the URL stable.

.github/workflows/ci.yml
```diff
@@ -27,8 +27,8 @@ jobs:
           rustup toolchain install ${{ matrix.toolchain }}
           rustup default ${{ matrix.toolchain }}
 
-          cargo build --all-features --verbose
-          cargo test --all-features --verbose
+          cargo check --all-features --verbose
+          cargo test --verbose
```
epatters: What is the effect of removing `--all-features` here? I ask because `catlog` has features and we want to run all its tests, even the feature-gated ones.

.github/workflows/ci.yml
```diff
@@ -177,6 +177,41 @@ jobs:
...
+        run: |
+          cd packages/backend
+          cp .env.development .env && cp .env.development ../migrator/.env
+          unset INVOCATION_ID
+          cargo run -p migrator apply
+
+      - name: Run backend tests
+        run: |
+          cd packages/backend
+          cargo test --all-features
```
epatters: Oh, it must be because the backend now has tests that require a live DB.

Perhaps the solution is to rename `rust_tests` to `catlog_tests` and restrict that action to the packages `catlog`, `catlog-wasm`, and `notebook-types` but still pass `--all-features`.

packages/backend/src/app.rs
```diff
@@ -32,20 +34,6 @@ pub struct AppCtx {
...
-    /// The total number of items matching the query criteria.
-    pub total: i32,
-
-    /// The number of items skipped.
-    pub offset: i32,
-
-    /// The items in the current page.
-    pub items: Vec<T>,
-}
-
```
epatters, line 54: Nice to be liberated from the sad world of paginating the results of DB queries!

packages/backend/src/schema.svg
epatters, line 1: There is already a `schema.svg` in `packages/package/`. We should be replacing that file rather than creating a new one in `src`, right?

packages/backend/src/user_state.rs
```diff
@@ -0,0 +1,863 @@
...
+                    )
+                })?;
+                Ok(Some(dt))
+            }
+            None => Ok(None),
+        }
+    }
+}
+
+mod permission_level_text {
```
epatters: Granted that I don't actually know how autosurgeon works, I was surprised to see these functions defined here because `Hydrate` and `Reconcile` were already derived for `PermissionLevel` in the `auth` module. Is this code redundant?

packages/backend/src/user_state.rs
```diff
@@ -0,0 +1,863 @@
...
+            "model" => Ok(DocumentType::Model),
+            "diagram" => Ok(DocumentType::Diagram),
+            "analysis" => Ok(DocumentType::Analysis),
+            other => Err(format!("unknown document type: {other}")),
+        }
+    }
+}
+
+/// Autosurgeon serialization of `DateTime<Utc>` as milliseconds since Unix epoch.
+mod datetime_millis {
```
epatters: Not too important, but given the side of this module, it might be helpful to move the helpers `datetime_millis` and `option_datetime_millis` to their own module.

packages/backend/src/user_state.rs
```diff
@@ -0,0 +1,863 @@
...
+        SELECT
+            refs.id AS "ref_id!",
+            snapshots.content->>'name' AS name,
+            snapshots.content->>'type' AS type_name,
+            snapshots.content->>'theory' AS theory,
+            refs.created AS "created_at!",
+            refs.deleted_at,
+            (COALESCE(
+                snapshots.content->'diagramIn'->>'_id',
+                snapshots.content->'analysisOf'->>'_id'
```
epatters: In the past, I've pushed hard to prevent these relationships from being hard-coded deep in the backend query logic because a future person innocently creating new document types or relations will have to know to look here and update if appropriate.

I'm going to push again :) I believe another strategy is possible.

In [`api.rs`](https://github.com/ToposInstitute/CatColab/blob/fb3e3fbb7dd28dc0934e95fe986fe0b920ab3504/packages/notebook-types/src/v0/api.rs) in `notebook-types`, I write this about `StableRef` and `Link`:

> Such a reference identifies a specific document, possibly at a specific version. The keys are prefixed with an underscore, e.g. `_id` instead of `id`, to avoid conflicts with other keys and unambiguously signal that the data occur at the *database* level, rather than merely the *document* level. The same convention is used in document databases like CouchDB and MongoDB.

With this in mind, rather than hard-coding `diagramIn` and `analysisOf`, I suggest scanning the document fields (at least the top-level fields, possibly even all fields recursively) for all occurrences of objects with `_id` and `type` fields. These are links/relations. Then, instead of recording only "parent" relations, record *all* relations along with their types. The types can be then used in the frontend to decide as appropriate what if anything is the document parent. As a result, all logic about parent/child relationships will be localized to the frontend.

packages/migrator/src/migrations/m20260225000000_user_state_subscriptions.rs
```diff
@@ -0,0 +1,358 @@
...
+                -- Get ref metadata from head snapshot
+                SELECT
+                    snapshots.content->>'name',
+                    snapshots.content->>'type',
+                    snapshots.content->>'theory',
+                    refs.created,
+                    refs.deleted_at,
+                    COALESCE(
+                        snapshots.content->'diagramIn'->>'_id',
+                        snapshots.content->'analysisOf'->>'_id'
```
epatters: Ah, so here's another place where these creep in. All the more reason to tweak the design.

packages/backend/src/user_state.rs
```diff
@@ -0,0 +1,863 @@
...
+            let permissions: Vec<PermissionInfo> =
+                row.permissions.0.iter().filter_map(|p| p.to_permission_info()).collect();
+
+            let info = DocInfo {
+                name: Text::from(row.name.unwrap_or_else(|| "untitled".to_string())),
+                type_name: row
+                    .type_name
+                    .expect("type_name should never be null")
+                    .parse()
+                    .expect("type_name should be a valid document type"),
```
epatters: We can't guarantee that `type_name` will be a valid document type, can we? In that case, we shouldn't panic the whole server when it's invalid.

epatters (changes requested):
Thanks Kasper, the new design is exciting and this looks very well done.

I've left some initial comments on the backend piece. Haven't look at the frontend changes yet. Most comments are small but I do raise one design issue.

packages/backend/src/user_state.rs
kasbah: Yes, sort of, it was left over from when I was considerably more confused about how I should handle `String`  (i.e. `ImmutableString`) vs `Text` in automerge. I cleaned it up with f71669ffab.

packages/backend/src/user_state.rs
kasbah: 52d6b461e760ac10bf326e25816851a7b01822c8

packages/backend/src/user_state.rs
kasbah: Makes sense, I have gone for calling these `depends_on` and `used_by` because I am not clear on who would be a parent and who a child in an instantiation relation. We don't yet do anything with the instantiation on the frontend but they are there now. 13c58ff50312452a1f1f533663dba85306bb1d40

packages/backend/src/main.rs
kasbah: https://github.com/ToposInstitute/CatColab/pull/976/changes/c295ec11a42b1c5f9ff69c7e6a4a9cc2eb05bf81

kasbah:
I realized it should just be a `state_doc_id` column in the `users` table. I also refactored to call it "state doc ID" everywhere instead of "state doc URL".

epatters:
Great, I'll do another round of review soon.

jmoggr:
This looks good, I have 2 PRs with some suggestions: #1130 and #1129 (which is stacked on top of #1130, but it would be easy to rebase if we don't like #1130).

kasbah:
I have been using this as a base for undo/redo and I am still not happy with the performance of this since adding public docs. I think we need to figure out a way this can go from 100s of docs to 1000s of docs without affecting performance in the future. 

Since we are not using the public docs at the moment I have removed them for now and will leave performance optimization for another PR. I have reverted including public docs: [4fe9d16](https://github.com/ToposInstitute/CatColab/pull/976/commits/4fe9d1637ab154a9dd7f468008a9a044451e6e3d)

kasbah:
In happy news Automerge, Autosurgeon and Samod have all gotten a release with the performance improvements. :rocket: :tada: I've upgraded them here.

packages/backend/src/user_state.rs
```diff
@@ -0,0 +1,887 @@
...
+    /// A diagram document.
+    #[autosurgeon(rename = "diagram")]
+    Diagram,
+    /// An analysis document.
+    #[autosurgeon(rename = "analysis")]
+    Analysis,
+    /// A document whose type is not recognized by this version of the backend.
+    #[autosurgeon(rename = "unknown")]
+    Unknown,
+}
```
epatters: 1. What benefit do we get by making this an enum rather than just a string in the backend?
2. If we do need this extra typing, I'd rather have it defined in and imported from `notebook-types` so that we have a single source for the document types. As it stands, we have the `Document` enum in `notebook-types` and `DocInfoType` here.

packages/backend/src/user_state.rs
```diff
@@ -0,0 +1,887 @@
...
+impl DbPermission {
+    /// Convert this database permission entry into a [`PermissionInfo`].
+    pub fn to_permission_info(&self) -> Option<PermissionInfo> {
+        let level = match self.level.as_str() {
+            "read" => PermissionLevel::Read,
+            "write" => PermissionLevel::Write,
+            "maintain" => PermissionLevel::Maintain,
+            "own" => PermissionLevel::Own,
+            _ => return None,
+        };
```
epatters: IIRC, at one point, I had figured out how to make `sqlx` automatically convert the string into a `PermissionLevel`, so that you could remove this struct `DbPermission` and use `PermissionInfo` directly.

For comparison, here's how to the do conversion for `Uuid`: https://github.com/launchbadge/sqlx/discussions/1683 I believe the same strategy should work here.

packages/backend/src/user_state.rs
```diff
@@ -0,0 +1,887 @@
...
+}
+
+impl DbRelation {
+    /// Convert this database relation entry into a [`RelationInfo`].
+    pub fn to_relation_info(&self) -> Option<RelationInfo> {
+        let ref_id = uuid::Uuid::parse_str(&self.ref_id).ok()?;
+        Some(RelationInfo {
+            ref_id,
+            relation_type: self.relation_type.clone(),
+        })
```
epatters: Same question here: looks like you could get a `Uuid` directly from `sqlx`? https://github.com/launchbadge/sqlx/discussions/1683

packages/backend/src/user_state.rs
```diff
@@ -0,0 +1,887 @@
...
+        }
+        _ => {}
+    }
+}
+
+/// Reads user state from the database.
+pub async fn read_user_state_from_db(user_id: String, db: &PgPool) -> Result<UserState, AppError> {
+    debug!(user_id = %user_id, "Reading user state from database");
+
+    let query_start = std::time::Instant::now();
```
epatters: Did you intend to leave this profiling code in place?

packages/backend/src/user_state.rs
```diff
@@ -0,0 +1,887 @@
...
+            for child in arr {
+                collect_relations(child, out);
+            }
+        }
+        _ => {}
+    }
+}
+
+/// Reads user state from the database.
+pub async fn read_user_state_from_db(user_id: String, db: &PgPool) -> Result<UserState, AppError> {
```
epatters, line 213: There's a lot of code spread across `user_state` and `user_state_updates` and I'm having trouble seeing the forest for the trees. That said, I'm wondering whether there is some redundancy across the two modules. E.g., this function and `build_doc_info_from_db` in `user_state_updates` seem to have considerable overlap (and both are pretty hefty functions). Is it possible/desirable to consolidate?

packages/backend/src/user_state_updates.rs
```diff
@@ -0,0 +1,734 @@
...
+
+use crate::app::{AppError, AppState};
+use crate::user_state::{
+    DEFAULT_DOC_NAME, DbPermission, DocInfo, DocInfoType, PermissionInfo, RelationInfo, UserInfo,
+    UserState, extract_relations_from_json, reconcile_user_state,
+};
+
+// ---------------------------------------------------------------------------
+// Direct user-state mutation helpers (called from RPC handlers)
+// ---------------------------------------------------------------------------
```
epatters: This section header says the same thing as the module-level doc comment, and I don't see any other sections in this file. Remove it?

packages/ui-components/src/document_type_icon.tsx
```diff
@@ -6,7 +6,7 @@ import { Match, Switch } from "solid-js";
 
 import { ModelFileIcon } from "./model_file_icon";
 
-export type DocumentType = "model" | "diagram" | "analysis";
+export type DocumentType = "model" | "diagram" | "analysis" | "unknown";
```
epatters: This will be the *third* place where the set of document types is defined.

packages/frontend/src/util/virtual_list.ts
```diff
@@ -0,0 +1,53 @@
+import { type Accessor, createMemo, createSignal } from "solid-js";
+
+/**
+ * A reactive virtual list utility.
+ *
+ * Unlike `@solid-primitives/virtual`'s `createVirtualList`, this properly
+ * tracks all accessor inputs (items, rootHeight, rowHeight) inside a memo so
+ * the computed output updates when any of them change.
+ */
```
epatters, line 7: I guess this code is adapted from that package? If so, I think it's good form to explicitly acknowledge that.

epatters (commented):
Thanks Kaspar. There's a lot to process here. I've left some questions/comments focusing on the backend piece. I'll try to make another pass soon.

My overall impression is that I'm glad the complexity has shifted from Postgres queries/notification log to Rust code, because I think the Rust code will be easier to maintain and involves less magic. That said, the code footprint on the backend is quite large and the complexity of adding new features to backend in the future will increase. So I'm really hoping that this is worth it on the frontend side, both in terms of dev experience and user experience.

jmoggr:
Are the performance concerns mentioned here:

> Since we are not using the public docs at the moment I have removed them for now and will leave performance optimization for another PR. 

Addressed by this?:

> In happy news Automerge, Autosurgeon and Samod have all gotten a release with the performance improvements. 🚀 🎉 I've upgraded them here.

I think benefits of the user-state-doc decrease if we need a parallel system for public docs. I'd be happy to help out with that or explore feasibility.

packages/backend/src/auth.rs
```diff
@@ -242,6 +275,11 @@ pub async fn set_permissions(
     insert_query.execute(&mut *transaction).await?;
 
     transaction.commit().await?;
+
+    if let Err(e) = update_ref_for_users(state, ref_id, old_holders).await {
+        tracing::error!(%ref_id, error = %e, "Failed to update user states after permission change");
+    }
```
epatters, line 281: I'm noticing that, for the first time, this PR handles some `AppError`s by emitting a trace rather than bubbling up the error and failing the RPC call.

I'm not necessarily opposed to that, but it does leave me wondering: by what principles do we decide how to handle errors?

It's not obvious to me why we're handling errors to do with user state updating differently than other errors.

packages/backend/src/user_state.rs
```diff
@@ -0,0 +1,841 @@
...
+
+/// Reconcile a `UserState` into an Automerge document, returning an `AppError` on failure.
+pub fn reconcile_user_state(
+    doc: &mut automerge::Automerge,
+    state: &UserState,
+) -> Result<(), AppError> {
+    doc.transact(|tx| reconcile(tx, state))
+        .map_err(|e| AppError::UserStateSync(format!("Failed to reconcile: {:?}", e)))?;
+    Ok(())
+}
```
epatters: Nit: it's sort of weird to me to put this function at the very top of the file, before the `UserState` struct is defined.

Now that I think of it, perhaps this should just be a method of `UserState`.

packages/frontend/src/util/immutable_string.ts
```diff
@@ -0,0 +1,19 @@
+import { ImmutableString } from "@automerge/automerge";
+
+/** Recursively convert any automerge `ImmutableString` values to native strings. */
+export function normalizeImmutableStrings<T>(value: T): T {
+    if (value instanceof ImmutableString) {
```
epatters: Looks like `automerge-repo` provides `isImmutableString` as the canonical way to do this check.

It's kinda annoying that we have to roll this helper ourselves, but I guess there's no more official way?

packages/frontend/src/util/virtual_list.ts
kasbah: No, it's a separate implementation. I've removed the comment. 44ceec275635e6510ba16d49cb56191a5cc05821

packages/frontend/src/util/immutable_string.ts
kasbah: I don't really mind so I've changed it already but: I did see `isImmutableString` before but wasn't sure why it was better than checking the instance. I know it checks for a `Symbol` tag, is that faster/better in some way? Just curious...

I don't think there ever will be an official way to do this because we are stripping information that would be useful to automerge if this were actually a collaboratively edited document.

packages/frontend/tsconfig.json
```diff
@@ -5,7 +5,7 @@
         "target": "ES2020",
         "useDefineForClassFields": true,
         "module": "ESNext",
-        "lib": ["ES2022", "DOM", "DOM.Iterable"],
+        "lib": ["ES2023", "DOM", "DOM.Iterable"],
```
kasbah, line 8: This is for using `toSorted` in document_list.tsx just FYI

packages/backend/src/auth.rs
kasbah: My thinking on this was if e.g. someone deletes a document and it all works except some *other* user's  state update (whom the doc is shared with) fails for some unknown reason, then it's not correct to present an error to the user that deleted the document. Their delete actually did work and what are they supposed to do about the state updates?

packages/backend/src/user_state.rs
kasbah: 1. We generate TS types from this and the frontend makes use of that. I could also imagine that we use this information in the backend in the future.
2. I was reluctant to add the necessary dependencies to derive traits to notebook-types. I've added them now and feature gated them as `backend`: a4ad17a29b7a60fba381fddab027d04c9bd045a2

kasbah:
@jmoggr 

> Are the performance concerns mentioned here:

>>    Since we are not using the public docs at the moment I have removed them for now and will leave performance optimization for another PR.

> Addressed by this?:

No, they are not.

> I think benefits of the user-state-doc decrease if we need a parallel system for public docs. I'd be happy to help out with that or explore feasibility.

It's not really public vs private it's just the number of docs I believe. We might need different ways of determining "these are the docs the user cares about right now", though with that we are sort of back to something akin to paginating I guess.

packages/backend/src/user_state.rs
kasbah: removed

packages/backend/src/user_state.rs
kasbah: It also seems we need to use `INITCAP` to do this cleanly which is some new SQL for me. 4a4defd1a1bc

Issue #995, opened 2026-02-04
`morLabelOrDefault` no longer working


tim-at-topos:
At some point the default name for arrows in analyses (https://github.com/ToposInstitute/CatColab/pull/549; note that `morNameOrDefault` was later (1268cbd0fc9261eba8290ffff5c41c9ece0a72ed) renamed to `morLabelOrDefault`) broke. Now if there is no name then the behaviour is as before: the row label in the analysis tab is blank instead of `src -> tgt`.

tim-at-topos:
Note that things become confusing with instantiation: should the default name in the below be `RG.(Prey -> Predator)` or `RG.(Grass -> Rabbit)`. @epatters do you have thoughts here?

<img width="1920" height="947" alt="Image" src="https://github.com/user-attachments/assets/7c336b03-0c7c-4699-8883-0b0c0794e97a" />

epatters:
> Note that things become confusing with instantiation: should the default name in the below be `RG.(Prey -> Predator)` or `RG.(Grass -> Rabbit)`?

Of those two possible names, the first one makes sense because `Prey` and `Predator` are names in the namespace of `RG`'s type, whereas the second one does make sense.

However, in this situation, I'd prefer a third possibility, simply `Grass -> Rabbit`. This also makes sense because `Grass` and `Rabbit` are names in the top-level namespace of this model. In fact, this label should be produced automatically because `Grass` is DoubleTT's *canonical* name for the object that also has the name `RG.Prey`. So I'm actually confused about why `morLabelOrDefault` is broken.

Edit: I suspect the reason is that empty label segments are being treated as meaningful rather than as nulls. I'll look into this.

PR #998, opened 2026-02-04
Fix `morLabelOrDefault`

kasbah:
Made a start at fixing #995


epatters:
Superseded by #1010. I think `morLabelOrDefault` is working as intended but there is a bug in how empty labels are handled in `Namespace`.

Issue #997, opened 2026-02-04
Unicode breaks mass-action equations + mystery double \text


tim-at-topos:
Alongside #986 , unicode breaks things! (also there's some `\text{\text` which seems weird).

<img width="805" height="354" alt="Image" src="https://github.com/user-attachments/assets/257e8a02-f0ef-4442-b3e1-2801a2fccf82" />

(From https://next.catcolab.org/model/019b6cf9-5587-7962-aa93-c52a9643e8ac/analysis/019b6cfb-36e6-77f1-a1e5-aa31cc6df9f8)

tim-at-topos:
What I've just noticed, which is bizarre, is that actually the unicode only breaks things *sometimes*... in the above screenshot, many of the subscript numbers work just fine

**Update.** Here's the actual KaTeX error:

```
ParseError: KaTeX parse error: Expected 'EOF', got '_' at position 145: …{v₁}}) \text{in_̲cloud}
```

PR #1000, opened 2026-02-04
Unbalanced mass-action ODE semantics

tim-at-topos:
A first implementation of unbalanced mass-action semantics (in search of a better name).

- [x] working for stock-flow
- [x] working for stock-flow with signed links
- [x] working for Petri nets (with single consumption rate per transition)
- [x] LaTeX equation rendering (within the simulation analysis)
- [x] separate "Unbalanced mass-action dynamics equations" analysis
- [x] write documentation

PR #1001, opened 2026-02-04
Fix double `\text` on LHS of latex equations

kasbah:
Resolves https://github.com/ToposInstitute/CatColab/issues/997


kasbah:
Re-based on #1000

packages/catlog-wasm/src/latex.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    use catlog::simulate::ode::LatexEquation;
+    use catlog::stdlib::analyses::ode::StockFlowMassActionAnalysis;
+    use notebook_types::current::{MorDecl, MorType, Ob, ObDecl, ObType};
+
+    use super::*;
+    use crate::model::DblModel;
+    use crate::theories::ThCategoryLinks;
+
+    fn backward_link_model() -> DblModel {
```
epatters: Reuse the backward link model from `catlog::stdlib` instead of duplicating here.

packages/catlog-wasm/src/latex.rs
epatters, line 1: I remain a little unsure where to put this stuff myself---ultimately, I suspect it should be in `catlog` not `catlog-wasm`---but since these functions are specialized to analyses, for now I'd just put them `analyses.rs` where the struct `ODELatex` is also defined, rather than in a new module called `latex`.

epatters (changes requested):
Thanks Kaspar! Two comments below.

packages/catlog-wasm/src/latex.rs
kasbah: I want this to use longer names though so that they are wrapped in `\text`. I don't think I can easily override the ones in stdlib?

epatters: Good point. However, I realized that this model was already essentially created in the `catlog-wasm` and refactored it to make it available for other tests (#1014). You can rebase off main and use that.

PR #1007, opened 2026-02-05
Factor linear/Lotka-Voltera ODE analyses through `PolynomialSystem`

### LINKED ISSUE #965: Factor linear/Lotka-Voltera ODE analyses through `PolynomialSystem`
Most of the existing ODE simulation analyses are mathematically based on polynomial vector fields; the only exception is the experimental Kuramoto system. However, for historical reasons, only the mass action analysis was implemented using the `PolynomialSystem` data structure.

The hard-coded vector fields for the linear and Lotka-Volterra ODE analyses should be removed and replaced with code that generates suitable `PolynomialSystem`s (linear in one case, quadratic in the other).

An immediate advantage to having a symbolic representation is that we can use @kasbah's new LaTeX pretty-printer (#954) to display the systems of equations for these analyses.

georgefst:
Closes #965.

Developed alongside @patrickaldis.

### DIFF
--- a/packages/catlog/src/stdlib/analyses/ode/linear_ode.rs
+++ b/packages/catlog/src/stdlib/analyses/ode/linear_ode.rs
@@ -7,18 +7,20 @@ use std::collections::HashMap;
 use std::hash::Hash;
 use std::ops::Add;
 
+use indexmap::IndexMap;
 use itertools::Itertools;
 use nalgebra::{DMatrix, DVector};
+use num_traits::Zero;
 
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 #[cfg(feature = "serde-wasm")]
 use tsify::Tsify;
 
-use super::{ODEAnalysis, SignedCoefficientBuilder};
-use crate::dbl::model::DiscreteDblModel;
+use super::{ODEAnalysis, Parameter, SignedCoefficientBuilder};
 use crate::simulate::ode::{NumericalPolynomialSystem, ODEProblem, PolynomialSystem};
 use crate::{
+    dbl::model::DiscreteDblModel,
     one::QualifiedPath,
     zero::{QualifiedName, rig::Monomial},
 };
@@ -51,9 +53,9 @@ pub fn linear_polynomial_system<Var, Coef>(
 ) -> PolynomialSystem<Var, Coef, u8>
 where
     Var: Clone + Hash + Ord,
-    Coef: Clone + Add<Output = Coef>,
+    Coef: Clone + Add<Output = Coef> + Zero,
 {
-    PolynomialSystem {
+    let system = PolynomialSystem {
         components: coefficients
             .row_iter()
             .zip(vars)
@@ -67,7 +69,8 @@ where
                 )
             })
             .collect(),
-    }
+    };
+    system.normalize()
 }
 
 impl SignedCoefficientBuilder<QualifiedName, QualifiedPath> {
@@ -81,7 +84,7 @@ impl SignedCoefficientBuilder<QualifiedName, QualifiedPath> {
         model: &DiscreteDblModel,
         data: LinearODEProblemData,
     ) -> ODEAnalysis<NumericalPolynomialSystem<u8>> {
-        let (matrix, ob_index) = self.build_matrix(model);
+        let (system, ob_index) = self.linear_ode_system(model);
         let n = ob_index.len();
 
         let initial_values = ob_index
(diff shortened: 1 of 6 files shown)


packages/catlog/src/stdlib/analyses/ode/linear_ode.rs
```diff
@@ -47,7 +47,7 @@ impl SignedCoefficientBuilder<QualifiedName, QualifiedPath> {
         &self,
         model: &DiscreteDblModel,
         data: LinearODEProblemData,
-    ) -> ODEAnalysis<LinearODESystem> {
+    ) -> ODEAnalysis<NumericalPolynomialSystem<u8>> {
```
georgefst, line 86: The exponent for linear systems is obviously always 1. So it's tempting to use or define a singleton type, but that might be overkill. Not sure how easy or conventional it is in Rust.

packages/catlog/src/zero/alg.rs
```diff
@@ -183,6 +183,29 @@ where
 // XXX: Lots of boilerplate to delegate the module structure of `Polynomial` to
 // `Combination` because these traits cannot be derived automatically.
 
+impl<Var, Coef, Exp> Sum for Polynomial<Var, Coef, Exp>
```
georgefst, line 186: There's nothing here that's actually specific to `Polynomial`. The exact same code could be used for any `AdditiveMonoid`. But I don't think Rust allows instances with that sort of flexibility. We could define some reusable helper, but probably not worth it.

Note that `Polynomial` is an `AdditiveMonoid` precisely when its coefficient type is, due to the instances:
- `impl<Var, Coef, Exp> Add for Polynomial<Var, Coef, Exp>`
- `impl<Var, Coef, Exp> Zero for Polynomial<Var, Coef, Exp>`

georgefst: Oh, and also, looking at the comment above this, maybe we should actually move the main `Impl` code to `Combination`, and just delegate to that here?

packages/catlog/src/zero/alg.rs
```diff
@@ -210,7 +233,7 @@ where
 impl<Var, Coef, Exp> Zero for Polynomial<Var, Coef, Exp>
 where
     Var: Ord,
-    Coef: Add<Output = Coef> + Zero,
+    Coef: Zero,
```
georgefst, line 237: This slight generalisation was useful during debugging. I'm not sure if it was necessary in the end, but there's no harm.

Coming from Haskell, I am a bit surprised that the compiler doesn't warn about redundant constraints.

packages/catlog/src/stdlib/analyses/ode/linear_ode.rs
patrickaldis: I'm not sure this accounts for constants (zero exponents) though. We'd need a type with 0 and 1

georgefst: Oh yeah, good point 🤦‍♂️

packages/catlog/src/zero/alg.rs
epatters: Yeah, I've found myself wishing for such warnings but AFAIK, the Rust compiler doesn't warn about either redundant constraints or unnecessary constraints.

packages/catlog/src/simulate/ode/lotka_volterra.rs
```diff
@@ -1,45 +1,45 @@
...
-    fn vector_field(&self, dx: &mut DVector<f32>, x: &DVector<f32>, _t: f32) {
-        let A = &self.interaction_coeffs;
-        let b = &self.growth_rates;
-        *dx = (A * x + b).component_mul(x);
+) -> NumericalPolynomialSystem<u8> {
+    NumericalPolynomialSystem {
+        components: interaction_coeffs
+            .row_iter()
+            .enumerate()
+            .zip(growth_rates.into_iter())
```
epatters: To avoid triggering the linter in CI, you can run `cargo clippy` locally before pushing.

georgefst: Thanks. Done.

georgefst:
> I'd suggest the following alternative design

In the last commit, we've refactored so that polynomials with symbolic coefficients are created first, and added some LaTeX tests to exercise the symbolic versions.

Hopefully this is roughly what you had in mind. If so, we'll then proceed with a similar change for the Lotka-Volterra systems.

packages/catlog/src/stdlib/analyses/ode/linear_ode.rs
georgefst: We could of course define a zero-or-one type instead. But still, probably overkill.

packages/catlog/src/stdlib/analyses/ode/linear_ode.rs
```diff
@@ -4,8 +4,11 @@
 //! [`linear_ode_analysis`](SignedCoefficientBuilder::linear_ode_analysis).
```
georgefst, line 4: Ah, I suppose this is no longer true now that this module also contains the new generalised `linear_polynomial_system` as well.

packages/catlog/src/stdlib/analyses/ode/linear_ode.rs
```diff
@@ -92,10 +134,98 @@ mod test {
...
+        let expected = expect![[r#"
+            $$
+            \begin{align*}
+            \frac{\mathrm{d}}{\mathrm{d}t} A &= (-0.3) A\\
+            \frac{\mathrm{d}}{\mathrm{d}t} B &= 0.5 X\\
+            \frac{\mathrm{d}}{\mathrm{d}t} X &= A + (-2) B
+            \end{align*}
+            $$
+            "#]];
+        expected.assert_eq(&matrix_example().to_latex_string());
```
georgefst: Rendered, this produces:

Image

epatters:
Thanks @georgefst and @patrickaldis, this is essentially what I was envisioning! How about you go ahead with the Lotka-Volterra system and then I'll re-review the whole thing?

georgefst:
> Thanks @georgefst and @patrickaldis, this is essentially what I was envisioning! How about you go ahead with the Lotka-Volterra system and then I'll re-review the whole thing?

Great. Yep, working on it now.

epatters (approved):

PR #1019, opened 2026-02-07
Add ELK as alternative layout engine for graph visualization

### LINKED ISSUE #133: Consider using elkjs for graph layout
[elkjs](https://github.com/kieler/elkjs) is a graph layout engine originally built for the Eclipse IDE. The [online examples](https://rtsys.informatik.uni-kiel.de/elklive/examples.html?e=ports%2FportConstraints) are quite impressive, demonstrating support for nested layouts and orthogonal edge routing, among other features. I expect ELK would do much better than Graphviz at computing layouts for directed wiring diagrams.

Like Graphivz, ELK would be a heavy dependency, in this case because the JavaScript is generated from Java using [GWT](https://en.wikipedia.org/wiki/Google_Web_Toolkit), but it doesn't really matter because we could lazy load the lib like we do with [Viz-js](https://github.com/mdaines/viz-js).

epatters:
In the graph visualization analyses, you can now choose between Graphviz and ELK.

Closes #133.

### DIFF
--- a/packages/frontend/src/visualization/elk.ts
+++ b/packages/frontend/src/visualization/elk.ts
@@ -14,7 +14,7 @@ import type * as GraphSpec from "./graph_spec";
 import { measureText } from "./measure";
 import type { ArrowStyle } from "./types";
 
-/** Elk node with extra styling data attached.
+/** ELK node with extra style data attached.
 
 ELK will ignore this extra data and just pass it through.
  */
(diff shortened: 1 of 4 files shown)


packages/frontend/package.json
```diff
@@ -87,6 +88,7 @@
         "vite-plugin-solid": "^2.11.10",
         "vite-plugin-wasm": "^3.5.0",
         "vitest": "^4.0.8",
-        "wasm-pack": "^0.14.0"
+        "wasm-pack": "^0.14.0",
+        "web-worker": "^1.5.0"
```
epatters, line 92: Following https://github.com/kieler/elkjs/issues/142#issuecomment-988478433, I had to add this dev dependency to make Vite dev happy.

packages/frontend/src/visualization/elk.ts
```diff
@@ -0,0 +1,171 @@
...
+    });
+
+    return { id: "root", children, edges, layoutOptions };
+}
+
+const nodePadding = 10;
+
+// XXX: How do we properly set these?
+const defaultFont = "1rem sans-serif";
+const monospaceFont = "1rem monospace";
```
epatters, line 82: This is janky, but I couldn't think of an easy/clean way to get the CSS font property that will actually be used when the SVG is rendered. That said, this actually seems to work, in my experiments yielding more precise node sizes than we can get out of Graphviz (which has an extremely archaic font system).

PR #1028, opened 2026-02-10
Translucent morphisms in DoubleTT elaborator

KevinDCarlson:
This PR removes the `opaque` `TmV` variant entirely, keeping morphism terms visible in the values. This required changing one test that expected a morphism term to evaluate to `opaque`.  I've also corrected a few typos.

### DIFF
--- a/packages/catlog/src/tt/val.rs
+++ b/packages/catlog/src/tt/val.rs
@@ -212,11 +212,7 @@ pub enum TmV_ {
     Meta(MetaVar),
 }
 
-/// Values for terms in the codiscrete mode, dereferences to [TmV_].
-///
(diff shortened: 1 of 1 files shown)


packages/catlog/src/tt/val.rs
epatters: I believe this doc comment needs to be updated.

epatters (changes requested):
Looks good! Thanks for fixing all those typos and mispellings. Just one comment below.

packages/catlog/src/tt/val.rs
KevinDCarlson: Fixed, thanks.

PR #1039, opened 2026-02-12
Embed fonts in SVG exports

kasbah:
This inlines the two Google fonts we use into the SVGs that people can download. Previously the exact fonts would just be decided by the system viewing the downloaded SVG.

### DIFF
--- a/packages/frontend/src/visualization/export_svg_button.tsx
+++ b/packages/frontend/src/visualization/export_svg_button.tsx
@@ -2,17 +2,19 @@ import Download from "lucide-solid/icons/download";
 import type { JSX } from "solid-js";
 
 import { IconButton } from "catcolab-ui-components";
-import { downloadSVG } from "./export_svg";
 
-/** Button to download an SVG. */
+/** Button to download an SVG with embedded fonts. */
 export function DownloadSVGButton(props: {
     svg?: SVGSVGElement;
     filename?: string;
     tooltip?: JSX.Element | string;
     size?: number;
 }) {
-    const download = () => {
-        props.svg && downloadSVG(props.svg, props.filename ?? "export.svg");
+    const download = async () => {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/util/google_font_inliner.ts
```diff
@@ -0,0 +1,78 @@
...
+            });
+
+            return Promise.all(fontLoadedPromises).then(() => {
+                return css;
+            });
+        });
+    }
+}
+
+export default GoogleFontInliner;
```
kasbah, line 78: I vendored this because it's a single file and didn't originally come with type definitions.

epatters: Copying in this dependency makes sense to me.

packages/frontend/src/visualization/export_svg.ts
```diff
@@ -56,3 +68,46 @@ function recurseComputedStyleToInlineStyle(element: StylableElement, cloned: Sty
...
+
+        if (monoFontName && monoFontName !== mainFontName) {
+            const monoInliner = new GoogleFontInliner(monoFontName);
+            const monoCss = await monoInliner.style();
+            fontStyles.push(monoCss);
+        }
+
+        return fontStyles.join("\n\n");
+    } catch (error) {
+        console.error("Could not extract font-face definitions from CSS custom properties:", error);
```
epatters: What's the thinking behind wrapping this in a big try-catch? Is this just defensive programming or do you have specific reason to think that this can fail?

kasbah: Mainly defensive but Google Fonts could be blocked or something.

kasbah: But you are right, even the error message is not right. I'll change it.

PR #1040, opened 2026-02-12
Configure separation parameter in undirected graph layouts

### LINKED ISSUE #709: Configure separation parameter in undirected graph layouts
The Graphviz algorithm for undirected graph layouts is very sensitive to the choice of the [`sep`](https://graphviz.org/docs/attrs/sep/) parameter. We should make it configurable in the UI. Currently, it is [hard-coded](https://github.com/ToposInstitute/CatColab/blob/271660eabc429f6c8da2a805dc55afa703f38192/packages/frontend/src/stdlib/analyses/graph_visualization.tsx#L115).

tibisabau:
## Description

Added configurable `sep` and `overlap` parameters for undirected graph layouts using Graphviz's neato algorithm. These parameters were previously hard-coded (using Graphviz defaults), making it difficult to achieve acceptable layout results for certain graphs. The changes expose these parameters in the UI through input fields that appear when the undirected layout engine is selected.

The `sep` parameter controls the separation between nodes, while `overlap` controls how node overlaps are handled. Both parameters are optional and default to Graphviz's standard behavior when left empty.

Closes #709.

## Checklist if Applicable

- [x] The tests passed – `pnpm --filter ./packages/frontend run test` 
- [x] Linting passed – `pnpm --filter ./packages/frontend run lint` (no issues found)
- [ ] Documentation has been added
- [ ] CHANGELOG.md has been updated

### DIFF
--- a/packages/frontend/src/visualization/graph_layout_config.ts
+++ b/packages/frontend/src/visualization/graph_layout_config.ts
@@ -9,11 +9,19 @@ export type Config = {
     /** Primary layout direction, when applicable. */
     direction?: Direction;
 
-    /** Separation parameter for undirected graph layout (Graphviz neato). */
-    sep?: number;
+    /** Separation margin for undirected graph layout (Graphviz neato), in points.
+     * Can be:
+     * - Single value: "1.0" (uniform margin)
+     * - Point value: "0.5,1.0" (separate X and Y margins)
+     * - Absolute: "+1.0" or "+0.5,1.0" (not scaled by node size)
+     * Default: "0.25"
+     */
+    sep?: string;
 
-    /** Overlap parameter for undirected graph layout (Graphviz neato). */
-    overlap?: Overlap;
+    /** Overlap removal algorithm for undirected graph layout (Graphviz neato).
+     * Default: OverlapRemoval.Prism ("false")
+     */
+    overlap?: OverlapRemoval;
 };
 
 /** Engines supported for graph layout. */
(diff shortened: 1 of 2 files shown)


packages/frontend/src/visualization/graph_layout_config.ts
```diff
@@ -8,6 +8,12 @@ export type Config = {
 
     /** Primary layout direction, when applicable. */
     direction?: Direction;
+
+    /** Separation parameter for undirected graph layout (Graphviz neato). */
+    sep?: number;
+
+    /** Overlap parameter for undirected graph layout (Graphviz neato). */
+    overlap?: string;
```
kasbah: Is there an enum from Graphviz for this?

kasbah:
Were you able to try it out yourself?

Some things I noticed:

1. It seems like "Prism algorithm" and "Remove overlaps" result in the same visualization. Are we sure these do different things?
2. Separation doesn't have any effect when we "Allow overlaps", can we easily remove it when no separation is set or does it complicate the code too much?

packages/frontend/src/visualization/graph_layout_config.ts
```diff
@@ -26,18 +32,37 @@ export enum Direction {
...
+    /** Remove overlaps (Graphviz default). */
+    False = "false",
+    /** Scale layout uniformly to remove overlaps. */
+    Scale = "scale",
+    /** Scale layout separately in x and y to remove overlaps. */
+    ScaleXY = "scalexy",
+    /** Allow overlaps. */
+    True = "true",
+    /** Prism algorithm for overlap removal. */
+    Prism = "prism",
```
kasbah: My quick research suggests we are not enumerating all the options here and we have at least one redundant one (`prism` as I suggested might be happening in my last comment).

```
| --------------------- | --------------------------------------------------- |
| true                  | (default) No overlap removal                        |
| false                 | Use Prism algorithm to remove overlaps              |
| "scale"               | Scale the layout to remove overlaps                 |
| "scalexy"             | Scale X and Y separately to remove overlaps         |
| "orthoxy" / "orthoyx" | Ortho-based overlap removal (X then Y, or Y then X) |
| "ortho"               | Ortho-based overlap removal                         |
| "voronoi"             | Use Voronoi-based technique                         |
| "compress"            | Compress layout after overlap removal               |
| "vpsc"                | Use quadratic optimization (if available)           |
| "prism" / "prismN"    | Proximity graph-based (N = number of attempts)      |
| "ipsep"               | Use mode="ipsep" (for neato/fdp)                    |
```

packages/frontend/src/visualization/graph_layout_config.ts
```diff
@@ -8,6 +8,12 @@ export type Config = {
 
     /** Primary layout direction, when applicable. */
     direction?: Direction;
+
+    /** Separation parameter for undirected graph layout (Graphviz neato). */
+    sep?: number;
```
kasbah: My quick research suggests this is not just a number.

> The sep Graphviz graph attribute specifies a margin to place around nodes when removing node overlaps (used with overlap removal). Its value is an addDouble or addPoint:
> - A single number (e.g., "4") — uniform margin in points added around each node
> - A point value (e.g., "4,8") — separate X and Y margins in points
> - Prefixed with + (e.g., "+4" or "+4,8") — the margin is added as an absolute value in points rather than being scaled by the node size

kasbah:
Hi, I am closing this and your other PR (#1041) because only responding with code is not very helpful. I can't tell if you have thought about the problems at all. 

You haven't really answered any of my questions. Feel free to re-open PRs if you are willing to respond to my questions.

PR #1041, opened 2026-02-12
Add plus button to instantiation cells

### LINKED ISSUE #891: Not obvious how to make multiple assignments in an instantiation
To assign multiple variables in an instantiation cell, as far as I can tell the only way to introduce a new assignment line is by hitting "Return" from a previous assignment line. This isn't very discoverable. Seems like there should be a clickable "+" for this, though that might be a bit busy.

Image

tibisabau:
## Description

This change improves the discoverability of adding new assignment lines in instantiation cells by introducing a visible "+" button. Previously, users could only add new assignments by pressing "Return" from an existing assignment line, which was not intuitive. The implementation adds an IconButton with a Plus icon below the specializations list that, when clicked, adds a new assignment at the end and automatically focuses it.

The solution follows the existing UI patterns in the codebase by:
- Using the `Plus` icon from `lucide-solid/icons/plus`
- Using the `IconButton` component from `catcolab-ui-components`
- Providing a tooltip "Add new assignment" for discoverability
- Using consistent styling aligned with the left margin of the specializations list

Closes #891.

## Checklist if Applicable

- [x] The tests passed – `pnpm --filter ./packages/frontend run test`
- [x] Linting passed – `pnpm --filter ./packages/frontend run lint` 
- [ ] Documentation has been added
- [ ] CHANGELOG.md has been updated

### DIFF
--- a/packages/frontend/src/model/instantiation_cell_editor.css
+++ b/packages/frontend/src/model/instantiation_cell_editor.css
@@ -31,3 +31,8 @@
         color: var(--color-gray-750);
     }
 }
+
+.add-specialization-button {
+    margin-left: 1.5rem;
+    margin-top: 0.25rem;
+}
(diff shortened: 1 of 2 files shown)


kasbah:
Closing, see https://github.com/ToposInstitute/CatColab/pull/1040#issuecomment-3933325079

Issue #1035, opened 2026-02-11
Merge mass-action and unbalanced mass-action


tim-at-topos:
Following #1000, it makes sense to combine the unbalanced mass-action analysis with the mass-action one, probably by simply having a "preserve mass" checkbox that toggles between the two.

PR #1045, opened 2026-02-13
Unbalanced mass-action dynamics for Petri nets

### LINKED ISSUE #1043: Unbalanced mass-action for Petri nets should support individual rates
Rather than each transition having global consumption and production rates, there should be one consumption rate for each input place and one production rate for each output place.

tim-at-topos:
Closes #1043 and #1035

### Functionality

- [x] Per-place rates

### UI

- [x] Toggle for balanced vs. unbalanced
- [x] Toggle (*if* unbalanced is selected) for per-transition or per-place rates
- [x] Hide unnecessary UI (e.g. the above toggles, long names in production/consumption tables) for stock-flow diagrams
- [x] Make the toggles actually do stuff

### Refactoring

- [x] Merge mass-action and unbalanced mass-action and make the toggles actually do stuff
- [x] ~~Put stochastic mass-action back in~~ Put stochastic mass-action in its own file

### Other

- [x] Update documentation
- [x] Uncomment all failing tests
- [x] Search for `TODO`


tim-at-topos:
@epatters This is a much larger PR than I'd originally intended because I didn't just implement per-place rates for Petri nets, but also refactored (i.e. merged) the mass-action and unbalanced mass-action analyses. Let me know if anything needs more documentation/explanation!

There are some known issues with this (#1048 and #1051) but these are small enough (I think) that I should fix them in future PRs.

Edit: This now also closes #1051

packages/catlog-wasm/src/analyses.rs
```diff
@@ -46,50 +46,70 @@ pub(crate) fn latex_ob_names_mass_action(model: &DblModel) -> impl Fn(&Qualified
...
-                .mor_generator_label(id)
-                .map_or_else(|| id.to_string(), |label| label.to_string());
-            format!("r_{{\\text{{cons}},\\,\\text{{{name}}}}}")
+/// Creates a closure that formats morphism names for mass-action LaTeX output.
+pub(crate) fn latex_mor_names_mass_action(model: &DblModel) -> impl Fn(&Term) -> String {
+    |id: &Term| match id {
+        Term::UndirectedTerm { transition } => {
+            let transition_name = model
+                .mor_generator_label(transition)
+                .map_or_else(|| transition.to_string(), |label| label.to_string());
```
epatters: I doubt this is what we want because it will expose a raw UUID when the transition is unlabeled.

However, assuming that's out of scope for this PR (this problem predates the PR), you can at least save some typing and degrade more gracefully by replacing this with `model.mor_namespace.label_string(transition)` and similarly below.

packages/catlog/src/stdlib/analyses/ode/mass_action.rs
```diff
@@ -1,105 +1,191 @@
 //! Mass-action ODE analysis of models.
 //!
+//!
```
epatters: Extra newline has crept in here.

packages/catlog/src/stdlib/analyses/ode/mass_action.rs
```diff
@@ -1,105 +1,191 @@
...
 //! Such ODEs are based on the *law of mass action* familiar from chemistry and
-//! mathematical epidemiology.
+//! mathematical epidemiology. Here, however, we also consider generalised version
```
epatters: "consider generalised" -> "consider a generalised"

packages/catlog/src/stdlib/analyses/ode/mass_action.rs
```diff
@@ -1,105 +1,191 @@
...
+    PerPlace,
+}
 
-    /// Map from object IDs to initial values (nonnegative reals).
-    #[cfg_attr(feature = "serde", serde(rename = "initialValues"))]
-    pub initial_values: HashMap<QualifiedName, f32>,
+/// Terms in the generated polynomial equations are *undirected* in the balanced case
+/// and *directed* in the unbalanced case.
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub enum Term {
```
epatters: I find the terminology here a bit confusing. AFAIU, this outer enum `Term` describes a rate parameter, not a term, whereas the inner enum `RateParameter` describes a rate parameter in the unbalanced case.

Also confusing to use `Balanced` and `Unbalanced` above, but `Undirected` and `Directed` here.

No specific suggestion from me, but I have the feeling that the nomenclature could be improved.

packages/catlog/src/stdlib/analyses/ode/mass_action.rs
```diff
@@ -317,10 +427,42 @@ impl StockFlowMassActionAnalysis {
...
+                    data.transition_consumption_rates.get(transition).cloned().unwrap_or_default()
+                }
+                (Direction::IncomingFlow, RateParameter::PerPlace { transition, place }) => data
+                    .place_production_rates
+                    .get(transition)
+                    .cloned()
+                    .unwrap_or_default()
+                    .get(place)
+                    .copied()
+                    .unwrap_or_default(),
```
epatters, line 434: Hmm, you shouldn't need to clone the inner hash map. This should be something like:

```rust
data.place_production_rates.get(transition).and_then(|rates| rate.get(place)).copied().unwrap_or_default()
```

Ditto below.

packages/catlog/src/stdlib/analyses/stochastic/mass_action.rs
epatters, line 1: Still not the right place for this module because no ODEs are involved. I'd suggest having `stdlib/analyses/ode/mass_action.rs` (as it currently exists) and `stdlib/analyses/stochastic/mass_action.rs` (for this file).

packages/catlog/src/stdlib/analyses/stochastic/mass_action.rs
```diff
@@ -0,0 +1,158 @@
+//! Stochastic mass action anaylsis of ODEs.
```
epatters, line 1: This docstring is incorrect for the same reason.

packages/frontend/src/help/logics/petri-net.mdx
epatters, line 1: Thanks for updating these docs!

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -18,7 +24,9 @@ import type { MassActionSimulator } from "./simulator_types";
 
 import "./simulation.css";
 
-import styles from "./mass_action.module.css";
+import invariant from "tiny-invariant";
+
+import { MassActionConfigForm } from "./mass_action_config_form";
```
epatters: The formatter has sadly failed here: these imports should go with the others, above the CSS import.

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -47,43 +88,203 @@ export default function MassAction(
...
+        for (const mg of morGenerators()) {
+            const mor = model.morPresentation(mg);
+            if (!mor) {
+                continue;
+            }
+            transitionInterface[mg] = {
+                inputs: [],
+                outputs: [],
+            };
+            for (const [_, ob] of collectProduct(mor.dom).entries()) {
```
epatters: A small point, but you don't need to call `entries` if you don't need the indices, so this should be

```typescript
for (const ob of collectProduct(mor.dom)) {
```

and similarly below.

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -47,43 +88,203 @@ export default function MassAction(
...
+                // For now, our only modal theory of relevance is that of *symmetric*
+                // monoidal categories, so we don't need to iterate over [i, ob], nor
+                // do we need to worry about ${i}.
+                transitionInterface[mg].inputs.push(ob.content);
+            }
+            for (const [_, ob] of collectProduct(mor.cod).entries()) {
+                invariant(ob.tag === "Basic");
+                transitionInterface[mg].outputs.push(ob.content);
+            }
+        }
```
epatters, line 112: Come to think of it, it would more idiomatic to replace the above with something like:

```typescript
const inputs = collectProduct(mor.dom).map((ob) => {
    invariant(ob.tag === "Basic");
    return ob.content;
});
const outputs = collectProduct(mor.cod).map((ob) => {
    invariant(ob.tag === "Basic");
    return ob.content;
});
transitionInterface[mg] = { inputs, outputs };
```

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -39,6 +50,36 @@ export default function MassAction(
...
+    // For morphisms, the data that we need now does depend on massConservationType.
+    // We don't simply want to get a list of morphism generators, but instead
+    // account for the entire *interface* of each morphism. In a Petri net, this
+    // consists of a list of input places and a list of output places for each
+    // transition; in a stock-flow diagram, this consists of a singleton list
+    // of input stocks and a singleton list of output stocks for each flow.
+    type TransitionInterface = Record<
+        QualifiedName,
+        { inputs: QualifiedName[]; outputs: QualifiedName[] }
+    >;
```
epatters, line 77: Since this doesn't have to be JSON-able, it's IMO cleaner to replace the JS object with a `Map`:

```typescript
type TransitionInterface = Map<
        QualifiedName,
        { inputs: QualifiedName[]; outputs: QualifiedName[] }
>;
```

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -47,43 +88,203 @@ export default function MassAction(
...
-    }, []);
+    });
 
-    const obSchema: ColumnSchema<QualifiedName>[] = [
+    const morGeneratorsInterfaces = createMemo<TransitionInterface>(() => {
+        const model = elaboratedModel();
+        if (!model) {
+            return {};
+        }
+        const transitionInterface: TransitionInterface = {};
```
epatters: Then this becomes

```typescript
const transitionInterface: TransitionInterface = new Map();
```

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -47,43 +88,203 @@ export default function MassAction(
...
+    // The schema that we use for the <FixedTableEditor> JSX element depends on the
+    // value of MassConservationType. We might as well construct all possibilities.
+
+    // Firstly, the case MassConservationType = Balanced
+    const morSchema: ColumnSchema<QualifiedName>[] = [
         {
             contentType: "string",
             header: true,
-            content: (id) => elaboratedModel()?.obGeneratorLabel(id)?.join(".") ?? "",
+            content: (mor) => elaboratedModel()?.obGeneratorLabel(mor)?.join(".") ?? "",
```
epatters: Uh oh, looks like you're calling `obGeneratorLabel` on a morphism. (I think this might currently work, but only by accident.)

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -47,43 +88,203 @@ export default function MassAction(
...
     ];
 
-    const morSchema: ColumnSchema<QualifiedName>[] = [
+    // Secondly, the case MassConservationType = Unbalanced(PerTransition)
+    const morInputSchema: ColumnSchema<QualifiedName>[] = [
         {
             contentType: "string",
             header: true,
-            content: (id) => morLabelOrDefault(id, elaboratedModel()) ?? "",
+            content: (mor) => elaboratedModel()?.obGeneratorLabel(mor)?.join(".") ?? "",
```
epatters: Same here. Please review this code carefully.

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -47,43 +88,203 @@ export default function MassAction(
...
+    ];
+
+    // Finally, the case MassConservationType = Unbalanced(PerPlace)
+    const morInputsSchema: ColumnSchema<[QualifiedName, QualifiedName]>[] = [
+        {
+            contentType: "string",
+            header: true,
+            content: ([mor, input]) =>
+                (elaboratedModel()?.obGeneratorLabel(input)?.join(".") ?? "") +
+                " -> " +
```
epatters: Seems like we ought to able to do better than ASCII art arrows.

packages/frontend/src/stdlib/analyses/mass_action.tsx
```diff
@@ -47,43 +88,203 @@ export default function MassAction(
...
+                                schema={morInputsSchema}
+                            />,
+                            <FixedTableEditor
+                                rows={morGeneratorsOutputs()}
+                                schema={morOutputsSchema}
+                            />,
+                        ];
+                }
+        }
+    });
```
epatters: Putting a `JSX.Element` in a memo almost certainly isn't what you want. Use the `Match` and `Switch` components in Solid instead. Kaspar can help you with this if needed.

packages/frontend/src/stdlib/analyses.tsx
```diff
@@ -167,7 +167,12 @@ export function massAction(
         help,
         component: (props) => <MassAction title={name} {...otherOptions} {...props} />,
         initialContent: () => ({
-            rates: {},
```
epatters, line 180: **Important**: this is a *backwards-incompatible change* that will break existing analyses in the system and might also cause a runtime error for those analyses.

The easiest path forward is to make this change in a backwards-compatible way, which means preserving the name and meaning of the field `rates`.

epatters (changes requested):
Thanks Tim, this is a great set of features! Some comments, mostly minor, below.

packages/catlog-wasm/src/analyses.rs
tim-at-topos: `ob_namespace` and `mor_namespace` are both private fields of `DblModel` — should I make them public?

packages/frontend/src/stdlib/analyses/mass_action.tsx
tim-at-topos: ah yes, got it! @kasbah explained it 🙏

packages/catlog-wasm/src/analyses.rs
epatters: Ah yes, I think it's fine to make them public.

tim-at-topos:
@epatters comments addressed!

epatters (approved):
Thanks Tim. I made a few further changes.

packages/catlog/src/stdlib/analyses/ode/mass_action.rs
```diff
@@ -392,46 +521,39 @@ mod tests {
...
+            ),
+        );
+        // Note that the catalyst c is not left unchanged unless f is balanced
         let expected = expect!([r#"
-            dx = (-f) c x
-            dy = f c x
-            dc = 0
+            dx = (-(x->[f])) c x
+            dy = (([f]->y)) c x
+            dc = (([f]->c) + -(c->[f])) c x
```
epatters, line 534: This test of the balanced mass action for Petri nets should not have been effectively deleted by replacing it with a test for unbalanced mass action. I believe the corresponding functionality is now broken.

PR #1047, opened 2026-02-14
Equation types in DoubleTT

### LINKED ISSUE #782: fnotation support for parsing models with equations
From talking to @olynch, it seems like it would be a straightforward extension of the parser to support equations of models of double theories. For example, here is a model of the trivial double theory:

```
[
  A : Object, B : Object, C: Object,
  f : (Id Object)[A, B], g : (Id Object)[B, C], h : (Id Object)[A, C], 
]
```

We need support to declare that this is moreover a commutative triangle. This is useful for ergonomically creating models in tests.

KevinDCarlson:
Resolves #782 

We add type syntax and values for identity types, with the intended usage being for extensional equality types for morphism terms. 

The formal syntax is `Id(TyS,TmS,TmS)` where the two terms will evaluate as members of the evaluation of the type, and similar for the value. This type is displayed and entered in plain text as `a == b`.

In the text elaborator, we throw errors if the user attempts to construct the equality type between two terms that are not in a `Morphism` type, or if their types are not themselves convertible. 

We have a snapshot test to look at the model generated by a commutative square, as well as an test internal to Rust to confirm that this model successfully gets an equation, as a stopgap while pretty-printing for equations remains unimplemented. 

The largest diff is in modelgen, but part of this diff count is spurious: we move a lot of methods from `modelgen::Model` to `ModelGenerator`. This is for consistency with the primary really new function, `make_mor`, which makes a morphism in a `ModelGenerator`'s model, but needs access to the theory in order to make the object of an identity morphism's basepoint. It would be possible to move the moved stuff back and put `make_mor` on `Model`, at the cost of asking it to carry around an extra `ObType` argument for the `Id` case that would be irrelevant in the other cases. 

For making composite morphisms, we include an unchecked `concat_paths` function, which is only used to concatenate paths built from morphism terms which we already know to match up from `syn`. This function could be moved onto Path if that seems desirable.

I haven't included any way for the elaborator to use the equations from the equality type itself, since there should be no need: we never need to run `equal_tm` modulo propositional equality on morphism terms. I also haven't included a `refl` equality term, since again there's no obvious need for it; we just let equalities eta-convvert to `tt`.

I haven't tried to do anything with the notebook elaborator yet. I think this requires changes to the notebook data structure first, right?


KevinDCarlson:
I haven't included any way for the elaborator to use the equations from the equality type itself, since there should be no need: we never need to run `equal_tm` modulo propositional equality on morphism terms. I also haven't included a `refl` equality term, since again there's no obvious need for it; we just let equalities eta-convvert to `tt`.

KevinDCarlson:
@epatters You were wondering about what I was muttering about an eta-conversion to `tt`; this is used in `eta_neu(tm)` where `tm : Id(_,_,_)` is an *equation*, rather than before when `tm:Hom(_,_,_)` was a morphism. That is, any two terms of an identity type will check as equal up to eta, so the type theory behaves extensionally in this way. I think that in fact we will never even actually run this eta in the current state.

packages/catlog/src/tt/modelgen.rs
```diff
@@ -231,8 +297,33 @@ impl<'a> ModelGenerator<'a> {
...
             TyV_::Meta(_) => None,
         }
     }
 }
+
+/// Concatenates two paths without validation.
+///
+/// This is safe when the type checker has already verified that the paths are
+/// composable (the target of `p` equals the source of `q`).
+fn concat_paths<V, E>(p: Path<V, E>, q: Path<V, E>) -> Path<V, E> {
```
epatters: I think we can remove this function and replace any usages with `Path::pair(p, q).flatten()`.

Incidentally, in case it's ever relevant to know, there's also a checked version of `flatten` called `flatten_in`.

packages/catlog/src/tt/modelgen.rs
```diff
@@ -202,15 +156,127 @@ impl<'a> ModelGenerator<'a> {
...
+        match &**val {
+            TmV_::Neu(n, _) => {
+                let mut segments = Vec::new();
+                let mut n = n.clone();
+                while let TmN_::Proj(n1, f, _) = &*n.clone() {
+                    n = n1.clone();
+                    segments.push(*f);
+                }
+                segments.reverse();
+                let name: QualifiedName = segments.into();
```
epatters: I'd extract this bit of code from `let mut segments = ...` to `segments.into`, used here and in `make_ob` above, into a helper method `make_name(n) -> QualifiedName` or something along those lines.

I almost did that in a previous PR, but it now seems more clearly justified.

epatters (changes requested):
Thanks Kevin, this is really nice work. Just a few small comments below.

packages/catlog/src/tt/modelgen.rs
epatters: I went ahead and did this because I had to duplicate this logic yet again in #1069.

packages/catlog/src/tt/modelgen.rs
```diff
@@ -202,15 +180,96 @@ impl<'a> ModelGenerator<'a> {
...
+    fn make_mor(&self, val: &TmV, mor_type: &MorType) -> Option<Mor> {
+        match &**val {
+            TmV_::Neu(n, _) => Some(self.mor_generator(n.to_qualified_name())),
+            TmV_::Id(x) => {
+                let ob = self.make_ob(x, &self.theory.src_type(mor_type))?;
+                Some(self.model.id(ob))
+            }
+            TmV_::Compose(f, g) => {
+                // FIXME: This isn't correct because the morphism types of `f`
+                // and `g` need not be the morphism type of their composite.
```
epatters, line 255: @KevinDCarlson, I noticed this problem while doing a little cleanup.

To avoid a build up of PRs and subsequent merge conflicts, I'm going to merge this as is but I have created #1070 as a tracking issue.

PR #1052, opened 2026-02-17
RFC: Set-theoretic models of double theories

epatters:
Read it here: <https://branch-rfc-0001--catcolab.netlify.app/math/rfc-0001.xml>


tim-at-topos:
This is very clear and outlines a pretty clear path forwards. I read through it with Jason and we had a few questions, but I don't think that any of them would lead to changing this RFC.

One thing I do wonder is if you have an idea of how the next part of the story works: I have some model for my ODE system (e.g. $\dot{y}=k_\mathrm{interaction}xy$ and $\dot{y}=k_\mathrm{growth}y$), and I have some theory (e.g. a single object with two loose morphisms: $+$ and $-$, i.e. the theory for signed categories but viewed "non-unitally"); how do I specify what it means to interpret models of the theory in my ODE system? It seems like there could be some answer given by acset migration, but it doesn't seem like it'll be as simple as "precompose with a functor between theories".

epatters:
Thanks Tim (and Jason) for reading it through. Glad to hear it basically makes sense.

I have sketchier ideas about the model migration story, which I'll take the opportunity to write up in a "future possibilities" section.

epatters:
@tim-at-topos, I've pushed another commit with a provisional story about model migration. There's much more to say about that, but that's as far as I'd like to go in this RFC.

math-docs/trees/rfc-0001.tree
```diff
@@ -0,0 +1,555 @@
...
+possible approaches to model migration. There has yet to be any theoretical
+study of migrating models of double theories.}
+
+\p{Here I'll sketch an approach inspired by conversations with David Jaz Myers,
+who was in turn inspired by ideas of Bob Atkey. In outline, to migrate models of
+some theory \dbl{T} into ODEs, one first #{\Sigma}-migrates into a larger theory
+containing both \dbl{T} and the theory of ODEs, then #{\Delta}-migrates into the
+theory of ODEs. The intermediate theory additionally contains "builder"
+operations that will, under #{\Sigma}-migration, freely populate the ODE with
+variables and terms. The technique is illustrated by the following example.}
```
epatters, line 456: David Jaz points out that this looks like a collage:

> Since the builder patterns all go in one direction, I wonder if this could be expressed as the collage of a kind of tight bimodule between theories.

This idea should be given more consideration.

tim-at-topos:
This is now live at https://next.catcolab.org/math/rfc-0001.xml

PR #1057, opened 2026-02-17
Deferrable foreign keys for PostgresSQL backend

### LINKED ISSUE #1030: Generate SQL schemas with cycles
Follow up to #843, in which the generation of a SQL schema from a CatColab model is limited to the case where the generating graph has no cycles.

Fixing this is less trivial than one might think because SQL is bad. Probably you'd want to do an approximate topological sort (like directed graph visualization algorithms tend to do), then having special handling for any foreign keys that violating acyclicity in the approximate sorting.

### LINKED ISSUE #1034: Return structured data in error variant of `toposort`
The fallible `toposort` graph algorithm returns a String error type when it detects a cycle. The error message includes the UUID of the vertex of the graph, which is unhelpful to the end-user.

We need
- a custom error type containing the relevant vertex
- all frontend interfaces which use this error type (the SQL analysis) to report this error with human-readable names, not the UUID

quffaro:
Closes #1030 and closes #1034


packages/frontend/src/stdlib/styles.module.css
```diff
@@ -14,3 +14,17 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: flex-end;
+    gap: 4px;
+    flex-wrap: nowrap;
+    white-space: nowrap;
+    flex-direction: row;
+}
+
+.headerContainer > * {
```
quffaro: I don't think this is necessary

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -164,11 +166,80 @@ where
...
+    G::V: Hash + std::fmt::Debug + Clone,
+{
+    /// Stores an array of topologically-sorted vertices.
+    pub stack: Vec<G::V>,
+
+    /// Stores the feedback vertices with their outneighbors.
+    pub cycles: IndexMap<G::V, Vec<G::V>>,
+}
+
+impl<G> ToposortData<G>
```
epatters: You can derive this automatically method using `Constructor` from the `derive_more` crate.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -164,11 +166,80 @@ where
     result
 }
 
+/// Contains both the topologically-sorted stack of vertices and feedback vertices.
+#[derive(Debug, Clone)]
+pub struct ToposortData<G>
+where
+    G: FinGraph,
+    G::V: Hash + std::fmt::Debug + Clone,
```
epatters: It's considered bad practice to put trait bounds on the struct itself. Put them only on the `impl`s. See <https://stackoverflow.com/a/66369912>

Moreover, you should make the type parameter be `V` (vertices) rather than `G` (graph).

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -164,11 +166,80 @@ where
...
+    fn new(stack: Vec<G::V>, cycles: IndexMap<G::V, Vec<G::V>>) -> Self {
+        Self { stack, cycles }
+    }
+}
+
+/// Error type for Topological sort storing erroneous vertices.
+pub enum ToposortError<G>
+where
+    G: FinGraph,
+    G::V: Hash + std::fmt::Debug,
```
epatters: Ditto for this enum.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -164,11 +166,80 @@ where
...
+            Self::CycleError(v) => write!(f, "{:#?}", v),
+            Self::SelfLoop(v) => write!(f, "self loop at node {:#?}", v),
+        }
+    }
+}
+
+type ToposortResult<G> = Result<ToposortData<G>, ToposortError<G>>;
+
+/// Implementation of topological sort which throws an error when it encounters a cycle.
+pub fn toposort_strict<G>(graph: &G) -> ToposortResult<G>
```
epatters: I find this docstring confusing: in Rust, errors aren't thrown, exactly. Does this mean that an `Err` variant is returned when a cycle is encountered?

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -164,11 +166,80 @@ where
...
+where
+    G: FinGraph,
+    G::V: Hash + std::fmt::Debug,
+{
+    toposort_impl(graph, true)
+}
+
+/// Implementation of topological sort which does not throw an error when it encounters cycle.
+pub fn toposort_lenient<G>(graph: &G) -> ToposortResult<G>
```
epatters: Does this mean that it always returns the `Ok` variant?

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -204,32 +278,45 @@ where
     finish_stack.reverse();
 
     discovered.clear();
+    let mut cycles: IndexMap<G::V, Vec<G::V>> = Default::default();
```
epatters: I'm guessing that this line can be simplified to `let mut cycles = IndexMap::new()`.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -204,32 +278,45 @@ where
...
         let mut cycle = false;
-        while let Some(j) = rev_dfs_next() {
-            if cycle {
-                return Err(format!("Cycle detected involving node {:#?}", j).to_owned());
+        while let Some((j, outs)) = rev_dfs_next(cycle) {
+            if cycle && is_strict {
+                return Err(ToposortError::CycleError(j));
+            }
+            if cycle && !is_strict {
+                // TODO do we need entry?
```
epatters: Lingering TODO. Is this meaningful?

epatters (changes requested):
Thanks Matt! I left a first round of questions and comments on the toposort alg.

I see lots of remaining TODOs, so maybe this wasn't intended to be reviewed yet?

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -2,12 +2,19 @@ import download from "js-file-download";
...
 import Download from "lucide-solid/icons/download";
+import { Highlight, Language } from "solid-highlight";
 import { For, Match, Show, Switch } from "solid-js";
 
 import { BlockTitle, ErrorAlert, IconButton } from "catcolab-ui-components";
 import type { ModelAnalysisProps } from "../../analysis";
 import * as SQL from "./sql_types.ts";
 
+import styles from "../styles.module.css";
+// eslint-disable-next-line import/no-unassigned-import
```
kasbah: We don't use eslint (we use oxlint and one eslint-plugin for it) 🤔

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -2,12 +2,19 @@ import download from "js-file-download";
...
+import "prismjs/components/prism-sql";
+// eslint-disable-next-line import/no-unassigned-import
+import "prismjs/themes/prism.min.css";
+
```
kasbah: It would be nice to add have `CodeView` component so that the next person can just import and use that without worrying about global imports. 

It seems possible, if a little tricky, to support all languages but lazily loaded. Maybe PrismJS isn't the best to do this in a clean way. My first thought was something tree-sitter based but a quick search didn't turn up any obvious solution.

kasbah (changes requested):
Left some comments about syntax highlighting.

packages/frontend/src/stdlib/analyses/sql.tsx
quffaro: whoops! fixed

packages/frontend/src/stdlib/analyses/sql.tsx
quffaro: I'm willing to work on that in a separate PR.

kasbah: Still for this PR what is currently here could go in a CodeView component and it only supports SQL.

packages/frontend/src/stdlib/styles.module.css
```diff
@@ -14,3 +14,13 @@
...
+
+.headerContainer {
+    display: flex;
+    align-items: center;
+    justify-content: flex-end;
+    gap: 4px;
+    flex-wrap: nowrap;
+    white-space: nowrap;
+    flex-direction: row;
+}
```
kasbah: If this is only used used for sql it should go in a `analyses/sql.module.css`

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -301,14 +304,62 @@ where
...
+    toposort_impl(graph, true).map(|t| t.stack)
+}
+
+/// Implementation of topological sort which does not return an error when it encounters cycle.
+pub fn toposort_lenient<G>(graph: &G) -> ToposortData<G::V>
+where
+    G: FinGraph,
+    G::V: Hash + std::fmt::Debug,
+{
+    toposort_impl(graph, false).expect("This should never returns an error")
```
kasbah:
```suggestion
    toposort_impl(graph, false).expect("`toposort_impl` with `is_strict=false` gave no result")
```

kasbah (changes requested):

quffaro:
I've closed out some outdated comments here. This draft PR #1252 addresses @kasbah's recommendation

packages/catlog-wasm/src/theories.rs
```diff
@@ -94,19 +94,21 @@ impl ThSchema {
...
-                            .unwrap_or_else(|| QualifiedLabel::single("".into()))
-                    },
-                )
+                analyses::sql::SQLAnalysis::new(backend)
+                    .render(
+                        model.discrete()?,
+                        |id| {
+                            model
+                                .ob_generator_label(id)
+                                .unwrap_or_else(|| QualifiedLabel::single("".into()))
```
epatters: This common use scenario can now be simplified to

```rust
model.ob_namespace.label_string(id)
```

packages/catlog-wasm/src/theories.rs
```diff
@@ -94,19 +94,21 @@ impl ThSchema {
...
+                        },
+                        |id| {
+                            model
+                                .mor_generator_label(id)
+                                .unwrap_or_else(|| QualifiedLabel::single("".into()))
```
epatters: Similarly, this can be replaced with

```rust
model.mor_namespace.label_string(id)
```

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -301,14 +304,62 @@ where
...
+    toposort_impl(graph, true).map(|t| t.stack)
+}
+
+/// Implementation of topological sort which does not return an error when it encounters cycle.
+pub fn toposort_lenient<G>(graph: &G) -> ToposortData<G::V>
+where
+    G: FinGraph,
+    G::V: Hash + std::fmt::Debug,
+{
+    toposort_impl(graph, false).expect("`toposort_impl` with `is_strict=false` gave no result")
```
epatters: See the Rust style guidelines for how to write the error messages for `expect`: https://doc.rust-lang.org/std/result/enum.Result.html#method.expect

So this should be something like "toposort in lenient mode should return a valid result"

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -301,14 +304,62 @@ where
...
+}
+
+/// Error type for Topological sort storing erroneous vertices.
+#[derive(Debug)]
+pub enum ToposortError<V> {
+    /// There is a cycle in the graph.
+    CycleError(V),
+    /// There is a self-loop in the graph.
+    SelfLoop(V),
+}
```
epatters: This is a bit confusing since a self-loop is particular kind of cycle (a cycle of length zero). Is the intended behavior that the first variant will be returned for cycles of length > 0? If so, that should be clarified.

Alternatively, if it's not worth distinguishing carefully, you could just return `V` instead of `ToposortError<V>`.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -334,17 +385,23 @@ where
...
-            return Err(format!("Cycle detected involving node {:#?}", s));
-        }
+        if position[&s] >= position[&t] && is_strict {
+            return Err(ToposortError::CycleError(s));
+        };
+
+        if position[&s] >= position[&t] && !is_strict {
+            let outs = graph.out_neighbors(&s).collect();
+            cycles.insert(s, outs);
+        };
```
epatters: Style: IMO clearer to write

```rust
if position[&s] >= position[&t] {
  if is_strict {
    ...
  else {
    ...
  }
}
```

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -393,24 +450,32 @@ mod tests {
     #[test]
     fn toposorting() {
         let g = SkelGraph::path(5);
-        assert_eq!(toposort(&g), Ok(vec![0, 1, 2, 3, 4]));
+        let result = toposort_strict(&g);
+        match result {
+            Ok(vec) => assert_eq!(vec, vec![0, 1, 2, 3, 4]),
+            Err(e) => panic!("toposort failed: {}", e),
+        }
```
epatters: This test is more verbose and less idiomatic than before. Revert to previous style.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -393,24 +450,32 @@ mod tests {
...
 
         let mut g = SkelGraph::path(3);
         g.add_vertices(1);
         g.add_edge(2, 3);
         g.add_edge(3, 0);
-        expect_test::expect!["Cycle detected involving node 3"]
-            .assert_eq(&toposort(&g).unwrap_err());
+        // let t = &toposort_strict(&g).unwrap_err();
+        // expect_test::expect!["Cycle detected involving node 3"].assert_eq(&format!("{t}"));
```
epatters: Lingering commented out tests.

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -393,24 +450,32 @@ mod tests {
...
 
         let mut g = SkelGraph::path(4);
         g.add_vertices(2);
         g.add_edge(1, 4);
         g.add_edge(4, 3);
         g.add_edge(5, 2);
-        assert_eq!(toposort(&g), Ok(vec![5, 0, 1, 2, 4, 3]));
+        if let Ok(sort) = toposort_strict(&g) {
+            assert_eq!(sort, vec![5, 0, 1, 2, 4, 3]);
+        }
```
epatters: Ditto here: write

```rust
assert_eq!(toposort_strict(&g), Ok(vec![5, 0, 1, 2, 4, 3]));
```

or if you prefer

```rust
assert_eq!(toposort_strict(&g).unwrap(), vec![5, 0, 1, 2, 4, 3]);
```

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -420,9 +485,10 @@ mod tests {
         g.add_edge("1-4", 1, 4);
         g.add_edge("4-3", 4, 3);
         g.add_edge("5-2", 5, 2);
-        let sort = toposort(&g).unwrap();
-        let (i0, i1) = (sort.iter().position(|&x| x == 5), sort.iter().position(|&x| x == 2));
-        assert!(i0.unwrap() < i1.unwrap());
+        if let Ok(sort) = toposort_strict(&g) {
+            let (i0, i1) = (sort.iter().position(|&x| x == 5), sort.iter().position(|&x| x == 2));
+            assert!(i0.unwrap() < i1.unwrap());
+        }
```
epatters, line 466: Ditto.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -32,35 +38,188 @@ impl Iden for &QualifiedLabel {
     }
 }
 
+/// Enum for specifying the behavior of a column. For example, an Ordinary column is simply
+/// a foreign key constraint.
+#[derive(Debug, Clone, PartialEq)]
+pub enum ColumnBehavior {
```
epatters: Is this a behavior? Seems more like a type. Perhaps call `ColumnType` or `ColumnKind`.

packages/catlog/src/stdlib/analyses/sql.rs
```diff
@@ -32,35 +38,188 @@ impl Iden for &QualifiedLabel {
...
+            ColumnBehavior::Ordinary { mor, tgt } => {
+                fk(&ob_label(src), &mor_label(mor), &ob_label(tgt)) + ";"
+            }
+            ColumnBehavior::Deferrable { mor, tgt } => {
+                fk(&ob_label(src), &mor_label(mor), &ob_label(tgt))
+                    + "\n"
+                    + r#"DEFERRABLE INITIALLY DEFERRED;"#
+            }
+            // this is unreachable, since attributes cannot be foreign keys.
+            ColumnBehavior::Attribute { mor: _, tgt: _ } => "".to_string(),
```
epatters: Rust has a built-in macro for this: https://doc.rust-lang.org/std/macro.unreachable.html

epatters (changes requested):
Thanks Matt! This is looking pretty good. Should be the final round of review.

packages/catlog/src/one/graph_algorithms.rs
quffaro: I agree returning `V`, this error enum is frivolous.

epatters (approved):
Looks great!

PR #1059, opened 2026-02-17
Add a GitHub "deployment" for backend-next

kasbah:
After #990 backend deploys are no longer being associated with commits. This adds a GitHub "deployment" for backend-next and should also re-associate backend deploys with their commits.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -180,14 +180,40 @@ jobs:
           state: 'failure'
           deployment-id: ${{ needs.create_deployment.outputs.deployment_id }}
 
-  deploy_backend:
-    name: Deploy backend to AWS
+  create_backend_deployment:
+    name: Create backend GitHub deployment
     runs-on: ubuntu-latest
     if: >
       github.event.workflow_run.conclusion == 'success' &&
       github.event.workflow_run.event != 'pull_request' &&
       github.event.workflow_run.head_branch == 'main'
+    outputs:
+      deployment_id: ${{ steps.create_deployment.outputs.deployment_id }}
+    steps:
+      - name: Create new GitHub deployment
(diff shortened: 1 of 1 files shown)


jmoggr:
Do we know why they became dissociated? I've never had to worry about associations before.

I don't think I really understand this deployment thing, do you have docs on it? A quick google search doesn't turn up much. 

This feels needlessly complicated, I think we've managed to increase our LoC while complicating our workflow. Previously the backend deployment in the yaml was as simple as possible, what are we gaining from this additional complexity? is there any reason it can't work the old way?

We can do this, but I'd like to understand it better.

kasbah:
They are disassociated because the deploy is now triggered from `workflow_run` and it seems GH doesn't automatically set the status on the commit that triggered the previous workflow. 

The docs for "deployments" are here https://docs.github.com/en/rest/deployments/deployments . We are only using the status side of this obviously. No 3rd parties involved (except Netlify in a way but we don't use any of their integration for this I believe). 

To be clear: we could also set the status on the commit via an API request. But having a "deployment" seems like a nicer way to set the status on the GH web UI over all. 

The main reason for splitting build and deploy is to be clearer about whether a workflow/job:

1. Produces artifacts
2. Deploys artifacts and needs access to secrets (and which environments i.e. preview /  staging / production)

Knowing 2. paves the way to allow us to start safely building (without approval) and deploying (with approval) external contributors' code.

jmoggr (approved):
Okay. I think I understand this. I have not tested it.

I do not love the additional complexity and do not think it is worth it if the amount of pain we have previously experienced is anything to go by. But I acknowledge that this does progress towards the goal of automatically running workflows on external contributors code, and that I do not see another way of easily achieving that outcome from our current position.

PR #1060, opened 2026-02-18
Equations between morphisms in notebook types

KevinDCarlson:
Ahead of notebook elaboration of equations, this PR introduces equations into the `notebook_types` modules `model, model_judgment, diagram_judgment`, thus necessitating an upgrade of notebook types to `v2`. Other than the modules already listed, the main change of interest is in `document`, which includes code for migrating `v1` to `v2` documents. The only actual modification this migration requires is changing the version number, but there's some slightly wordy unpacking and repacking seemingly necessary since we have to reconstruct `ModelJudgment` and `DiagramJudgment` types for the new version. 

It's probably possible to argue for not bothering with upgrading `DiagramJudgment` at this time, since we have no type theoretic treatment of diagrams yet, but I figured it's a minor enough change which we may want before long and versioning `notebook_types` is a bit fiddly. Some more specific comments are attached in the diff.


packages/catlog/src/tt/notebook_elab.rs
```diff
@@ -233,7 +233,7 @@ impl<'a> Elaborator<'a> {
             Some(l) => l,
             None => return self.ty_error(InvalidDblModel::InvalidLink(name)),
         };
-        let notebook_types::v1::LinkType::Instantiation = link.r#type else {
+        let notebook_types::current::LinkType::Instantiation = link.r#type else {
```
KevinDCarlson: I was a bit confused about this.

packages/notebook-types/src/bin/migrate_examples.rs
```diff
@@ -3,8 +3,8 @@ use std::fs;
 use std::path::Path;
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let input_dir = Path::new("examples/v0");
-    let output_dir = Path::new("examples/v1");
+    let input_dir = Path::new("examples/v1");
```
KevinDCarlson: These are hard-coded because it shouldn't need to be re-run after each time a new version is merged.

packages/notebook-types/src/v1/mod.rs
```diff
@@ -5,15 +5,11 @@ pub use v0::{analysis, api, cell, diagram_judgment, model, model_judgment, path,
 pub mod document;
 pub mod notebook;
 
-pub use analysis::*;
```
KevinDCarlson, line 8: IIUC, these were all public for when v1 was `current`.

packages/notebook-types/src/v2/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+pub use theory::*;
+
+#[cfg(test)]
+mod test {
+    use super::document::Document;
+    use crate::test_utils::test_example_documents;
+
+    #[test]
+    fn test_v2_examples() {
+        test_example_documents::<Document, _>("examples/v2", |_, _| {});
```
KevinDCarlson: Just tests that deserialization runs without panicking.

packages/notebook-types/src/v2/model.rs
```diff
@@ -0,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+use tsify::Tsify;
+
+pub use crate::v1::model::*;
+
+/// An equation between two morphisms in a model of a double theory.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi, missing_as_null)]
+pub struct Eqn(pub Mor, pub Mor);
```
KevinDCarlson: I couldn't think of any necessary variants of equations.

KevinDCarlson:
I found it was necessary to remove TSIFY annotations from some old types to make the versioning work here. This didn't cause problems on the v0 -> v1 upgrade because the redefined types were structurally identical, but we actually had duplicated type definitions in the WASM output.

epatters (changes requested):
Knowing when to bump the notebook types version can be confusing. The main principle is this: we only bump the version and write a migration for changes that would break existing user data. Changes or additions to the notebook types for which existing data would remain valid do not trigger a version bump.

The most common example of such a safe change is adding new variants to an enum. And fortunately that's exactly what we have here.

So this PR can be much simplified by just adding the `Equation` variant to the existing v1 type `ModelJudgment`. We did the same thing when adding instantiations:

- #761

Issue #1017, opened 2026-02-06
Dedicated types for error reporting in notebook elaborator


epatters:
The text elaborator for DoubleTT has great error reporting; the notebook elaborator not so much. Partly we're missing integration work; for example, the frontend is not displaying textual error messages for error variants received from the core, an issue that predates the DoubleTT imlpementation. But I think the fundamental issue is more conceptual: distinguishing between elaboration and validation.

Currently, the notebook elaborator is using the enum `catlog::dbl::model::InvalidDblModal` as its error type. This enum was originally intended for validating models constructed by any means, especially programmatically. Reusing it for the notebook elaborator has already caused variants to leak into it that don't make sense for validation (`UnsupportedFeature`, `InvalidLink`). In other places, the notebook elaborator simply doesn't return any useful error info at all (`ob_syn`, `ob_chk` return only `Option`s).

We should separate the error types for elaboration and validation. While this may cause a little redundancy, the key advantage is that errors detected earlier can be reported better, with clear sign-posting about where they occurred. The text elaborator in its current form is a good example of this principle.

Follow-up to #1016 that, if implemented properly, should also solve #967.

PR #1061, opened 2026-02-19
Notebook elaboration for equality types

KevinDCarlson:
Depends on #1047 and #1060. Elaboration of notebooks containing equations to the type theory. Note that the diffs to modules in `tt` other than `notebook_eval` itself are spurious, due to the double parentage.


packages/catlog/src/dbl/discrete/model.rs
```diff
@@ -59,7 +59,7 @@ impl DiscreteDblModel {
         let category_errors = self.category.iter_invalid().map(|err| match err {
             InvalidFpCategory::Dom(e) => Invalid::Dom(e),
             InvalidFpCategory::Cod(e) => Invalid::Cod(e),
-            InvalidFpCategory::Eq(eq, errs) => Invalid::Eq(eq, errs),
+            InvalidFpCategory::Eqn(eq, errs) => Invalid::Eqn(Some(eq), errs),
```
KevinDCarlson: I changed these two kinds of "Eq" to "Eqn" because "Eq" is used so often in Rust.

packages/catlog/src/dbl/model.rs
```diff
@@ -290,7 +290,7 @@ pub enum InvalidDblModel {
     CodType(QualifiedName),
 
     /// Equation between morphisms has one or more errors.
-    Eq(usize, NonEmpty<InvalidPathEq>),
+    Eqn(Option<usize>, NonEmpty<InvalidPathEq>),
```
KevinDCarlson: I'm not so sure about this one--in the notebook elaborator we don't have a sequence of equations we're iterating over to provide this usize, so I leave it blank. But that might be a smell that we should be providing errors some different way.

packages/catlog/src/one/path.rs
```diff
@@ -620,6 +620,9 @@ pub enum InvalidPathEq {
     /// Right-hand side of equation is not a valid path in the graph.
     Rhs,
 
+    /// Left- and right-hand sides live in different graphs.
+    Graph,
```
KevinDCarlson: Also not entirely clear whether this variant makes sense, but I'm happier about it than those in `dbl::model`.

epatters:
Now that #1047 and #1060 have been merged, you'll need to rebase this one and fix the merge conflicts.

KevinDCarlson:
> Now that #1047 and #1060 have been merged, you'll need to rebase this one and fix the merge conflicts.

Done.

packages/catlog/examples/tt/notebook/commutative_square.json
```diff
@@ -0,0 +1,138 @@
...
+      "0b4acecb-32ba-4617-a6c0-872d7556125a",
+      "ec087bad-4f64-4739-ae37-91c08fd0d2a2",
+      "6cda57a8-a7ec-4094-af8c-3fc087aeb3ef",
+      "3eb0ede7-3d32-493a-979b-555d63eb4146",
+      "71283813-ba66-45d0-9375-132f1c1878e7",
+      "40bcaa10-5fc0-485d-83a7-8b061925ff31",
+      "2cfffa10-7189-40e5-96e0-ab598e79c43d"
+    ]
+  },
+  "version": "2"
```
epatters: This should be version 1 after the changes to #1060.

packages/catlog/src/tt/notebook_elab.rs
```diff
@@ -227,6 +277,68 @@ impl<'a> Elaborator<'a> {
...
+            }
+        }
+        match (NonEmpty::from_vec(errors), lhs, rhs) {
+            (None, Some((lhs_s, lhs_v, lhs_ty)), Some((rhs_s, rhs_v, _))) => {
+                let ty_s = TyS::id(self.evaluator().quote_ty(&lhs_ty), lhs_s, rhs_s);
+                let ty_v = TyV::id(lhs_ty, lhs_v, rhs_v);
+                (ty_s, ty_v)
+            }
+            (Some(errors), _, _) => {
+                self.ty_error(InvalidDblModel::Eqn(None, errors)) //The assumption in InvalidDblModel that we should already have the vector of equations built up, so as to give the index in the first argument here, doesn't hold in this case. It would be best in principle not to use InvalidDblModel here before we've begun to build a DblModel...
```
epatters: I'm very surprised that the formatter didn't balk at this, but we shouldn't have an inline comment going on this long. Put the comment above this line and hard wrap it (which the formatter *should* do automatically).

As for the substance of the comment, you're absolutely right: the `InvalidDblModel` error type was not intended for this feature, a problem going back to the original implementation of the notebook elaborator. We should have a dedicated error type; see #1017.

So I won't block on the PR on this, but perhaps you can take on #1017 once you've wrapped up this PR.

packages/catlog/src/tt/notebook_elab.rs
```diff
@@ -227,6 +277,68 @@ impl<'a> Elaborator<'a> {
...
+            }
+            _ => unreachable!(),
+        }
+    }
+
+    fn equation_cell(&mut self, eqn_decl: &nb::EqnDecl) -> (NameSegment, LabelSegment, TyS, TyV) {
+        let name = NameSegment::Uuid(eqn_decl.id); //Kind of funny that the decl's id produces the cell's name but the decl's name produces the cell's label.
```
epatters: Ditto about the formatting of this comment.

But yes, this is an annoying historical artifact that will require breaking changes + migration to the notebook types to fix. We used to say "name" for what we now call "label," and "name" and "id" are now synonymous, esp. because textual names in the text elaborator correspond to the UUID names in the notebook elaborator. So I'd suggest just removing this comment: we know what we need to fix; we just need to decide when we're ready to bite the bullet and make the nomenclature consistent.

We should probably track this in an issue...

packages/catlog/src/one/path.rs
```diff
@@ -620,6 +620,9 @@ pub enum InvalidPathEq {
...
+
```
epatters: It's a bit of an abuse to add this variant to this type, which about how a path in a (single) graph can be invalid. Although it's more annoying, the correct thing to do is make a new error type, for I guess paths in models of double theories.

packages/catlog/src/dbl/model.rs
```diff
@@ -299,7 +299,7 @@ pub enum InvalidDblModel {
...
```
epatters: Why is this an `Option` now?

*Edit*: Ok, from the comment below, I now remember what's going. Can you please put a `FIXME` in the comment above as a reminder to undo this once we've fixed #1017.

PR #1064, opened 2026-02-20
Fix mass-action equations analyses for Petri nets

### LINKED ISSUE #1063: Mass-action equations analysis broken for Petri nets
When trying to recreate https://github.com/ToposInstitute/CatColab/issues/986 I found that both the unbalanced and "balanced" mass-action dynamics equations analyses result in an error:

```
Error: Model should be of a discrete tabulator theory
```

This might have been introduced in #1045 so I'm going to look into it.

tim-at-topos:
closes #1063


epatters:
@tim-at-topos, is there anything left to do here?

tim-at-topos:
unless you can see a way to reduce the duplication between `mass_action_equations_modal()` and `mass_action_equations_tab()`, this is all done!

Issue #1065, opened 2026-02-20
Incompatible changes to `MassActionProblemData`


tim-at-topos:
In #1045 the struct/interface `MassActionProblemData` was changed by adding the following fields:

- `massConservationType`
- `transitionProductionRates`
- `transitionConsumptionRates`
- `placeProductionRates`
- `placeConsumptionRates`

This means that previous analysis documents with a mass-action dynamics analysis will no longer load correctly, since their `MassActionProblemData` will be missing these fields.

tim-at-topos:
Sorry for introducing such a bad breaking bug... I have no idea how I didn't see this in testing.

epatters:
That's alright! This is why we have a staging server.

Also, in retrospect, I should have noticed that problem when reviewing.

PR #1066, opened 2026-02-20
Instantiate default values for MassActionProblemData

tim-at-topos:
This is one attempt to close https://github.com/ToposInstitute/CatColab/issues/1065

### DIFF
--- a/packages/frontend/src/stdlib/analyses/mass_action.tsx
+++ b/packages/frontend/src/stdlib/analyses/mass_action.tsx
@@ -37,6 +37,45 @@ export default function MassAction(
 ) {
     const elaboratedModel = () => props.liveModel.elaboratedModel();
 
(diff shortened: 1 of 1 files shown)


tim-at-topos:
@epatters @kasbah I could think of a few ways to fix #1065 but didn't know which would be best. This seems like the messiest but simplest way. Two alternatives that came to mind:

1. Implement `Default` in the Rust struct and use this to fill in any missing fields. (I tried this but had serde issues and couldn't get it to work).
2. Rewrite `mass_action.tsx` to include a bunch of `?` whenever accessing any of these newly introduced fields. (I didn't try this but it does sound more sensible to me than the solution in this current PR).

epatters:
On a second thought, I propose what should be a more convenient solution in #1068.

PR #1068, opened 2026-02-20
Use initial content of analysis also as the default content

### LINKED ISSUE #1065: Incompatible changes to `MassActionProblemData`

epatters:
This is relevant only when new fields are added to the analysis content, in which case it will help with backwards compatibility.

@tim-at-topos, this should fix #1065 while also making it easier to make such changes in the future. Can you confirm that this works for you?

This PR assumes that `initialContent` is cheap to call, which seems safe enough, certainly for all our existing analyses.

### DIFF
--- a/packages/frontend/src/analysis/analysis_editor.tsx
+++ b/packages/frontend/src/analysis/analysis_editor.tsx
@@ -45,7 +45,7 @@ export function AnalysisNotebookEditor(props: { liveAnalysis: LiveAnalysisDoc })
 }
 
 /** Editor for a notebook cell in an analysis notebook. */
-function AnalysisCellEditor(props: FormalCellEditorProps<Analysis<unknown>>) {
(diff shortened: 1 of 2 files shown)


tim-at-topos:
yup, this works for me and is considerably neater than #1066 (in particular, this doesn't actually modify the content of extant analyses)

epatters:
On further thought, this seems incorrect because it doesn't actually add the new fields to the persisted data, which is likely to break code that attempts to *mutate* those fields assuming they exist. Trying again with #1071.

PR #982, opened 2026-02-02
Use integers for initial values in stochastic mass-action problem data

quffaro:

PR #1071, opened 2026-02-23
Allow analysis content to be extended with new fields

### LINKED ISSUE #1065: Incompatible changes to `MassActionProblemData`

epatters:
Second attempt to resolve #1065 in a general way.

### DIFF
--- a/packages/frontend/src/analysis/analysis_editor.tsx
+++ b/packages/frontend/src/analysis/analysis_editor.tsx
@@ -1,12 +1,4 @@
-import {
-    createEffect,
-    createMemo,
-    createSignal,
-    Match,
-    Switch,
-    untrack,
-    useContext,
-} from "solid-js";
+import { Match, Switch, useContext } from "solid-js";
 import { Dynamic } from "solid-js/web";
 import invariant from "tiny-invariant";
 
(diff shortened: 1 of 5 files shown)


packages/frontend/src/stdlib/analyses/stochastic_mass_action.tsx
```diff
@@ -58,7 +58,7 @@ export default function StochasticMassAction(
             content: (id) => elaboratedModel()?.obGeneratorLabel(id)?.join(".") ?? "",
         },
         createNumericalColumn({
-            name: "Initial values (positive integers)",
+            name: "Initial value",
```
kasbah, line 61: Why are you changing this back? There should be some kind of indication to the user what they can enter here. The field just goes red and undoes the change when it's not valid. https://github.com/ToposInstitute/CatColab/pull/982#discussion_r2759094412

kasbah (changes requested):
As mentioned in meeting it would be good to have a way to easily run the migration for tests. Here is what I did for the db-dump test to be able to simulate: https://github.com/ToposInstitute/CatColab/pull/1075/changes#diff-198762286b0988782327bb267e59a5dcc74560b361c940cbe7b49a6736e381acL159

packages/frontend/src/stdlib/analyses/stochastic_mass_action.tsx
epatters: I changed it back because the long column header causes the table to take up lots of unnecessary space and, more importantly, because it's inconsistent with how we display other parameter columns. This was particularly glaring in my test analysis notebook which happened to put a deterministic and a stochastic mass-action analysis side-by-side.

I agree that we should communicate constraints like this (including nonnegativity constraints for floats), but I'd like it to be done in a visually nicer way (perhaps with a tooltip?) and consistently across similar analyses.

epatters:
I've pushed another commits that gets the migration logic out of a component into a more appropriate place.

PR #1075, opened 2026-02-23
Test backwards compatibility of analyses with DB data

kasbah:
This adds a test for backwards compatibility of analyses. We will run this test on production data in a private Github repo. I added an SSH key for that private Github repo to access production.

### DIFF
--- a/infrastructure/scripts/dump-notebook-fixtures
+++ b/infrastructure/scripts/dump-notebook-fixtures
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+# Dumps all analysis documents and their referenced models and diagrams from a database as a JSON object.
+#
+# Usage: dump-notebook-fixtures <ssh-target> <output-file>
+# Example: dump-notebook-fixtures catcolab@backend-next.catcolab.org fixtures.json
+#
+# Output format:
+# {
+#   "analyses": [...],  // Array of analysis documents
+#   "models": [...],    // Array of model documents referenced by analyses (directly or via diagrams)
+#   "diagrams": [...]   // Array of diagram documents referenced by diagram analyses
+# }
+
+set -euo pipefail
+
+ssh_target="${1:?Usage: dump-notebook-fixtures <ssh-target> <output-file>}"
+output_file="${2:?Usage: dump-notebook-fixtures <ssh-target> <output-file>}"
+
+echo "Dumping analysis documents, models, and diagrams from $ssh_target to $output_file..." >&2
+
+# Run the query remotely via SSH. The remote script:
+# 1. Extracts DATABASE_URL from the agenix secrets file
+# 2. Queries PostgreSQL to get all non-deleted analysis documents
+# 3. Queries PostgreSQL to get all model documents referenced by model analyses
+# 4. Queries PostgreSQL to get all diagram documents referenced by diagram analyses
+# 5. Queries PostgreSQL to get all model documents referenced by those diagrams
+# 6. Returns a JSON object with all three arrays
+ssh "$ssh_target" bash -s > "$output_file" <<'REMOTE_SCRIPT'
+set -euo pipefail
+
+# Read DATABASE_URL from the agenix-decrypted secrets file
+if [[ ! -f /run/agenix/catcolabSecrets ]]; then
+  echo "Error: /run/agenix/catcolabSecrets not found on remote host" >&2
+  exit 1
+fi
+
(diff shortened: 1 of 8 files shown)


jmoggr (approved):

jmoggr:
Code looks fine and does what it sets out to do. Some notes/comments:

- It might be good to document somewhere in this code how this is meant to be run. It's not obvious looking at the code that this will be run in a separate repo. Someone could look at this and reasonably come to the conclusion that it is dead code.

- `infrastructure/scripts/dump-notebook-fixtures` uses bash and the JSON document snapshots in the DB. We'd like to be moving away from both. It feels like something that ought to be re-written in rust, since that would get rid of the bash and allow us to operate on the automerge documents instead of the snapshots.

- I don't love how this complicates our build and testing pipeline, since we are now splitting stuff across multiple repositories. The only idea I've had:
  - don't print the test logs and instead create private artifacts. This adds the additional step of needing to download the artifacts to view the logs when tests fail.
    - It's not obvious to me if this is better or worse than a separate repo. I lean towards better, but am sympathetic to the other perspective.

infrastructure/scripts/dump-notebook-fixtures
epatters, line 1: Jason wrote:

> infrastructure/scripts/dump-notebook-fixtures uses bash and the JSON document snapshots in the DB. We'd like to be moving away from both. It feels like something that ought to be re-written in rust, since that would get rid of the bash and allow us to operate on the automerge documents instead of the snapshots.

I won't block the PR on it, but I agree with Jason's comment. Another advantage to writing it in Rust is that we'd check that the SQL queries are valid at compile time.

packages/frontend/src/theory/theory.ts
```diff
@@ -292,11 +292,27 @@ export type AnalysisMeta<T> = {
...
     /** Component that renders the analysis. */
     component: ModelAnalysisComponent<T>;
+
+    /** Optional run function for testing backward compatibility.
+
+    When present, this function takes a compiled model and analysis content and
+    runs the analysis. It exercises the same WASM deserialization path as the
+    component would at runtime.
+     */
+    run?: (model: DblModel, data: T) => unknown;
```
epatters, line 393: If we're going down this route, then rather than adding a duplicative, optional field that people have to know about and remember to set, we should refactor the analyses to have a uniform interface based on `run` or similar. After all, every analysis we have does run *something* and we already have a semi-standardized set of verbs like `simulate` and `check`.

This might also help with integrating web workers in a uniform way (#1105) but I'm less sure about that.

packages/frontend/src/stdlib/notebook_backwards_compat.db-dump-test.ts
```diff
@@ -0,0 +1,316 @@
...
+            const theoryId = migratedModel.theory as string;
+            const theory = await stdTheories.get(theoryId);
+
+            const instantiated = new DblModelMap();
+            const compiledModel = elaborateModel(
+                migratedModel.notebook as ModelNotebook,
+                instantiated,
+                theory.theory,
+                modelRefId,
+            );
```
epatters, line 183: I don't like that we've had to duplicate all this logic for migrating and elaborating the model here. That logic has often changed in the past and it's likely to continue to change in the future.

packages/frontend/src/stdlib/notebook_backwards_compat.db-dump-test.ts
```diff
@@ -0,0 +1,316 @@
...
+                return;
+            }
+
+            const diagramJudgments: DiagramJudgment[] = NotebookUtils.getFormalContent(
+                migratedDiagram.notebook as DiagramNotebook,
+            );
+
+            const compiledDiagram = elaborateDiagram(diagramJudgments, theory.theory);
+
+            compiledDiagram.inferMissingFrom(compiledModel);
```
epatters, line 284: Ditto here about duplicating logic for elaborating diagrams.

epatters (changes requested):
Thanks Kaspar. It will be important for us to improve our testing infrastructure as the system grows in complexity and I'm glad you've started to work on that.

Taken together, though, I think what my comments below show is that our system is not currently architected to make this testing easy or convenient. I suspect the problem is that too much of the pipelines `model -> analysis` and `model -> diagram -> analysis` are stuck in components or are otherwise not reusable in a standalone script. We should fix that before proceeding because as is I fear that this code will itself become a maintenance problem. I'd be happy to help think through what refactoring is needed if you like.

PR #1081, opened 2026-02-26
Allow ObTypes and MorTypes to check and synthesize correctly

### LINKED ISSUE #1070: Use bidirectional inference in DoubleTT model generation
Starting with #1047, in the DoubleTT model generation, we need to

- *check* object types when generating the objects to be the domain or codomain of a morphism generator
- *synthesize* morphism types when constructing the left/right hand sides of an equation

In my understanding, bidirectional inference, already used in other parts of DoubleTT, is the right approach for this situation.

KevinDCarlson:
Resolves #1070 .

### DIFF
--- a/packages/catlog/src/tt/modelgen.rs
+++ b/packages/catlog/src/tt/modelgen.rs
@@ -188,8 +188,8 @@ impl<'a> ModelGenerator<'a> {
         }
     }
 
-    /// Constructs an application of an object operation, if the theory is modal. 
-    /// 
+    /// Constructs an application of an object operation, if the theory is modal.
+    ///
     /// Returns the constructed object along with the expected object type of the result.
     fn ob_app(&self, name: &NameSegment, tm_v: &TmV) -> Option<(Ob, ObType)> {
         let name: QualifiedName = [*name].into();
@@ -224,7 +224,7 @@ impl<'a> ModelGenerator<'a> {
     }
 
     /// Attempts to make an object of a model from a term.
-    /// 
+    ///
     /// Returns the object together with the appropriate type.
     fn make_ob_synth_type(&self, val: &TmV) -> Option<(Ob, ObType)> {
         match &**val {
@@ -245,7 +245,7 @@ impl<'a> ModelGenerator<'a> {
     }
 
     /// Attempts to make an object of a model from a term.
-    /// 
+    ///
     /// Also checks that the term constructs an object of the given type, returning only the object if successful.
     fn make_ob_check_type(&self, val: &TmV, ob_type: &ObType) -> Option<Ob> {
         match &**val {
@@ -264,7 +264,7 @@ impl<'a> ModelGenerator<'a> {
     }
 
     /// Attempts to make a morphism of a model from a term.
-    /// 
+    ///
     /// Also returns the expected morphism type of the result.
     fn synth_mor(&self, val: &TmV) -> Option<(Mor, MorType)> {
         match &**val {
@@ -290,7 +290,7 @@ impl<'a> ModelGenerator<'a> {
     }
 
     /// Attempts to make a morphism from a term of the given morphism type.
-    /// 
+    ///
     /// At this time, all morphism constructors allow for type synthesis, but
     /// eventually this will change.
     fn make_mor(&self, val: &TmV, mor_type: &MorType) -> Option<Mor> {


packages/catlog/src/tt/modelgen.rs
```diff
@@ -180,14 +180,6 @@ impl<'a> ModelGenerator<'a> {
         self.extract(vec![], &tm_v, ty).unwrap_or_else(Namespace::new_for_uuid)
     }
 
-    /// Constructs a generating object.
-    fn ob_generator(&self, name: QualifiedName) -> Ob {
```
KevinDCarlson, line 184: Incorporated into synth_ob.

packages/catlog/src/tt/modelgen.rs
```diff
@@ -196,16 +188,19 @@ impl<'a> ModelGenerator<'a> {
         }
     }
 
-    /// Constructs an application of an object operation, if possible.
-    fn ob_app(&self, name: QualifiedName, tm_v: &TmV) -> Option<Ob> {
+    /// Constructs an application of an object operation, if the theory is modal, along with
+    /// the expected object type of the result.
+    fn ob_app(&self, name: &NameSegment, tm_v: &TmV) -> Option<(Ob, ObType)> {
```
KevinDCarlson, line 194: Could be incorporated into synth_ob.

packages/catlog/src/tt/modelgen.rs
```diff
@@ -227,40 +222,76 @@ impl<'a> ModelGenerator<'a> {
...
+
+    /// Attempts to make an object of a model from a term, checking that the term constructs an object
+    /// of the given type.
     fn make_ob(&self, val: &TmV, ob_type: &ObType) -> Option<Ob> {
         match &**val {
-            TmV_::Neu(n, _) => Some(self.ob_generator(n.to_qualified_name())),
-            TmV_::App(name, tm_v) => self.ob_app([*name].into(), tm_v),
+            // ob_type checked recursively.
             TmV_::List(elems) => {
-                let el_type = ob_type.clone().list_arg().unwrap();
```
KevinDCarlson, line 236: Seemed like we didn't actually want to panic on None here?

packages/catlog/src/tt/modelgen.rs
```diff
@@ -196,16 +188,19 @@ impl<'a> ModelGenerator<'a> {
         }
     }
 
-    /// Constructs an application of an object operation, if possible.
-    fn ob_app(&self, name: QualifiedName, tm_v: &TmV) -> Option<Ob> {
+    /// Constructs an application of an object operation, if the theory is modal, along with
+    /// the expected object type of the result.
```
epatters: Thank you for taking the time to write/improve the doc comments!

Some advice on writing doc comments from the [Rust book](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html):

> Everything before the first empty line will be reused to describe the component in searches and module overviews... It is good practice to keep the summary to one line: concise writing is a goal of good documentation.

E.g., I'd write this doc comment as something like:

```rust
/// Constructs an application of an object operation, if the theory is modal.
///
/// Also returns the expected object type of the result.
```

Ditto for a few other doc comments below.

packages/catlog/src/tt/modelgen.rs
```diff
@@ -227,40 +222,76 @@ impl<'a> ModelGenerator<'a> {
...
+                Some((ob, ob_type.clone()))
+            }
+            TmV_::App(name, tm_v) => self.ob_app(name, tm_v),
+            _ => None,
+        }
+    }
+
+    /// Attempts to make an object of a model from a term, checking that the term constructs an object
+    /// of the given type.
     fn make_ob(&self, val: &TmV, ob_type: &ObType) -> Option<Ob> {
```
epatters: I wonder if we could tweak the names of the methods to make the synth/check distinction more obvious, perhaps:

- `synth_ob` -> `make_ob_synth_type`
- `make_ob` -> `make_ob_check_type`

Open to the other suggestions.

packages/catlog/src/tt/modelgen.rs
```diff
@@ -227,40 +222,76 @@ impl<'a> ModelGenerator<'a> {
...
-                Some(self.model.compose2(mf, mg))
+                let (mf, mtf) = self.synth_mor(f)?;
+                let (mg, mtg) = self.synth_mor(g)?;
+                Some((self.model.compose2(mf, mg), self.theory.compose_types2(mtf, mtg)?))
             }
             _ => None,
         }
     }
 
+    /// Attempts to make a morphism from a term of the given morphism type.
```
epatters, line 292: E.g., here you'd want to put a blank line after the first line in the doc comment.

epatters (dismissed):
Thanks Kevin, this looks great. Just a few minor comments.

PR #1085, opened 2026-02-28
Pretty-print equations in models of discrete theories

KevinDCarlson:
This pretty-prints equations in discrete double models for the first time. An example is being tested in the snapshot test `test_equality.dbltt`, which exhibits output looking like so: 

```
generate CommutativeSquare
#/ result: model generated by 4 objects and 4 morphisms
#/ NW : Entity
#/ NE : Entity
#/ SW : Entity
#/ SE : Entity
#/ t : NW -> NE : Hom Entity
#/ l : NW -> SW : Hom Entity
#/ r : NE -> SE : Hom Entity
#/ b : SW -> SE : Hom Entity
#/ t ⋅ r = l ⋅ b : Hom Entity(NW,SE)
```

Arguably the suffix could be any of 

- `: NW -> SE` 
- `(Hom Entity)(NW,SE)`
- `Hom Entity NW SE`, or if we modify the morphism type printing, `
- `Hom(Entity)(NW,SE)`. 

I think the last option is probably my favorite. 

(EDIT: commit added to pick the last option.)

### DIFF
--- a/packages/catlog/src/dbl/model.rs
+++ b/packages/catlog/src/dbl/model.rs
@@ -42,7 +42,7 @@ use serde::{Deserialize, Serialize};
 use tsify::Tsify;
 
 use super::theory::DblTheory;
-use crate::one::{Category, FgCategory, InvalidPathEq, Path, PathEq};
+use crate::one::{Category, FgCategory, InvalidPathEq, Path};
 use crate::tt::util::pretty::*;
 use crate::zero::{Namespace, QualifiedName};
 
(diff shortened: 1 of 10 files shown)


packages/catlog/src/dbl/model.rs
```diff
@@ -179,15 +179,28 @@ pub trait MutDblModel: FgDblModel {
...
+    }
+
+    /// Pretty prints an equation in the model.
+    fn eqn_to_doc<'a>(
+        &self,
+        _eqn: &PathEq<Self::Ob, Self::MorGen>,
+        _ob_ns: &Namespace,
+        _mor_ns: &Namespace,
+    ) -> D<'a> {
+        unreachable!() //Default implementation models have no equations, so won't call this.
```
KevinDCarlson: I'm leaving this here because the cargo fmt mysteriously wouldn't catch it and it drove me crazy for half an hour

PR #1087, opened 2026-03-02
Enable inline KaTeX in rich text editor

jmoggr:
This works around issues that automerge-prosemirror has with text in inline nodes by putting the text on an attribute and using a custom node view to capture the input.

### DIFF
--- a/packages/frontend/src/components/rich_text_editor/rich_text_editor.tsx
+++ b/packages/frontend/src/components/rich_text_editor/rich_text_editor.tsx
@@ -33,6 +33,7 @@ import {
     increaseListIndet,
     insertLinkCmd,
     insertMathDisplayCmd,
+    insertMathInlineCmd,
     toggleBulletList,
     toggleOrderedList,
     turnSelectionIntoBlockquote,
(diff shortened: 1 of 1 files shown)


epatters (changes requested):
Thanks Jason. In my limited testing so far, it works! Excited that this now possible, even if you're having to work around upstream issues.

Feature request:

- [ ] Add a toolbar button for inline math next to the existing button for display math (this is largely just to advertise that the feature exists)

Usability issues:

- [ ] When editing inline math, the left/right arrow keys cannot be used to exit on either side, which is counterintuitve
- [ ] Possibly related, when inline math is the first content in the text field, it seems not possible to move the cursor to the left of it
- [ ] If you click outside the inline math while editing, any changes are discarded rather than committed, which is counterintuitive

Cosmetic issues:

- [ ] When editing inline math, the opening `$` is right next to the text but there is a gap between between the text and closing `$`. Ideally this would be consistent.

jmoggr:
> Add a toolbar button for inline math next to the existing button for display math (this is largely just to advertise that the feature exists)

I'm not sure what icon to use for this, I used "variable": `(x)`

>  When editing inline math, the left/right arrow keys cannot be used to exit on either side, which is counterintuitve

fixed!

>  Possibly related, when inline math is the first content in the text field, it seems not possible to move the cursor to the left of it

Unfortunately this is not related, but appears to be due to a 6yo bug in firefox: [prosemirror issue](https://github.com/ProseMirror/prosemirror/issues/1091). I tried a couple of workarounds but I'm not confident they were pareto optimal. This works as expected in Chrome.

>  If you click outside the inline math while editing, any changes are discarded rather than committed, which is counterintuitive

fixed!

>  When editing inline math, the opening $ is right next to the text but there is a gap between between the text and closing $. Ideally this would be consistent.

fixed!

PR #1088, opened 2026-03-03
Add postgres and netcat to nix dev shell

kasbah:


jmoggr (approved):
<3 that means your using the nix shell! You will be assimilated.

PR #1091, opened 2026-03-03
Fix UUIDs being shown for unnamed morphisms in equations

### LINKED ISSUE #986: Don't show UUIDs in mass-action dynamics equations analysis
**Describe the bug**
UUIDs are shown in the mass-action dynamics equations analysis.

**Screenshots**
From https://next.catcolab.org/analysis/019bb369-eb0a-73b1-bd2c-445f50da5156

Image

kasbah:
fixes #986

### DIFF
--- a/packages/catlog-wasm/src/analyses.rs
+++ b/packages/catlog-wasm/src/analyses.rs
@@ -172,19 +172,16 @@ pub(crate) fn mass_action_equations_modal(
 
 #[cfg(test)]
 mod tests {
-
     use catlog::dbl::modal::{List, ModalMorType, ModalOb, ModalObType};
     use catlog::dbl::model::{ModalDblModel, MutDblModel};
     use catlog::simulate::ode::LatexEquation;
-    use catlog::stdlib::analyses::ode;
-    use catlog::stdlib::theories::th_sym_monoidal_category;
+    use catlog::stdlib::{analyses::ode, theories};
     use catlog::zero::{LabelSegment, Namespace, QualifiedName};
     use std::rc::Rc;
     use uuid::Uuid;
 
     use super::*;
-    use crate::model::tests::backward_link;
-    use crate::model::{DblModel, DblModelBox};
+    use crate::model::{DblModel, tests::backward_link};
 
     #[test]
     fn unbalanced_mass_action_latex_equations() {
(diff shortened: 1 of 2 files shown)


packages/catlog-wasm/src/model.rs
```diff
@@ -411,6 +411,55 @@ impl DblModel {
...
+                Some((dom, cod))
+            }
+        }
+    }
+
+    /// Gets a label string for a modal object, handling products/tensors.
+    ///
+    /// For a generator like `S`, returns `"S"`. For a product like `tensor([S, I])`, returns
+    /// `"[S, I]"`.
+    fn modal_ob_label_string(&self, ob: &ModalOb) -> Option<String> {
```
epatters: With the changes suggested above, this whole method can be deleted.

packages/catlog-wasm/src/model.rs
```diff
@@ -411,6 +411,55 @@ impl DblModel {
...
+
+    /// Gets label strings for the domain and codomain of a morphism generator.
+    ///
+    /// Returns `Some((dom_label, cod_label))` when the morphism has a domain
+    /// and codomain whose labels can be resolved from the namespace.
+    pub fn mor_generator_dom_cod_label_strings(
+        &self,
+        id: &QualifiedName,
+    ) -> Option<(String, String)> {
+        match &self.model {
```
epatters: You shouldn't need to special case for each kind of model. Do something like:

```rust
let (dom, cod) = all_the_same!(match &self.model {
  DblModelBox::[Discrete, DiscreteTab, Modal](model) => {
    (Quoter.quote(model.get_dom(id)?), 
     Quoter.quote(model.get_cod(id)?))
  }
})?;
match (dom, cod) {
  (Ob::Basic(dom_id), Ob::Basic(cod_id)) => {
    // Get labels from `ob_namespace`
  }
  _ => None
}
```

packages/catlog-wasm/src/analyses.rs
```diff
@@ -178,4 +200,140 @@ mod tests {
...
+            LatexEquation {
+                lhs: "\\frac{\\mathrm{d}}{\\mathrm{d}t} \\text{yyy}".to_string(),
+                rhs: "(\\rho_{\\text{xxx} \\to \\text{yyy}}) \\text{xxx} \\text{yyy}".to_string(),
+            },
+        ];
+        assert_eq!(equations, expected);
+    }
+
+    #[test]
+    fn petri_net_unnamed_transition_equations() {
```
epatters: With the suggested below, you can delete this whole verbose test case because the same code pathways will be used for all kinds of models, including models of modal theories.

packages/catlog-wasm/src/model.rs
kasbah: I refactored it with 37cb24bdfbd. I am a little confused about what the different `Ob` types represent and when they occur, to be honest. I do definitely seem to need to handle more cases than `Ob::Basic` as `Some`.

packages/catlog-wasm/src/analyses.rs
kasbah: This case still covers the case of the original bug report where we have multiple objects e.g. `[S, I]` as dom/cod. I renamed it and cut it down a bit with 3e9869cdc60395 , but it does seem like we should keep it.

epatters: Ah yes, my mistake.

packages/catlog-wasm/src/model.rs
```diff
@@ -411,6 +411,51 @@ impl DblModel {
...
+            Ob::Basic(s) => {
+                let name = QualifiedName::deserialize_str(s).ok()?;
+                Some(self.ob_namespace.label_string(&name))
+            }
+            Ob::App { ob, .. } => {
+                // FIXME: This is incorrect in general. The design issue is that
+                // this pretty printer claims to handles all models, but is
+                // customized to Petri nets as free SMCs where we prefer to omit
+                // the tensor application.
+                self.ob_label_string(ob)
```
epatters, line 447: I've noted an underlying problem with our current design.

I'm going to merge this PR as is since it's an improvement over the current state of affairs as we prepare for a release. I've created #1096 to track further improvements.

PR #1105, opened 2026-03-09
Run analysis simulations in webworkers

jmoggr:


kasbah:
Is there any indication that this is faster than the non web-worker execution?

jmoggr:
Faster at running the simulations? Probably not. I didn't measure, but I didn't notice a meaningful difference. If anything I would expect the simulations to be slightly slower given IO and some amount of duplicated work. The webworker can't run the blocking simulation any faster than the main browser thread. 

Faster at producing a responsive page? Yes. Before analysis page I was testing with froze for ~5s after loading until all the simulations were completed, now the page becomes responsive immediately. I think we still spending ~100ms to kick all this off from the main thread, but I think that's acceptable.

Unfortunately the analysis Tim shared with me for testing is not public: http://localhost:5173/analysis/019cafd4-19fb-7b62-9e11-ceda2da5da67

---

There is another major benefit beyond speed: This also allows you to change parameters while the simulation is running, cancelling the previous run. Before it was relatively easy to cause a simulation to permanently freeze a page (exponential O(n) in the simulation).

packages/frontend/src/stdlib/analyses/kuramoto.tsx
```diff
@@ -7,19 +7,18 @@ import {
...
 import { createModelODEPlot } from "./model_ode_plot";
-import type { KuramotoSimulator } from "./simulator_types";
 
 import "./simulation.css";
 
 /** Analyse a model using first- or second-order Kuramoto dynamics. */
 export default function Kuramoto(
     props: ModelAnalysisProps<KuramotoProblemData> & {
-        simulate: KuramotoSimulator;
+        analysisId: string;
```
kasbah, line 21: So we need to go stringly typed for this? Can we not at least have the exact string literals on both sides?

```ts
"petri-net" | "primitive-stock-flow" | "primitive-signed-stock-flow" | ...
``` 

I also think a better name would be `analysisType` .

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
```diff
@@ -1,94 +1,257 @@
...
+                    .catch((err) => {
+                        if (requestId !== latestRequestId) {
+                            return;
+                        }
+                        setLoading(false);
+                        setData(() => mapErr(String(err)));
+                    });
+            }, DEBOUNCE_MS);
+        }),
+    );
```
kasbah, line 192: We should have a `debounce` utility and use that.

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
```diff
@@ -1,94 +1,257 @@
...
+        (error) => ({
+            plotData: { tag: "Err", content: error },
+            latexEquations: [],
+        }),
     );
 }
 
 /** Reactively compute the symbolic ODE equations for a model in LaTeX.
 
+This stays on the main thread as it is typically fast.
```
kasbah: Not sure this comment is useful.

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
```diff
@@ -1,94 +1,257 @@
...
+                const requestId = ++requestIdCounter;
+                latestRequestId = requestId;
+
+                const request: SimulationRequest = {
+                    requestId,
+                    theoryId,
+                    analysisId,
+                    notebook: JSON.parse(JSON.stringify(currentNotebook)),
+                    refId,
+                    params: JSON.parse(JSON.stringify(currentParams)),
```
kasbah: I believe `worker.postMessage` does a structured clone so this is a solid store (proxy object) issue so I think it's correct to use `solid-js/store`'s `unwrap`: https://github.com/ToposInstitute/CatColab/pull/1113/changes/a5961cf6bc5cec05329f504438557c9f090f1bfc

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
```diff
@@ -1,94 +1,257 @@
...
-export function createModelODEPlot(
-    validatedModel: Accessor<ValidatedModel | undefined>,
-    simulate: (model: DblModel) => ODEResult,
-) {
-    return createMemo<JsResult<ODEPlotData, string> | undefined>(
-        () => {
-            const validated = validatedModel();
-            if (validated?.tag !== "Valid") {
+    const theoryId = liveModel.liveDoc.doc.theory;
+    const refId = liveModel.liveDoc.docHandle.documentId;
```
kasbah: I'm still trying to figure out if this could actually occur in our codebase but there are some reactivity concerns around the captured `theoryId` and `refId`. I made a test and fix for it in https://github.com/ToposInstitute/CatColab/pull/1113

epatters:
> Is there any indication that this is faster than the non web-worker execution?

Right, to build on Jason's response, this isn't about making the simulations faster per se but making the rest of the application be more responsive while simulations run.

If you really want to stress test this feature, try to doing a loop finding analysis into a large causal loop diagram, with a high upper limit on the lengths of the paths considered. Because the homomorphism finding has combinatorial complexity, this blows up quickly. In fact, I had to make the fix #419 a while back because the whole browser would freeze if you didn't impose a limit on the path length.

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
kasbah: Having thought about it for a little while. It seems unlikely we will hit this (i.e. theory or refId switched out from under and analysis block) but it seems like better general practice to be ready for it nonetheless.

jmoggr: >  It seems unlikely we will hit this

I traced it through and did not find any cases where the `theoryId` and `refId` changed independently of the analysis block, but the concern is warranted.

As a possible middle ground I'm re-reading `theoryId` and `refId` inside the debounced callback, so they would be re-read if the analysis block changes. I think this is neater than plumbing them through and just as safe/correct.

If this changes is acceptable to you, then I think it makes the test you wrote no longer relevant?

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
jmoggr: Thanks, TIL. I made the change.

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
jmoggr: Removed

packages/frontend/src/stdlib/analyses/kuramoto.tsx
jmoggr: This comes from [`AnalysisOptions`](https://github.com/ToposInstitute/CatColab/blob/8b27fa3bf7f3f9e5ff1a1a6f6a0a3e97d4de09d7/packages/frontend/src/stdlib/analyses.tsx#L17) where `id` is a string. The analysis factory functions take a `Partial<AnalysisOptions>` which implies that the `id` can be overridden. This existing behavior seems kind of wrong though, since the id must concretely match something for the simulation dispatch to work. 

It does not look like we override the id anywhere.

I'd like to keep this as is in this PR, I'll stack another PR with a proposed change that tightens up these types.

packages/frontend/src/stdlib/analyses/model_ode_plot.ts
jmoggr: Done: https://github.com/ToposInstitute/CatColab/pull/1105/changes/a6e9ddb12108e65afe1778077241b5e2818e2f89#diff-b524eb3833e63e30ee343fc87acc396648d22ce2603e0ea0cb3b72204b1da776

kasbah:
Needs a rebase/merge and would be good to get @epatters eyes on this too.

epatters:
Yep, I'm planning to review this one too.

PR #1111, opened 2026-03-10
Nix rust toolchain fixups

kasbah:
- We use nightly `rustfmt` in CI so we should line these up. 
- `normalize_comments` can just be part of the config (not really nix related)
- Remove duplicate definitions between `rust-toolchain.toml` (used by `flake.nix`) and the `flake.nix` file itself

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -116,7 +116,6 @@
           buildInputs =
             with pkgs;
             [
-              clippy
               darkhttpd
               esbuild
               lld
@@ -133,7 +132,6 @@
               python312Packages.jupyter-server
               python312Packages.requests
               python312Packages.websocket-client
-              rust-analyzer
               rustToolchain
               nightlyRustfmt
               sqlx-cli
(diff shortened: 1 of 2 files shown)


jmoggr:
Considering the bigger picture for a moment: If we're open to nixifying more, the jobs `run_tests`, `rust_formatting` and `rust_lints` could be made consistent with `check_rust_docs`, or even make them all part of the same nix check.

This is probably out of scope, but I'd just like to put it out there as we consider what it means for things to be consistent.

kasbah:
I was surprised by the nightly for `fmt` a while ago but there was a reason for it and @epatters insisted on it.

When not in nix, if using rustup one can do `cargo +nightly fmt`.

epatters:
> Should we be consolidating on the nightly `rustfmt`? The use of nightly in CI was a something we did previously in CI, and not removing it fully was probably a bug.

We're using nightly `rustfmt` for the comment normalization feature. While I acknowledge there can be a range of reasonable opinion on the tradeoffs, ultimately I prefer having a clean and consistently formatted codebase even if it requires a little CI jankery.

PR #1115, opened 2026-03-11
Add check for backend command `generate-bindings`

kasbah:
Checks that we ran generate-bindings i.e. no diff in the output when it is run again.

### DIFF
--- a/packages/backend/pkg/src/index.ts
+++ b/packages/backend/pkg/src/index.ts
@@ -13,6 +13,20 @@
 
 */
 import type { Query, Mutation, Subscription } from "@qubit-rs/client";
+export type JsonValue = number | string | boolean | Array<JsonValue> | { [key in string]?: JsonValue } | null;
+export type NewPermissions = { 
+/**
+ * Base permission level for any person, logged in or not.
+ */
+anyone: PermissionLevel | null, 
+/**
+ * Permission levels for users.
+ *
+ * A mapping from user IDs to permission levels.
+ */
+users: { [key in string]?: PermissionLevel }, };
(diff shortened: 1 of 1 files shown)


packages/backend/src/main.rs
```diff
@@ -80,22 +97,7 @@ async fn main() {
...
-
-            rpc::router()
-                .as_codegen()
-                .write_type(&path, TypeScript::new())
-                .expect("Failed to write TypeScript bindings");
-
-            info!("Successfully generated TypeScript bindings to: {}", path.display());
-            return;
-        }
+        Command::GenerateBindings => unreachable!(),
```
kasbah, line 100: I moved this so we can run generate-bindings without the DATABASE_URL env variable.

jmoggr:
I'm wondering if this should become part of the nix build process instead of a check. We have a similar pattern for catcolab-wasm: in local dev it's generated with a pnpm/cargo command, when deployed it's built as a Nix package. I supposed a possibly meaningful difference is that the generate artifact is committed, but IMO we probably shouldn't be doing that anyway.

The change itself fine if we don't want to open the bigger can of worms right now.

kasbah:
We can talk about that separately but I am only aiming for a quick fix of the existing dev experience with this.

jmoggr:
What exactly was the issue that this was meant to solve? I could see a couple ways this could help, but I'm wondering what specifically motivated this?

kasbah:
I want to freely run `generate-bindings` in the dev CLI (#1106), because it's a necessary build step, so it's annoying that the files are changing when I do that.

PR #1117, opened 2026-03-12
Visualize composition pattern as UWD

### LINKED ISSUE #1021: Visualize undirected wiring diagrams using elkjs
Following up #133 and #1019, we should be able to visualize the UWDs implicit in compositional notebooks using ELK. As a proof of concept, see this [small example](https://rtsys.informatik.uni-kiel.de/elklive/elkgraph.html?compressedContent=IYGw5g9gTglgLgCwLYC4AEJgE8CmUcAmAUEQHYQE5oBm+OACtHAM5oDeRaXaADkwMIRSzOFGAxSLdADEAkgA0AogBEA+gGVZyxZ258ocAIIgYYUkhySAdJWrAAriDjpls9QBUASrIBCAVXcVXS5MXHwCK2YeYABjCTArckoAOQocHxw4AHccSwAZbDxmdABWAAZgtBwQAGsraIICePQAIhAcajgAXgBGMoAaNFgwBG7ywbgIHl6BtAAjCDhJpBmWkm5eJl4e9jQAPX04SJhKdEVDDzQAX0rD3gBmXYOmY9O0c8ubyqSqUh2ODYbQ6CYSicSSYpoORKNSabSVIFbHj-faHV44M4XdzXBF6JEAJieaOYJwx7yxOI2VzQ3zSaFIhIBgK4wKEIjEEikUIUKg0Wh0zJZSJRzwM6Mxn1xQoMvEZqJeJLeH2xNyplUIYF+PXqOwAtAA+NAAKx66oImvp2p4hINxvxZotDJ1aFtRvtGw1v3x9RthqN93W3B+xv+Uow2Ag9jgaAA2mgSQAvMklQYlNAAXUqqqDdLd7DDoUj0bjieTqYzWdplGNjyZzMLUdj8ZgSdK5czao95qoJpdhuRDp7j1tPADGyIVyAA) I cooked up by hand.

kasbah:
closes #1021 

image

https://branch-kb-composition-pattern-viz--catcolab.netlify.app/model/019cd8bf-5bb3-7003-b929-e41633daec59/analysis/019cd8c1-3d66-79e1-a079-a803696c4476

### DIFF
--- a/packages/frontend/src/help/logics/causal-loop-delays.mdx
+++ b/packages/frontend/src/help/logics/causal-loop-delays.mdx
@@ -27,8 +27,6 @@ import { HelpAnalysisById } from "../logic_help_detail"
 
 <HelpAnalysisById theory={props.theory} analysisId="diagram"/>
 
-<HelpAnalysisById theory={props.theory} analysisId="composition-pattern"/>
-
 <HelpAnalysisById theory={props.theory} analysisId="negative-loops">
 Balancing loops are composites of paths within the diagram that start and finish at the same point, and whose overall sign (given by multiplying all the signs of each constituent path) is positive.
 Balancing loops tend to be stabilising.
--- a/packages/frontend/src/stdlib/theories/causal-loop-delays.ts
+++ b/packages/frontend/src/stdlib/theories/causal-loop-delays.ts
@@ -68,7 +68,6 @@ export default function createCausalLoopDelaysTheory(theoryMeta: TheoryMeta): Th
                 description: "Visualize the causal loop diagram",
                 help: "visualization",
             }),
-            analyses.compositionPattern(),
             analyses.motifFinding({
                 id: "negative-loops",
                 name: "Balancing loops",
--- a/packages/frontend/src/visualization/undirected_wiring_diagram_elk.ts
+++ b/packages/frontend/src/visualization/undirected_wiring_diagram_elk.ts
@@ -103,11 +103,9 @@ export function uwdToElk(uwd: UWD | undefined): ElkNode {
 
     // Build a map from junction name to its outer port ID.
     const junctionOuterPort = new Map<string, string>();
-    if (uwd) {
-        for (const port of uwd.outerPorts) {
-            if (port.junction != null) {
-                junctionOuterPort.set(port.junction, outerPortId(port.name));
-            }
+    for (const port of uwd?.outerPorts ?? []) {
+        if (port.junction != null) {
+            junctionOuterPort.set(port.junction, outerPortId(port.name));
         }
     }
 
@@ -116,19 +114,16 @@ export function uwdToElk(uwd: UWD | undefined): ElkNode {
     // merged along common segments.
     const edges: ElkExtendedEdge[] = [];
     let edgeIndex = 0;
-
-    if (uwd) {
-        for (const box of uwd.boxes) {
-            for (const port of box.ports) {
-                const outerPort =
-                    port.junction != null ? junctionOuterPort.get(port.junction) : undefined;
-                if (outerPort) {
-                    edges.push({
-                        id: `wire-${edgeIndex++}`,
-                        sources: [outerPort],
-                        targets: [boxPortId(box.name, port.name)],
-                    });
-                }
+    for (const box of uwd?.boxes ?? []) {
+        for (const port of box.ports) {
+            const outerPort =
+                port.junction != null ? junctionOuterPort.get(port.junction) : undefined;
+            if (outerPort) {
+                edges.push({
+                    id: `wire-${edgeIndex++}`,
+                    sources: [outerPort],
+                    targets: [boxPortId(box.name, port.name)],
+                });
             }
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ announcement and a blog post. Minor versions are not announced but allow
 features and fixes to be released with greater frequency. Minor versions often
 include notable new features.
 
+## [Unreleased]
+
+### Added
+
+- New analysis: visualize composition pattern of instantiated models as an
+  undirected wiring diagram
+  ([#1117](https://github.com/ToposInstitute/CatColab/pull/1117/))
+
 ## [v0.5.0](https://github.com/ToposInstitute/CatColab/releases/tag/v0.5.0) (2026-03-09)
 
 ### Added
--- a/packages/frontend/src/help/logics/causal-loop.mdx
+++ b/packages/frontend/src/help/logics/causal-loop.mdx
@@ -25,8 +25,6 @@ Variables can "encourage" or "discourage" each other, determined by the sign (po
 
 <HelpAnalysisById theory={props.theory} analysisId="diagram"/>
 
-<HelpAnalysisById theory={props.theory} analysisId="composition-pattern"/>
-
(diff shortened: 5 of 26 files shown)


packages/frontend/src/stdlib/theories/causal-loop-delays.ts
```diff
@@ -68,6 +68,7 @@ export default function createCausalLoopDelaysTheory(theoryMeta: TheoryMeta): Th
                 description: "Visualize the causal loop diagram",
                 help: "visualization",
             }),
+            analyses.compositionPattern(),
```
epatters: This analysis should be applicable to models of *any* theory (well, for now any theory that `canInstantiateModels`, which will eventually be all of them).

So, rather than manually adding this analysis to every theory specification, it should be automatically injected into the list of constructors, like we do for [instantiations](https://github.com/ToposInstitute/CatColab/blob/b1b229f4b72c9ad445ffd7bc0977d3a877f52035/packages/frontend/src/model/model_editor.tsx#L99).

packages/frontend/src/help/logics/causal-loop-delays.mdx
```diff
@@ -27,6 +27,8 @@ import { HelpAnalysisById } from "../logic_help_detail"
 
 <HelpAnalysisById theory={props.theory} analysisId="diagram"/>
 
+<HelpAnalysisById theory={props.theory} analysisId="composition-pattern"/>
```
epatters: Ditto about these: we shouldn't manually add this analysis to the help page for every logic.

packages/frontend/src/visualization/undirected_wiring_diagram_elk.ts
epatters, line 1: I can imagine arguments for either, but I think I'd prefer to put these files `undirected_wiring_diagram_*` in `src/visualization` rather than `src/stdlib/analyses`. (The file `composition_pattern.tsx` should of course remain here.)

packages/frontend/src/visualization/undirected_wiring_diagram_elk.ts
```diff
@@ -0,0 +1,326 @@
...
+
+/** Layout of a box in a UWD. */
+export interface UwdBoxLayout {
+    x: number;
+    y: number;
+    width: number;
+    height: number;
+    label: string;
+    ports: UwdPortLayout[];
+}
```
epatters: Not a big deal, but I'd be tempted to simplify the types slightly by removing `UwdOuterLayout`, making `label` optional in `UwdBoxLayout`, and then using `UwdBoxLayout` for both `outer` and `boxes` fields.

packages/frontend/src/stdlib/analyses/undirected_wiring_diagram_elk.ts
```diff
@@ -0,0 +1,326 @@
...
+        }
+    }
+
+    // Build wire edges: connect ports that share a junction directly to each
+    // other, creating edges between every pair of ports in each junction group.
+    const edges: ElkExtendedEdge[] = [];
+    let edgeIndex = 0;
+
+    for (const ports of junctionPorts.values()) {
+        for (let i = 0; i < ports.length; i++) {
```
epatters: Isn't the idiomatic way to write this `for (const source of ports)` since you don't otherwise need the indices `i` and `j`?

packages/frontend/src/stdlib/analyses/undirected_wiring_diagram_elk.ts
```diff
@@ -0,0 +1,326 @@
...
+            if (port.junction != null) {
+                const list = junctionPorts.get(port.junction) ?? [];
+                list.push(outerPortId(port.name));
+                junctionPorts.set(port.junction, list);
+            }
+        }
+    }
+
+    // Build wire edges: connect ports that share a junction directly to each
+    // other, creating edges between every pair of ports in each junction group.
```
epatters: Wow, I didn't realize that ELK would handle this correctly on its own: I figured you'd have to manually create the junction nodes.

Have you tested that it works as expected when more than two ports point to the same junction?

packages/frontend/src/stdlib/analyses/undirected_wiring_diagram_elk.ts
```diff
@@ -0,0 +1,326 @@
...
+            "elk.direction": elkDirection,
+        },
+    };
+}
+
+/** Parse the ELK layout result into a `UwdLayout`. */
+export function parseUwdElkLayout(root: ElkNode): UwdLayout {
+    const outerElk = root.children?.[0];
+    if (!outerElk) {
+        return emptyLayout();
```
epatters: Is this special handling actually needed? Seems like you'd get the same result by replacing `outerElk.` with `outerElk?.` below.

packages/frontend/src/visualization/undirected_wiring_diagram_elk.ts
```diff
@@ -0,0 +1,326 @@
...
+            height: outerHeight,
+            ports: outerPorts,
+        },
+        boxes,
+        wireEdges,
+    };
+}
+
+/** Parse the layout of an ELK port relative to a parent offset. */
+function parsePortLayout(port: ElkPort, parentX: number, parentY: number): UwdPortLayout {
```
epatters: Might be premature generalization, but it seems like this functionality is generic enough to go into `elk.ts`.

packages/frontend/src/stdlib/theories/causal-loop-delays.ts
kasbah: I can't say I'm a fan of how dynamic this approach is. Is this https://github.com/ToposInstitute/CatColab/commit/b2dd99b11edc40b8c0236f6b7b9f7eb07d4f75a3 sort of what you are thinking about or do you think I'm still missing a nicer way to do this?

packages/frontend/src/stdlib/analyses/undirected_wiring_diagram_elk.ts
kasbah: Ah, no you are right. Somehow I got very fixated on just testing the simple case. I will need to re-work this.

kasbah:
I've updated this to properly handle multiple edges going to one outer port.

image


I've addressed all feedback except I'm still not totally clear on the best strategy regarding the injection of the analysis block capability https://github.com/ToposInstitute/CatColab/pull/1117#discussion_r2926299879

packages/frontend/src/visualization/undirected_wiring_diagram_elk.ts
kasbah: I moved a bunch of the types and this helper into elk.ts

packages/frontend/src/stdlib/analyses/undirected_wiring_diagram_elk.ts
kasbah: There was a reason for `i` and `j` here because we needed an offset but could have probably been done better. This was replaced wholesale in re-working for the junctions.

packages/frontend/src/stdlib/theories/causal-loop-delays.ts
epatters: I agree. I've pushed a commit with a more declarative and more general approach. Still room for improvement but I think this is good enough for now.

kasbah: Looks much better, thanks!

PR #1121, opened 2026-03-12
Fix Safari font measurement for ELK visualizations

kasbah:
Relative units like `rem` are resolved relative to the canvas element itself (or the document root for rem). When the canvas is not in the DOM there's no inherited styling context. The behavior becomes browser-dependent:
- Chrome/Firefox: Tend to resolve `1rem` against the document root's font size even for detached canvases, so it "works" 
- Safari: Appears to handle this more strictly it seems to fail to resolve `rem `properly producing incorrect measurements.

### DIFF
--- a/packages/frontend/src/visualization/font_utils.ts
+++ b/packages/frontend/src/visualization/font_utils.ts
@@ -5,9 +5,9 @@ import invariant from "tiny-invariant";
  * Returns a font specification string suitable for Canvas 2D context.
  */
 export function getMainFont(): string {
-    return `1rem ${getComputedStyle(document.documentElement)
-        .getPropertyValue("--main-font")
-        .trim()}`;
+    const style = getComputedStyle(document.documentElement);
+    const rootFontSize = parseFloat(style.fontSize);
+    return `${rootFontSize}px ${style.getPropertyValue("--main-font")}`;
 }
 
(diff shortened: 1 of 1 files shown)


packages/frontend/src/visualization/font_utils.ts
```diff
@@ -5,7 +5,8 @@ import invariant from "tiny-invariant";
  * Returns a font specification string suitable for Canvas 2D context.
  */
 export function getMainFont(): string {
-    return `1rem ${getComputedStyle(document.documentElement)
+    const rootFontSize = parseFloat(getComputedStyle(document.documentElement).fontSize);
+    return `${rootFontSize}px ${getComputedStyle(document.documentElement)
```
epatters: Can we make this a little cleaner by avoid redundant calls?

```typescript
const style = getComputedStyle(document.documentElement);
const rootFontSize = parseFloat(style.fontSize);
return `${rootFontSize}px ${style.get[...]}`;
```

epatters (changes requested):
Thanks for the fix! I had a bad feeling about that `1 rem`...

packages/frontend/src/visualization/font_utils.ts
kasbah: Done, thanks.

PR #1122, opened 2026-03-12
Move RFCs from the /math/ forest to /rfc/ Quarto

### LINKED ISSUE #1120: RFCs should be available at /rfc/
Currently, RFCs live in the `math-docs` forest, which means that they are accessed at `next.catcolab.org/math/rfc-XXXX.xml`. Given that not all RFCs will be mathematical, and that we (by design) do not expect RFCs to be non-linear complex documents with transclusions etc., we should

1. move RFCs to `next.catcolab.org/rfc/XXXX`
2. migrate to some markdown format for RFCs (most likely with some static-site generator, maybe Quarto)

tim-at-topos:
closes #1120

### DIFF
--- a/rfc/0002.md
+++ b/rfc/0002.md
@@ -30,8 +30,8 @@ writing the RFC was to understand the implementation well enough for it to be
 extended by the author and others. As documentation, it is not really accurate.
 It does not describe everything in the original implementation, while also it
 specifies new features (lists, operations) not present in the original
-implementation. In that sense, it *is* in the spirit of an RFC, representing
-a design at a particular point in time, as understood by a particular person.
+implementation. In that sense, at least, this document *is* in the spirit of an
+RFC, as it represents a design for DoubleTT at a particular point in time.
 :::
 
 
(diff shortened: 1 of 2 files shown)


tim-at-topos:
@epatters This is essentially finished — making things look nice with better fonts etc can always come later — but I haven't really written anything on the "About" page. Feel free to add more words there if you have some ideas 🤷 

I'll mark this as ready for review once I've merged RFC-0002 :-)

tim-at-topos:
@epatters the automatic deploys for this are essentially useless because of the routing in `App.tsx`:

```typescript
    {
        path: "/rfc/*",
        component: (route) => {
            const url = `https://next.catcolab.org${route.location.pathname}`;
            window.location.replace(url);
            return null;
        },
    },
```

Two options for viewing this before merging:

1. I push a temporary commit that removes this, so you don't get redirected to the as-yet non-existent `next.catcolab.org/rfc/`, and then add this back in afterwards
2. You build and view this locally (e.g. just run `quarto preview` in the `/rfc/` directory) and then trust that the routing works (and hey ho the dev build is for dev so it's ok)
3. Bonus option: you suggest a better idea because I can only think of the two above :-)

epatters:
Thanks Tim! I'll just build and view it locally.

PR #1128, opened 2026-03-16
Migrate CI/CD to Nix, serve frontend + docs from backend instead of Netlify

jmoggr:
For consideration: this is how I would structure the CI of this project if I was starting it from scratch. This is meant to demonstrate an alternative structure that meets my personal definition of "clean" that I keep alluding to in my reviews of CI changes.

The high level changes are:
- If it can be done in Nix, it's done in Nix
  - The entire build can be done locally, the github action has no input on the build process
  - Reduces the amount of code that is only runnable in CI
  - Removes the need for an explicit build step/workflow in CI, in Nix building and running checks are the same thing 
- Host frontend ourselves instead of netlify
  - Ensures that the the frontend and backend deployments are always in sync
  - We still have the netlify frontend previews for pull requests
  - Simplifies the deploy process
  - Removes an external dependency and makes the whole application self contained
- Deploy both staging and production from github using the same github code

Note: this hasn't been tested deployed and would require some more work if we actually wanted to use any of this.


jmoggr:
The Netlify preview works again.

PR #1129, opened 2026-03-16
Update user state from autosave listener instead of DB trigger

jmoggr:
Extract snapshot relations and update user state docs directly in the Rust autosave listener, eliminating the pg_notify round trip that re-read the snapshot from the database.


packages/backend/src/user_state.rs
```diff
@@ -13,6 +13,9 @@ use crate::app::{AppError, AppState};
 use crate::auth::PermissionLevel;
 use crate::autosurgeon_datetime::{datetime_millis, option_datetime_millis};
 
+/// Default name for documents without a name.
+pub(crate) const DEFAULT_DOC_NAME: &str = "untitled";
```
kasbah: Is there a reason that you are using `pub(crate)` for these attributes and `fn`s you are adding? Since we are exposing the crate's API only for integration testing and for use in `main.rs` this reads to me as "I don't want this available to integration tests or main", which doesn't seem right, but I might be missing something so wanted to ask.

PR #1132, opened 2026-03-16
Serve algjulia-interop as a service through the backend

### LINKED ISSUE #1109: Remote compute running Julia and AlgebraicJulia interop process
With https://github.com/ToposInstitute/CatColab/pull/835, we have the ability to start a local server which runs Julia, has many AlgebraicJulia packages loaded in its environment, receives JSON messages from CatColab analyses, and sends Julia computations back to CatColab. In order for more people to use this functionality, it would be helpful if there were an AWS (or other cloud compute server) which is simply always running this AlgebraicJulia interop server. 

It would have to be limited in how much compute we give it if it's open to the public, and perhaps there could be credentials for an instance that's slightly less limited which just devs have access to.

jmoggr:
Closes #1109 

I've tested this in the VM, but I expect there to be issues when it's first deployed. I don't think we should merge this until I/we have confirmed that it can be successfully deployed to staging manually, otherwise it will likely break CI.

Running other peoples code on servers I am responsible for makes me nervous. I have hardened the systemd service for julia to the best of my ability and am going to have an LLM spend the night iterating through all of the other options to find the minimal set of permissions that still works. This is probably better isolation than what docker offers by default and I don't think we can do much better without full virtualization. I might look into options for hardening the kernel as well, I've not done that on Nix and Nix might make that easy and worth doing.

### DIFF
--- a/infrastructure/modules/catcolab/services.nix
+++ b/infrastructure/modules/catcolab/services.nix
@@ -185,16 +185,29 @@ with lib;
         StateDirectory = lib.removePrefix "/var/lib/" juliaDepotPath;
         ExecStartPre = "${juliaFhsPkg}/bin/julia-interop --project=${juliaProjectPath} -e 'using Pkg; Pkg.instantiate()'";
         ExecStart = "${juliaFhsPkg}/bin/julia-interop --project=${juliaProjectPath} --threads auto ${juliaProjectPath}/scripts/endpoint.jl Catlab";
+        # Security hardening. The following are incompatible and excluded:
+        # - MemoryDenyWriteExecute: Julia's JIT and PCRE regex JIT need W^X memory.
+        # - ProtectProc/ProcSubset: bwrap reads /proc/sys/kernel/overflowuid.
+        # - RestrictNamespaces: bwrap needs user namespaces.
+        # - PrivateDevices: breaks bwrap device access.
         NoNewPrivileges = true;
-        ProtectSystem = "full";
+        ProtectSystem = "strict";
         ProtectHome = true;
         PrivateTmp = true;
         ProtectKernelTunables = true;
         ProtectKernelModules = true;
+        ProtectKernelLogs = true;
         ProtectControlGroups = true;
+        ProtectClock = true;
+        ProtectHostname = true;
         RestrictSUIDSGID = true;
-        LockPersonality = true;
         RestrictRealtime = true;
+        RestrictAddressFamilies = [ "AF_INET" "AF_INET6" "AF_UNIX" ];
+        LockPersonality = true;
+        RemoveIPC = true;
+        SystemCallArchitectures = "native";
+        CapabilityBoundingSet = "";
+        UMask = "0077";
         ReadWritePaths = juliaDepotPath;
       };
     };


infrastructure/julia.nix
```diff
@@ -0,0 +1,43 @@
+# Julia packaging for CatColab: pre-built binary wrapped in an FHS environment.
+#
+# Julia doesn't work well with pure Nix packaging because it expects a standard
+# Linux filesystem layout for native library loading. This fetches the official
+# binary tarball and wraps it in a buildFHSEnv that provides libstdc++ and other
+# native libraries needed by AlgebraicJulia packages.
+{ pkgs }:
+let
+  juliaVersion = "1.11.6";
```
epatters: We should bump the Julia version to v1.12, the latest stable release.

packages/backend/pkg/src/index.ts
epatters, line 1: This diff shouldn't be here, but in view of #1136, it should disappear anyway after rebasing off main.

packages/backend/Cargo.toml
```diff
@@ -24,6 +24,7 @@ notebook-types = { version = "0.1.0", path = "../notebook-types" }
 qubit = { version = "1.0.0-beta.0", features = ["ts-serde-json", "ts-uuid", "ts-chrono"] }
 rand = "0.8"
 regex = "1.11.1"
+reqwest = { version = "0.12", default-features = false, features = ["rustls-tls"] }
```
epatters: Any reason not to use the latest v0.13, which apparently also uses rustls by default?

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -93,7 +96,7 @@ export default function TabularView(
         },
 
         async ({ model, diagram }) => {
-            const response = await fetch("http://127.0.0.1:8080/acsetcolim", {
+            const response = await api.fetch("/julia/acsetcolim", {
```
epatters: I can imagine wanting to use a local Julia server when developing new analyses based on Julia. How should we handle that? One idea is add a Vite env variable defaulting to the above but settable to localhost if desired.

jmoggr: This reveals that there are now two methods of running Julia locally: with and without the backend. When the backend is running locally this code works fine with the Julia requests being proxied through the backend. However this did break the "traditional" method of running Julia locally.

I implemented your suggested fix and set the variable by default for local development so no workflows should need to change.

packages/backend/Cargo.toml
jmoggr: Nope, this was cut and pasted from another project of mine, which has just been habit after figuring out the TLS issue once. It's nice that it's the default now!

infrastructure/julia.nix
jmoggr: done

.github/workflows/build.yml
```diff
@@ -90,7 +91,7 @@ jobs:
         with:
           path: |
             packages/frontend/dist
-          key: frontend-build-${{ steps.cache-key.outputs.git-hash }}-${{ github.event_name == 'release' && 'production' || 'staging' }}
+          key: frontend-build-v2-${{ steps.cache-key.outputs.git-hash }}-${{ github.event_name == 'release' && 'production' || 'staging' }}
```
epatters: Is this `v2` supposed to be here?

jmoggr: sorry there was debugging stuff left in, I forgot to force push again.

epatters:
Great, this should be ready to go once the CI passes. It's not immediately obvious to me why it's failing.

jmoggr:
Fixed, it was a type error in a test file that didn't get included.

PR #1135, opened 2026-03-17
Fix ts-rs output stability

kasbah:
fixes #1133 

I am proposing fixes for ts-rs https://github.com/Aleph-Alpha/ts-rs/pull/489 and this uses our forks to pull this change in.

PR #1136, opened 2026-03-17
Generate backend API bindings instead of tracking them in git

jmoggr:
Benefits:
- permanently fixes issues around binding output stability
- no longer tracks generate artifacts in git
- follows same pattern as the artifacts for catlog-wasm

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -8,7 +8,7 @@
         "build:wasm": "npm --prefix ../catlog-wasm run build:browser",
         "build:bindings": "cargo run -p backend -- generate-bindings",
         "build:deps": "npm run build:wasm && npm run build:bindings",
-        "build:nix": "tsc -b && vite build --sourcemap true",
+        "build": "tsc -b && vite build --sourcemap true",
         "compile": "tcm -p 'src/**/*.module.css' . && tsc -b",
         "preview": "vite preview",
         "format": "biome format --write",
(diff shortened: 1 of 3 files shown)


packages/frontend/package.json
```diff
@@ -8,8 +8,9 @@
         "staging": "vite --mode staging",
         "dev": "vite --host",
         "build:wasm": "npm --prefix ../catlog-wasm run build:browser",
+        "build:bindings": "cargo run -p backend -- generate-bindings",
         "build:nix": "tsc -b && vite build --sourcemap true",
-        "build": "npm run build:wasm && npm run build:nix",
+        "build": "npm run build:wasm && npm run build:bindings && npm run build:nix",
```
kasbah: I think it would be convenient to have

```
"build:deps" : "npm run build:wasm && npm run build:bindings"
```

and also finally add this to dev (because I always forget)

```
"dev": "npm run build:deps && vite --host"
```

jmoggr: I agree, added.

packages/frontend/package.json
```diff
@@ -5,19 +5,21 @@
     "version": "0.1.0",
     "type": "module",
     "scripts": {
-        "staging": "vite --mode staging",
-        "dev": "vite --host",
         "build:wasm": "npm --prefix ../catlog-wasm run build:browser",
+        "build:bindings": "cargo run -p backend -- generate-bindings",
+        "build:deps": "npm run build:wasm && npm run build:bindings",
         "build:nix": "tsc -b && vite build --sourcemap true",
```
kasbah: Just while we are looking at this. Why is this one called "build:nix" when it's not got anything to do with nix directly?

jmoggr: It's used by `packages/frontend/default.nix`. We could just call `tsc -b && vite build --sourcemap true` from `frontend/default.nix` but I've been trying to keep all the npm build stuff located in one place with one point of entry.

I agree that it's non obvious, and the focus on a single point of entry was driven by other since-fixed painful architectural problem.

If you think we should get rid of it, then we should. I'm undecided and don't feel strongly either way.

epatters:
I assume this would obviate the need for #1135 and those upstream fixes?

kasbah:
Yep, and this needs your approval now that I've pushed some code here it seems.

PR #1138, opened 2026-03-17
Tighten up docs on hash mismatch

kasbah:
These are a bit too wordy for a situation where we normally "just want to get the check to pass". Also they talk about "instructions given below the error message" which don't seem to exist.

### DIFF
--- a/dev-docs/fixing-hash-mismatches.md
+++ b/dev-docs/fixing-hash-mismatches.md
@@ -6,19 +6,26 @@ title: "Fixing hash mismatches in Nix"
 
 #### pnpm Dependencies in the Frontend
 
-This only applies to the `frontend` package.
+This only applies to the `frontend` package. The pnpm hash is located in `packages/frontend/default.nix`
+within the `pkgs.fetchPnpmDeps` block.
 
-The following error occurs when a dependency has changed but the Nix hash has not:
+When a pnpm dependency has changed but the Nix hash has not, running `nix build .#frontend` will fail
+with a hash mismatch that looks like this:
 ```
-> ERR_PNPM_NO_OFFLINE_TARBALL  A package is missing from the store but cannot download it in offline mode. The missing package may be downloaded from https://registry.npmjs.org/@automerge/prosemirror/-/prosemirror-0.2.0-alpha.0.tgz.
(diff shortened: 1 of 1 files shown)


jmoggr (approved):
Those instructions appear for hash mismatches that are not related to pnpm. `pnpm.fetchDeps` rolls their own hash checking, so it has a non-standard error message. I'll open a follow up PR with this clarification.

kasbah:
Wait, why do we set the `pnpmDeps.hash` to empty? I just ran it without doing that and it tells me the hash that I need to copy in.

```

$ nix build .#frontend
error: hash mismatch in fixed-output derivation '/nix/store/4wpp80j18vvm232ii1ajl2kqnbfgvzq2-frontend-pnpm-deps.drv':
         specified: sha256-vuBwhtNTTRbpgPZS+AQDybASYM9rwWYG8l0bscVQUso=
            got:    sha256-sxczRF8IsYqQzmAAv+IiFeWJygqHCVnSk8fEuy5d1JM=
```

jmoggr:
Yeah... we only need to do that for non `pnpm.fetchDeps` hashes, since `pnpm.fetchDeps` is special (with nicer behavior!). At the time I wrote the docs there were more "standard" hash mismatch issues and I muddled up the issues in the docs. The instructions will still get you the same results.

jmoggr:
I'm working on the fixes, I'll push them here.

jmoggr:
I pushed the changes. The `pnpm` specific docs are improved and we kept the docs for other hash mismatches.

kasbah:
Somehow, later on a different branch, I was not getting the "got" hash when I was doing this again and had to put in empty string. I'm really not sure what I was doing differently.

jmoggr:
I vaguely recall seeing something like that at some point, but dismissed it since there were bigger nixy changes. I'll have claude try to grind out a reproduction.

kasbah:
We could just put the advice back to put in the empty string since that always works even if it's a tiny bit more work.

epatters (approved):
Feel free to merge when you're satisfied.

epatters:
@kasbah, I will go ahead and merge this since it seems an improvement and discussion has stalled out. If there is a lingering problem about needing to put the empty hash, please file an issue about it.

PR #1142, opened 2026-03-17
Fix build for staging falling back to production due to bash

jmoggr:
#1136 implicitly changed how the the shell expansion of `-- --mode staging`. It became a positional argument instead of an option. This is kind of hard to trace because the code does not interact directly.


epatters (approved):
Thanks for the fix! Cursed bash.

PR #1144, opened 2026-03-18
Extract a generic depth-first graph traversal algorithm

### LINKED ISSUE #719: Dedicated depth-first search algorithm for graphs
The current toposort algorithm in CatColab (borrowed from petgraph) inlined depth-first search, whereas it would be more useful to have a general purpose DFS algorithm which is called algorithms such as toposort.

tslil-topos:
## What

`catlog` contains a `toposort` function for graphs which uses mutli-start DFS (twice), this DFS was previously inlined. This change extracts the DFS procedure for `FinGraph` into a hopefully more re-usable method for the library.

Closes #719 

## Notes

* introduces an enum `TraversalDirection` because `Graph` is the category-theoretic graph (we need to know whether we're going into or out of a node), as well as a builder-pattern struct `DFS/DFSBuilder`
* adds some test cases for depth first traversal
* there is potentially a performance/memory difference between the old code and the new code because we now create new `Vec`s for the stack on every invocation of `traverse`. It's difficult for me to reason about this without understanding the compiler optimisation more
* honestly i'm not sure that this refactor brings immediate value, in fact looking around at the algorithms it's quite possible that `toposort` could have been re-written in terms of `spec_order`

### DIFF
--- a/packages/catlog/src/one/graph_algorithms.rs
+++ b/packages/catlog/src/one/graph_algorithms.rs
@@ -1,5 +1,6 @@
 //! Algorithms on graphs.
 
+use derivative::Derivative;
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::hash::Hash;
 
@@ -166,6 +167,7 @@ where
 
 /// An enum which parameterises the traversal direction for the depth first
 /// traversal function.
+#[derive(Debug)]
 pub enum TraversalDirection {
     /// Visit out_neighbors for each vertex.
     Outward,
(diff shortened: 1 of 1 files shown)


packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -164,6 +164,91 @@ where
...
+            if let Some(ref mut callback) = on_discover {
+                callback(nx.clone());
+            }
+            let successors: Vec<_> = match traversal_direction {
+                TraversalDirection::Outward => graph.out_neighbors(&nx).collect(),
+                TraversalDirection::Inward => graph.in_neighbors(&nx).collect(),
+            };
+            for succ in successors {
+                if succ == nx {
+                    return Err(format!("Self loop at node {:#?}", nx));
```
epatters: Not a graph algorithms person myself, but I feel like the more common/expected behavior of a DFS algorithm is that when it encounters a cycle, including as a special case a self-loop, it should detect it and avoid looping infinitely and otherwise proceed without error, instead of returning an `Err` as here.

(I suspect this is just leftover logic from the toposort alg from which it was extracted.)

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -164,6 +164,91 @@ where
...
+/// `discovered` may be persisted across calls to prevent visiting nodes.
+pub fn depth_first_traversal<G: FinGraph, F1, F2>(
+    graph: &G,
+    start_vertex: G::V,
+    discovered: &mut HashSet<G::V>,
+    traversal_direction: TraversalDirection,
+    mut on_complete: Option<F1>,
+    mut on_discover: Option<F2>,
+) -> Result<(), String>
+where
```
epatters, line 211: It's always ungainly in Rust, which has only positional arguments, to have so arguments to a function or method. In this case, what I'd probably do is split up the internal state persisted across invocations and the arguments for a single run, something like:

```rust
struct DFS<V> {
    discovered: HashSet<V>,
    traversal_direction: TraversalDirection
}

impl<V> DFS<V> {
    pub fn traverse(graph, start_vertex, ...)
}
```

If you're not familiar with it, you might also check out the [builder](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) pattern, which is an idiomatic way to work around Rust's lack of keyword arguments.

epatters (changes requested):
Thanks tslil, this is nice! A few comments below.

> honestly i'm not sure that this refactor brings immediate value, in fact looking around at the algorithms it's quite possible that toposort could have been re-written in terms of spec_order

I'm inclined to agree that this is a preemptive refactor, which can run the risk of being premature, but I've got a feeling that one way or another we'll end up writing more graph algorithms that require a DFS. So hopefully it will pay off.

packages/catlog/src/one/graph_algorithms.rs
tslil-topos: thanks, that's a good point. I don't really understand why the original toposort was written the way it was (speed?) so in the latest revision i've removed this erroring here and changed the way cycles are detected. DFS is now agnostic to cycles

packages/catlog/src/one/graph_algorithms.rs
tslil-topos: thanks for pointing this out, this is a little fiddly with the callbacks here (owning them through Box is ergonomically terrible) so i didn't originally go for it. Nevertheless i think that the end result is cleaner, take a look at the latest revision and let me know what you think

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -164,68 +164,222 @@ where
...
+{
+    /// Sets the traversal direction (outward or inward along edges).
+    /// Defaults to [`TraversalDirection::Outward`].
+    pub fn traversal_direction(mut self, dir: TraversalDirection) -> Self {
+        self.traversal_direction = Some(dir);
+        self
+    }
+
+    /// Sets a callback invoked when a vertex is first discovered.
+    pub fn on_discover<D2: FnMut(G::V)>(self, on_discover: D2) -> DFSBuilder<'a, G, D2, C> {
```
tslil-topos: it took me quite some time to figure out how to do builders in the presence of callbacks _without_ ownership of the callbacks, i landed on this---there may be better conventional wisdom out there @epatters ?

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -164,68 +164,222 @@ where
...
+
+    for v in graph.vertices() {
+        if dfs.discovered.contains(&v) {
+            continue;
+        }
+        dfs.traverse(v);
+    }
+    finish_stack.reverse();
+
+    // Instead of multiple DFSs backwards to validate that we have no cycles, we
```
tslil-topos, line 333: pointing out that i changed the algorithm here, not sure why the toposort was written the way it was, but given that we're doing a dfs from every node we can just do the usual cycle detection story here

packages/catlog/src/one/graph_algorithms.rs
```diff
@@ -164,68 +164,222 @@ where
...
+    graph: &'a G,
+    discovered: HashSet<G::V>,
+    traversal_direction: TraversalDirection,
+    on_discover: Option<D>,
+    on_complete: Option<C>,
+}
+
+impl<'a, G: FinGraph, D, C> DFS<'a, G, D, C>
+where
+    G::V: Hash + 'a,
```
tslil-topos: here and elsewhere i made the simplifying assumption that vertices live as long as the graph does

packages/catlog/src/one/graph_algorithms.rs
epatters: Yes, storing references to callbacks in Rust is annoying as far as I've ever been able to tell. This seems reasonable to me.

PR #1145, opened 2026-03-19
Add flag and typing to `DblTheory` to match RFC-0001

### LINKED ISSUE #1099: Allow for non-unital double theories
Following Section 5 of [RFC-0001](https://next.catcolab.org/math/rfc-0001.xml), we should relax the `hom_type` method in the `DblTheory` trait to return an `Option<MorType>` instead of a `MorType`. This should be accompanied by adding a flag to concrete types for double theories signalling whether or not they are unital, i.e. allow loose identities to be constructed.

tslil-topos:
The particular interpretation of that RFC here is as follows:
* constraint: change no code outside of the theory definitions
* constraint: allow users of the DblTheory trait to depend on the type system to know whether hom_types are always present
* implementation trick: "shallow" dependant typing, `fn hom_type(&self, x: Self::ObType) -> <Self::Kind as DblTheoryKind>::Wrap<Self::MorType>;`
  * when `DblTheoryKind` is `Unital`, Wrap = Id, when it's `NonUnital`, Wrap = Option

Notes:
* there was a blanket implementation of DblTheory from VDCWithComposites. This has been macro-ised as `imlp_dbl_theory`, relaxed from blanket to allow more flexible use
* the error messages in hom_op were slightly more informative, distinguishing self.unit_ext(y) from self.through_unit(cell, 0) failures, now these failures are not distiguished and the underlying logic has been moved to VDCWithComposites under the name "unit_arrow"
* docstrings may need more work to adhere to the high standards we have already

Several other design approaches were weighed and dismissed:
* BaseDlbTheory with sub traits DblTheory and NonUnitalDblTheory: introduces more code, more concepts to juggle, lots of duplication
* hom_type returns Option always, with flag on impl: no type system guarantees, adds .unwrap() boilerplate to call sites, forces us to specialise the blanket implementation anyway
* `DblTheory<Unital>` and `DblTheory<NonUnital>` as separate traits: loss of ergonomics at call sites, would be two copies of every method, Rust doesn't pick for you
* UnitalDblTheory supertrait with user contracts: still not strong enough guarantee, have to update call sites

Closes #1099

### DIFF
--- a/packages/catlog/src/dbl/theory.rs
+++ b/packages/catlog/src/dbl/theory.rs
@@ -84,14 +84,14 @@ pub use super::modal::theory::*;
 ///
 /// This trait uses a generic associated type ([`Wrap`](DblTheoryKind::Wrap)) to
 /// control the return type of [`DblTheory::hom_type`] and
-/// [`DblTheory::hom_op`]. For [`Categorical`] theories, `Wrap<T>` is just `T`
-/// (hom types always exist). For [`SetTheoretic`] theories, `Wrap<T>` is
+/// [`DblTheory::hom_op`]. For [`Unital`] theories, `Wrap<T>` is just `T`
+/// (hom types always exist). For [`NonUnital`] theories, `Wrap<T>` is
 /// `Option<T>` (hom types may not exist).
 pub trait DblTheoryKind: fmt::Debug {
     /// Wraps a type to reflect whether values are guaranteed to exist.
     ///
-    /// For [`Categorical`], this is the identity (`T`).
-    /// For [`SetTheoretic`], this is `Option<T>`.
+    /// For [`Unital`], this is the identity (`T`).
+    /// For [`NonUnital`], this is `Option<T>`.
     type Wrap<T>;
 
     /// Converts a wrapped value into an `Option`, for code generic over the kind.
@@ -101,14 +101,14 @@ pub trait DblTheoryKind: fmt::Debug {
     fn pure<T>(value: T) -> Self::Wrap<T>;
 }
 
-/// Categorical double theories guarantee that every object type has a hom type.
+/// Unital double theories guarantee that every object type has a hom type.
 ///
 /// Models of a categorical theory assign *categories* (not just sets) to each
 /// object type: the hom type provides the morphisms.
 #[derive(Debug)]
-pub struct Categorical;
+pub struct Unital;
 
-impl DblTheoryKind for Categorical {
+impl DblTheoryKind for Unital {
     type Wrap<T> = T;
 
     fn into_option<T>(wrapped: T) -> Option<T> {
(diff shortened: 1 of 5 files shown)


packages/catlog/src/dbl/theory.rs
```diff
@@ -67,24 +67,90 @@
...
+    /// Wraps a value that is known to exist.
+    fn pure<T>(value: T) -> Self::Wrap<T>;
+}
+
+/// Categorical double theories guarantee that every object type has a hom type.
+///
+/// Models of a categorical theory assign *categories* (not just sets) to each
+/// object type: the hom type provides the morphisms.
+#[derive(Debug)]
+pub struct Categorical;
```
epatters: To bikeshed the terminology: I don't much like the "set-theoretic" versus "category-theoretic" emphasis because, among other reasons, the categorical models can be valued Set or Cat (depending on whether you use lax or normal lax functors). To be sure, I have no one but myself to blame for this, but nothing better came to mind while writing the RFC.

Since we're now committing to code, we might as well try to do better. @tim-at-topos suggested that we say "non-unital" and "unital" instead of "set-theoretic" and "category-theoretic," which is more technical but also more precise. So I vote we go with that unless someone wants to the make case for different terminology (in which case I'm happy to hear it).

packages/catlog/src/dbl/theory.rs
```diff
@@ -187,84 +256,147 @@ pub trait DblTheory {
...
+                use $crate::dbl::category::{VDblCategory, VDCWithComposites};
+                let y = self.cod(&f);
+                let y_ext = self.unit_ext(y)?;
+                let cell = self.compose_cells($crate::dbl::tree::DblTree(
+                    $crate::one::tree::OpenTree::linear(vec![
+                        $crate::dbl::tree::DblNode::Spine(f),
+                        $crate::dbl::tree::DblNode::Cell(y_ext),
+                    ]).unwrap(),
+                ));
+                self.through_unit(cell, 0)
```
epatters: I can't immediately recall why this code was here instead of in `VDCWithComposites`, but it'd be nice to move it outside the macro to keep the macro code as simple as possible.

packages/catlog/src/dbl/theory.rs
tslil-topos: thanks, i've made these changes

packages/catlog/src/dbl/theory.rs
tslil-topos: I've moved this general method to VDCWithComposites under the new name "unit_arrow" (happy to rename) where it does seem to be most logical, and consequently shrunk the macro code.

tslil-topos:
thanks for the review Evan, i'm glad you like this approach. Apologies for the force push there, so used to working with jj locally that i forgot that github still can't accurately track things across histories.

PR #1146, opened 2026-03-19
Replace biome with oxlint and oxfmt

kasbah:
Oxlint seems to be making a bit more progress in the realm of "fast linters for TS/JS" than Biome so I thought I'd give it a try. It has more advanced linting of SolidJS specifically by being able to use the Solid ESLint plugin. 

On the whole I tried to capture what we are already doing as as much as possible by turning off rules. It does seem to catch some very mild reactivity concerns and prompted me to clean up some minor things. I feel like it would catch some reactivity blunders I've made in the past.  There are definitely some false positives so we have to decide if hitting/working around them is worth it.

I've left Biome in place for now, there don't seem to be any conflicts, but I'm sure we could hit them so maybe we would want to turn Biome linting off at some point and just use it for formatting.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,8 +126,15 @@ jobs:
         run: |
           pnpm install
 
+      - name: Install Rust toolchain from file
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          # Don't override flags in cargo config files.
+          rustflags: ""
+
       - name: Format/linting/import sorting
         run: |
+          pnpm --filter "./packages/*" run build:deps
           pnpm --filter "./packages/*" run ci
 
   build_nixos_system:
@@ -163,11 +170,6 @@ jobs:
       - name: Setup pnpm
         uses: pnpm/action-setup@v4
 
-      - name: Setup wasm-pack
-        uses: jetli/wasm-pack-action@v0.4.0
-        with:
-          version: 'v0.13.1'
-
       - name: Run dev setup
         run: |
             docker run --name catcolab-postgres -e POSTGRES_USER=catcolab -e POSTGRES_PASSWORD=password -e POSTGRES_DB=catcolab -p 5432:5432 -d postgres:15
@@ -199,4 +201,3 @@ jobs:
     - name: Run frontend tests with VM
       run: |
         nix build .#checks.x86_64-linux.frontendTests -L --no-sandbox
-


kasbah:
I was a bit concerned with the churn of changing formatter so I tried to minimize it but there is still some especially around import sorting and .mdx files. 

I also checked what biome things we had enabled/used by default and tried to find equivalents. Fully switched over now.

packages/frontend/src/components/json_import.tsx
```diff
@@ -16,7 +16,7 @@ interface JsonImportProps {
  * File size is currently limited to 5MB.
  *
  */
-export const JsonImport = ({ onImport, validate }: JsonImportProps) => {
+export const JsonImport = (props: JsonImportProps) => {
```
epatters, line 19: Nice that the linter caught this classic way to break reactivity!

packages/frontend/src/stdlib/analyses/mass_action_config_form.tsx
```diff
@@ -19,7 +19,7 @@ export function MassActionConfigForm(props: {
             : undefined;
 
     return (
-        <FormGroup compact style="min-width: 286px">
+        <FormGroup compact style={{ "min-width": "286px" }}>
```
epatters, line 22: Can we configure the linter to flag inline CSS styling? :) IMO, we should ban that in our codebase.

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -31,7 +31,14 @@ const tooltip = () => (
...
     return (
-        <div style="display: flex; align-items: center; justify-content: flex-end; gap: 4px;">
+        <div
+            style={{
+                display: "flex",
+                "align-items": "center",
+                "justify-content": "flex-end",
+                gap: "4px",
+            }}
+        >
```
epatters, line 41: This inline CSS shouldn't be here either! (I know it's not your fault, but I can't resist flagging it.)

packages/frontend/src/stdlib/analyses/tabular_view.tsx
```diff
@@ -11,43 +12,45 @@ import "./tabular_view.css";
 */
 function ACSetTable(props: { model: DblModel; rawdata: Record<string, string[]>; obId: string }) {
     // The primary key of this table is given by `rawdata[obname]`
-    const rows: Array<string> = props.rawdata[props.obId] || [];
-    const obname = props.model.obGeneratorLabel(props.obId)?.join(".") || "";
+    const rows = () => props.rawdata[props.obId] || [];
+    const obname = () => props.model.obGeneratorLabel(props.obId)?.join(".") || "";
```
epatters, line 16: Nice, lots more reactivity problems caught here.

packages/frontend/package.json
```diff
@@ -74,11 +74,11 @@
         "uuid": "^13.0.0"
     },
     "devDependencies": {
-        "@biomejs/biome": "^2.3.8",
         "@mdx-js/mdx": "^2.3.0",
         "@types/html-escaper": "^3.0.4",
         "@types/mdx": "^2.0.13",
         "@types/node": "^24.0.0",
+        "eslint-plugin-solid": "^0.14.5",
```
epatters, line 81: Just checking here: it is intentional to add `eslint-plugin-solid` in addition to `oxlint` at the root level, because `oxlint` doesn't provide this functionality?

packages/frontend/src/stdlib/analyses/mass_action_config_form.tsx
kasbah: We probably can, but I'm not sure I agree, since CSS is often so fiddly, without inline styles you tend to get classes like `containerWrapperWrapper` with just one attribute like `padding-left: 1px` which doesn't seem like an improvement.

packages/frontend/src/stdlib/analyses/sql.tsx
kasbah: That one I agree should not be inlined.

packages/frontend/package.json
kasbah: That's right. The view is we might have non-solid packages at some point that don't need the plugin.

packages/frontend/src/stdlib/analyses/mass_action_config_form.tsx
kasbah: I just checked and we would need a custom plugin because the only existing one is for React and doesn't work with Solid.

PR #1149, opened 2026-03-22
Retrospective RFC on DoubleTT

epatters:
I've converted the docs on DoubleTT, originating in

- #922

into an RFC. As explained in the new, expanded document, that's a bit unusual, but I think it makes sense because I'm hoping to evolve DoubleTT in a new direction that will make some of this material obsolete. So, like any RFC, this document represents a snapshot of a design at a point in time.

Apologies @tim-at-topos as this will conflict with #1122 and will need to be converted to Quarto. However, the source material was already in forester, so most of that work would have to be done at some point anyway.

### DIFF
--- a/math-docs/trees/refs/dreyer-2005.tree
+++ b/math-docs/trees/refs/dreyer-2005.tree
@@ -1,0 +1,5 @@
+\title{Understanding and evolving the ML module system}
+\author{Derek Dreyer}
+\date{2005}
+\taxon{reference}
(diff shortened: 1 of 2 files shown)


tim-at-topos:
I feel conflicted about RFC-0002 having a date prior to RFC-0001... but also that's the kind of thing that will make me chuckle in the future so oh well!

tim-at-topos:
@epatters it seems like there's a hanging **TODO** in https://next.catcolab.org/math/unstable-266.xml

epatters:
Oh, dang. That's been there since January.

I'll do something about that once the RFC has been converted to the new format.

PR #1150, opened 2026-03-23
Notify client when server version changes mid-session

jmoggr:
Notify long running clients when the backend changes.

The frontend is checking if the git hash returned in an RPC header is different from the last hash it saw. This sidesteps the issue where the frontend and backend are often deployed independently, with one changing but not the other.

This does not handle the case where:
- there is a new version of the frontend available
- the browser loads a stale frontend from cache (this is unlikely, since netlify handles cache busting for us).

However I don't think it's possible to handle those cases while we are deploying the frontend and backend independently. While this isn't perfect, I think it's fairly minimal and would have at least saved myself a couple for :facepalm: moments.

### DIFF
--- a/packages/frontend/src/App.tsx
+++ b/packages/frontend/src/App.tsx
@@ -28,7 +28,13 @@ const Root = (props: RouteSectionProps) => {
     invariant(repoUrl, "Must set environment variable VITE_AUTOMERGE_REPO_URL");
 
     const firebaseApp = initializeApp(firebaseOptions);
-    const api = new Api({ serverUrl, repoUrl, firebaseApp });
+    const [isVersionMismatch, setVersionMismatch] = createSignal(false);
+    const api = new Api({
+        serverUrl,
+        repoUrl,
+        firebaseApp,
+        onVersionChange: () => setVersionMismatch(true),
+    });
 
     const [isSessionInvalid] = createResource(
         // oxlint-disable-next-line solid/reactivity -- createResource fetcher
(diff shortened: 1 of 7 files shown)


epatters (commented):
Thanks Jason. Seems reasonable to me but I'm wondering if this could be handled with less specialized machinery, while also "pushing" the update rather than waiting to "pull" it via RPC, by using the new Automerge meta-document. What do you think, @kasbah?

jmoggr:
> by using the new Automerge meta-document

I think that is a good idea, and is strictly better than this. If @kasbah then I'll create an issue and close this PR.

kasbah:
I agree, could you try to implement this on top of `kb/automerge-user-state` and see how well you think it works for this sort of thing @jmoggr ?

packages/frontend/src/App.tsx
```diff
@@ -100,6 +109,25 @@ export function SessionExpiredModal() {
...
+                    </p>
+                    <Button variant="positive" onClick={() => location.reload()}>
+                        Reload Page
+                    </Button>
+                </Content>
+            </Portal>
+        </Dialog>
+    );
+}
+
```
kasbah, line 130: If the modal is going to block whatever the user is doing we might as well just force reload the page?

jmoggr: Bad UX imo. The user does not expect the page to "randomly reload" and we'll have people compail/report it, better to communicate. We have a similar pattern when the client detects a stale session.

>  could you try to implement this on top of kb/automerge-user-state

will do!

kasbah: True, I guess my complaint was that we should have a less intrusive modal component (toast) as an option but it's fair to say we shouldn't pile that on to this work.

epatters:
In light of the discussion above, I'm going to close this one out. Now that #976 is merged, it should be easier to build on if/when you're ready to try that.

PR #1151, opened 2026-03-23
Bypass the DB for user state updates

kasbah:
Note this is an improvement for #976 that I've put here for ease of review. 

Instead of  PG notify this handles updates by patching the automerge doc directly. ~~It performs noticably better when e.g. deleting a document. Tested with about 3k `DocInfo` in user state.~~ I can't seem to replicate that performance improvement I thought I saw.

Note that most of the changes are in `user_state_updates.rs` which GH diff helpfully hides for you.


packages/backend/src/user_state.rs
```diff
@@ -463,180 +468,93 @@ pub async fn read_user_state_from_db(user_id: String, db: &PgPool) -> Result<Use
...
     user_id: &str,
     user_state: &UserState,
 ) -> Result<DocumentId, AppError> {
-    debug!(
-        user_id = %user_id,
-        document_count = user_state.documents.len(),
-        "Creating new user state document"
-    );
+    let persisted_doc_id: Option<(String,)> =
+        sqlx::query_as("SELECT state_doc_id FROM users WHERE id = $1 AND state_doc_id IS NOT NULL")
```
jmoggr, line 519: We could write a migration to generate automerge docs for each existing user, which would allow the column to become `NOT NULL`. The migrator was written with exactly this sort of case in mind: being able to run arbitrary rust/automerge code during migrations.

PR #1155, opened 2026-03-23
Generate Automerge docs in migration to enforce NOT NULL on `state_doc_id`

jmoggr:
I think this is worth doing? Nulls are bad and everything should be more like Rust, but this didn't simplify things as much as I hoped it would. The current user creation process is kinda weird.


kasbah:
I'm unsure about this because it adds some complexity for not much value. How is setting this to an empty document better than setting this to the empty string? How is that better than setting it to null? 

The user state docs are designed to  be ephemeral: we persist them but if we don't have the doc for some reason we just create a new one on demand. We also always re-populate it when the server restarts no matter what. Being explicit about "we actually haven't created and persisted this doc id" seems clearer to me.

jmoggr:
Closing. While I dislike nulls, I agree with Kaspar that the complexity isn't adding much.

PR #1158, opened 2026-03-24
Enable LaTeX in CI action to build RFCs

tim-at-topos:
https://github.com/ToposInstitute/CatColab/pull/1122 didn't remember that the CI doesn't have LaTeX installed, so this (hopefully) fixes that

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -268,7 +268,14 @@ jobs:
         uses: quarto-dev/quarto-actions/setup@v2
         with:
             version: 1.8.26
-            tinytex: true
+
+      - name: Setup TinyTeX
+        uses: r-lib/actions/setup-tinytex@v2
+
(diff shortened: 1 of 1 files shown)


tim-at-topos:
I *think* that Quarto uses the R install of TinyTeX, which automatically installs missing packages as it tries to build? But I have no way of confirming this... so maybe I'll just add all the obvious dependencies just in case 🤷

epatters:
See my comments on Zulip about the redirects. I'd really like for us to not continue in this blind way, guessing at what might work.

epatters:
If you look at the build log for the RFC action, you can see that it's failing to build the LaTeX snippets:

> luaotfload | db : Font names database not found, generating new one.
> luaotfload | db : This can take several minutes; please be patient.))
> 
> ! LaTeX Error: File `luatex85.sty' not found.

Issue #1098, opened 2026-03-05
RFC-0001: tracking issue


tim-at-topos:
This is the tracking issue for freezing [RFC-0001](https://next.catcolab.org/rfc/0001).

tim-at-topos:
Having spoken with @epatters and @tslil-topos , here is my current sketch plan:

#### A. Dev facing

1. Allow for non-unital double theories
   https://github.com/ToposInstitute/CatColab/pull/1145
   https://github.com/ToposInstitute/CatColab/pull/1161
2. Implement the non-unital modal double theory of Polynomial ODE systems (4.1 in the [RFC](https://next.catcolab.org/rfc/0001)), i.e. non-unital symmetric multicategories
    https://github.com/ToposInstitute/CatColab/pull/1162
3. Refactor the corresponding analyses to actually build models in this theory before passing them into the ODE solver
    https://github.com/ToposInstitute/CatColab/issues/1177

#### B. User facing

1. Make a *logic* for Polynomial ODE systems
  https://github.com/ToposInstitute/CatColab/pull/1176
2. Configure some notebook cell editors to allow users to build models in this theory in a nice-looking way
  https://github.com/ToposInstitute/CatColab/pull/1211
3. Add the two obvious analyses: (i) show the equations, and (ii) solve the equations.
  https://github.com/ToposInstitute/CatColab/pull/1176/
4. Add another logic for *signed* polynomial ODE systems
  https://github.com/ToposInstitute/CatColab/pull/1194

**Once these are done, I would consider this RFC addressed.**

---

The obvious next step would be to think about the "Future possibilities" section, which would probably look like the following:

#### C. Future possibilities

1. Create a "Custom ODE dynamics" analysis that allows users to apply their own ODE models to their own e.g. CLD models
2. Figure out how much this implementation differs from the mathematical story sketched out in the RFC and what we should do about that

PR #1161, opened 2026-03-25
Make modal theories generic over kind of double theory

tslil-topos:
> Warning, this is potentially suboptimal and i'd like to solicit opinions on this approach.

Part of #1098 

## Background
This whole story started because Rust doesn't allow us to impl traits on _instances_, only on concrete types, so the decision point was choosing to make ModalDblTheory generic over DblTheoryKind, from which the rest of this follows.

## Implementation details
ModalDblTheory is now generic over Kind: DblTheoryKind, mirroring the existing Unital/NonUnital split in DblTheory. The existing VDCWithComposites impl is also made generic over Kind. Two concrete DblTheory impls are still generated via impl_dbl_theory! for `ModalDblTheory<Unital>` and `ModalDblTheory<NonUnital>` separately, rather than as a single blanket impl, because the two cases differ in the return type of hom_type/hom_op (MorType vs `Option<MorType>`).

This means that when writing `impl<Kind: DblTheoryKind>` blocks that need DblTheory methods, Rust cannot infer the associated types ObType, MorType, ObOp, MorOp from the bound alone - it only knows the bound holds for each concrete Kind, not generically. Consequently, all impl blocks on `ModalDblModel<Kind>` that depend on DblTheory carry an explicit where clause:

    where ModalDblTheory<Kind>: DblTheory<
        ObType = ModalObType, MorType = ModalMorType,
        ObOp = ModalObOp, MorOp = ModalMorOp,
    >

This covers Category, FgCategory, DblModel, FpDblModel, MutDblModel, Validate, PrintableDblModel, Display, and the inherent impl containing infer_ob_type/infer_mor_type/ob_has_type.

Everything connected to the frontend (tt, modelgen, analyses, wasm bindings) and to ModalDblModel instantiation is Unital only so this is not a functionality change. 

`TheoryDef<Kind>` is made generic so that it can hold a `ModalDblTheory<Kind>`,  but its impl block (constructors and all DblTheory-dispatching methods) is restricted to `TheoryDef<Unital>`. All existing call sites in stdlib, tt, and  catlog-wasm are annotated with `<Unital>` explicitly.

## Points for review

Besides the entire approach taken, i have not carefully indexed whether it makes sense to have certain things which are currently generic over `DblTheoryKind` only exist for `Unital`, for example. One might imagine that things whose flavour is such that they were designed with composition-ish use-cases should not be extended to work over NonUnital (even if that would make sense). This question would require more understanding of the codebase than i presently have.

The macro  `impl_dlb_theory!` can and should be improved to inspect the kind automatically to prevent user-error for non-sensical implementations, although perhaps that can be a future PR.

## Planned work on-top of this

Implement "actually" non-unital theories, extend models to allow non-unital variants, and figure out how to communicate this with the analyses.

### DIFF
--- a/packages/catlog/src/dbl/modal/model.rs
+++ b/packages/catlog/src/dbl/modal/model.rs
@@ -74,10 +74,7 @@ impl MorListData {
 
 /// A model of a modal double theory.
 #[derive(Clone)]
-pub struct ModalDblModel<Kind>
-where
-    Kind: DblTheoryKind,
-{
+pub struct ModalDblModel<Kind> {
     theory: Rc<ModalDblTheory<Kind>>,
     ob_generators: HashFinSet<QualifiedName>,
     mor_generators: ComputadTop<ModalOb, QualifiedName>,
@@ -107,9 +104,7 @@ impl<Kind: DblTheoryKind> ModalDblModel<Kind> {
 
 #[derive(RefCast)]
 #[repr(transparent)]
-struct ModalDblModelObs<Kind>(ModalDblModel<Kind>)
-where
-    Kind: DblTheoryKind;
+struct ModalDblModelObs<Kind>(ModalDblModel<Kind>);
 
 impl<Kind: DblTheoryKind> Set for ModalDblModelObs<Kind> {
     type Elem = ModalOb;
--- a/packages/catlog/src/dbl/modal/theory.rs
+++ b/packages/catlog/src/dbl/modal/theory.rs
@@ -258,10 +258,7 @@ pub type ModalMorOp = DblTree<ModalObOp, ModalMorType, ModalNode>;
 /// A modal double theory.
 #[derive(Debug, Derivative)]
 #[derivative(Default(new = "true"))]
-pub struct ModalDblTheory<Kind>
-where
-    Kind: DblTheoryKind,
-{
+pub struct ModalDblTheory<Kind> {
     _kind: PhantomData<Kind>,
     ob_generators: HashFinSet<QualifiedName>,
     arr_generators: ComputadTop<ModalObType, QualifiedName>,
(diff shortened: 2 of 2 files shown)


tslil-topos:
Thanks for your thoughts, comments, and time here @epatters . I broadly agree with what you have said, there is far too much boilerplate in the approach for the original PR.

I spent quite some time bashing against the limitations of the Rust type checker/system, and i thought i'd give this one more go. The root cause for all of the noise around using these traits was _not_ their existence as traits or the dependent typing that they implied, but rather the singular quirk that i implemented two separate (exhaustive) versions of DblTheory  for ModalDblTheory so the compiler could not decide what to do (although again, these are exhaustive).

There's something of an escape hatch for us in this particular instance, where because we're doing runtime validation anyway, we can move that work into the Unital/NonUnital trait implementations themselves and then essentially do type directed dispatch. This allows us to give a uniform implementation for ModalDblTheory as DblTheory of the matching kind, regardless of what that kind is. The only remaining boilerplate after that for the rest of the codebase is propagating this kind information --- unfortunately because we have some newtypes floating about we land up with `ModalDblModelObs<Kind>` for example.

However i agree with you that in principle this could be a mismatch of expectations. If we wanted `Kind` to be "closed" then yes, we should do something else entirely (flags, trait hierarchy, whatever else we're interested in).  On the other hand if we want this available as typing information then we're probably looking at some kind of variation on what i've tried here, or perhaps more liberally (from the type system perspective), the const generic things you suggested above.

Note that there's still some cleanup to be done on the current state of the PR, the macro in particular is still a little complicated. I also learnt about PhantomData which might clean up `_kind` in ModalDblTheory.

So to me this whole PR and further development reduces to a question of what we want: "open" vs "closed", is it important that everything know about unitality when handling (derived) objects?

tslil-topos:
Thanks Evan, i think that the code as it is now is about as neat as the current approach will allow (this took me on a detour of learning about the many ways in which Rust does not enforce trait bounds, in particular in `type Foo = Bar<...>` and `struct Foo<...>` locations). 

The changes since you last looked at it amounted to:
* minor cleanup of the macro, it can't really be reduced too much further because it has to serve a dual use-case for now (Tabulators + Discrete are always only Unital, but Modal has Kind)
* follow the "sealed trait" pattern to prevent future shenanigans on DblTheoryKind
* remove extraneous trait bounds that weren't doing anything
* flip to PhantomData in ModalDblTheory 

Let me know what you think

PR #1162, opened 2026-03-25
Theory of polynomial ODE systems and toy Lotka-Volterra style model

tslil-topos:
The background work here is standard, fiddle with trait bounds to make NonUnital things work, make sure we track Default.

The confusing part is entirely contained in the way that the example Lotka-Volterra model is built, it would be great to have a breakdown of _why_ what is written there "works" (or is wrong)

Note that this is stacked on #1161 and relates to #1098


tslil-topos:
I have rebased this on-top of the most recent changes in #1161

packages/catlog/src/stdlib/theories.rs
```diff
@@ -204,9 +204,14 @@ pub fn th_multicategory() -> ModalDblTheory<Unital> {
     th_generalized_multicategory(List::Plain)
 }
 
+/// The theory of a non-unital symmetric multicategory.
+pub fn th_sym_multicategory() -> ModalDblTheory<NonUnital> {
+    th_generalized_multicategory(List::Symmetric)
+}
```
epatters, line 210: Naming things is hard, but I feel about confident about this one: "symmetric multicategory" has a standard meaning and this function should return the unital double theory for that categorical structure.

packages/catlog/src/stdlib/theories.rs
```diff
@@ -204,9 +204,14 @@ pub fn th_multicategory() -> ModalDblTheory<Unital> {
...
+
 /// The theory of a generalized multicategory over a list monad.
-fn th_generalized_multicategory(list: List) -> ModalDblTheory<Unital> {
-    let mut th = ModalDblTheory::new();
+fn th_generalized_multicategory<Kind: DblTheoryKind>(list: List) -> ModalDblTheory<Kind> {
```
epatters: Ditto: following Cruttwell-Shulman, "generalized multicategory" has a fairly standard meaning, and I don't think we should make this function parametric over the theory kind.

I suggest defining a new function, for the reason above and also because (2) we can take the opportunity to use meaningful names and (3) `th_generalized_multicategories` is missing axioms that we will later add---axioms that don't make sense for this non-unital theory.

So, following the RFC, I suggest:

```rust
fn th_polynomial_ode_system() -> ModalDblTheory<NonUnital> {
    let mut th = ModalDblTheory::new();
    th.add_ob_type(name("State"));
    let x = ModeApp::new(name("State"));
    th.add_mor_type(name("Contrib"), x.clone().apply(Modality::List(List::Plain)), x);
    th
}
```

packages/catlog/src/stdlib/models.rs
```diff
@@ -167,6 +167,74 @@ pub fn sir_petri(th: Rc<ModalDblTheory<Unital>>) -> ModalDblModel<Unital> {
     model
 }
 
+/// An example of Lotka–Volterra dynamics viewed as a non-unital theory for a symmetric multicategory.
+pub fn lotka_volterra_dynamics(th: Rc<ModalDblTheory<NonUnital>>) -> ModalDblModel<NonUnital> {
```
epatters, line 171: With the caveat that DoubleTT will need to tweaks to emit non-unital models, you can replace this unpleasant code with a call to ``tt:modelgen::Model::from_text``, with the input string being something like:

```text
[
  A : State,
  B : State,
  C : State,
  A_growth : Contrib[[A], A],
  ...
  AB_interaction : Contrib[[A, B], A],
]
```

packages/catlog/src/stdlib/theories.rs
tslil-topos: agreed :100:

packages/catlog/src/stdlib/theories.rs
tslil-topos: thanks, agreed

packages/catlog/src/stdlib/theories.rs
```diff
@@ -204,6 +204,20 @@ pub fn th_multicategory() -> ModalDblTheory<Unital> {
...
+pub fn th_sym_multicategory() -> ModalDblTheory<Unital> {
+    th_generalized_multicategory(List::Symmetric)
+}
+
+/// The theory of simple polynomial ODE systems.
+pub fn th_polynomial_ode_system() -> ModalDblTheory<NonUnital> {
+    let mut th = ModalDblTheory::new();
+    th.add_ob_type(name("State"));
+    let x = ModeApp::new(name("State"));
+    th.add_mor_type(name("Contrib"), x.clone().apply(Modality::List(List::Symmetric)), x);
```
tslil-topos, line 217: note that this is `Symmetric`, the order does not matter

packages/catlog/src/stdlib/models.rs
tslil-topos: i might leave allowing DoubleTT to make non-unital theories as a later change on main after this merges, if that's alright? I think that this change set out to do what it needed to, and for other things like #1167 it makes sense to get this in.

epatters: Sounds good, I'm happy to wait on that. Do you mind filing an issue so that we don't forget?

packages/notebook-types/bindings/user_state.ts
epatters, line 1: Somehow this file snuck in here. I'm surprised: is this not git-ignored?

packages/catlog/src/stdlib/models.rs
```diff
@@ -167,6 +167,74 @@ pub fn sir_petri(th: Rc<ModalDblTheory<Unital>>) -> ModalDblModel<Unital> {
...
+    model.add_ob(c.clone(), ob_type.clone());
+    // The growth terms, corresponding to
+    // dA/dt += g_A A
+    // dB/dt += g_B B
+    // dC/dt += g_C C
+    model.add_mor(
+        name("A_growth"),
+        a.clone().into(),
+        a.clone().into(),
+        ModalMorType::Zero(ob_type.clone()),
```
epatters: I missed this on my first review: this isn't right because the theory is non-unital. There is only one morphism in the theory, `Contrib`. So this should be:

```rust
model.add_mor(
        name("A_growth"),
        ModalOb::List(List::Symmetric, vec![a.clone().into()]),
        a.clone().into(),
        mor_type.clone(),
    );
```

and similarly below.

packages/notebook-types/bindings/user_state.ts
tim-at-topos: @kasbah do you know what this file is? it's been appearing for me recently too

packages/catlog/src/stdlib/models.rs
tslil-topos: @tim-at-topos this is how it was initially before your https://github.com/ToposInstitute/CatColab/pull/1162/commits/b6abb443b7508fc343bbc31986083bf4ef21a4b2 ?

tim-at-topos: no, I had never noticed this mistake, back in our initial version of this I think!

packages/catlog/src/stdlib/models.rs
tslil-topos: https://github.com/ToposInstitute/CatColab/issues/1181

thanks!

packages/catlog/src/stdlib/models.rs
epatters: What this points to is that the validation code should be updated (perhaps you're already working on this in #1178, I haven't checked) so that the model above fails to validate in the non-unital case.

packages/notebook-types/bindings/user_state.ts
kasbah: Who knows who could be to blame? (Don't run `git blame`. I blame my tools: it's `ts-rs`'s weird default behaviour!) https://github.com/ToposInstitute/CatColab/pull/1182

packages/catlog/src/stdlib/models.rs
tslil-topos: indeed, this should have failed to validate. For now i'll make an issue to track this #1185

packages/notebook-types/bindings/user_state.ts
tslil-topos: i have removed the file from this change, thanks

PR #1167, opened 2026-03-27
feature: Maybe modality, theory for schemas with nullable types

tslil-topos:
## Mathematics 

It looks like the current stratification is:
* theories are virtual double categories with adjectives
* models of theories are functors into Set (perhaps themselves with adjectives)

However on the surface of things this appears insufficient for the task that this change sets out to accomplish (and perhaps more broadly at odds with the C-set formulation of databases)

Instead this change takes the point of view that:
* theories are virtual double categories with adjectives
* models are fibred virtual double categories over theories with adjectives
* _instances_ are functors from models into Set with adjectives

As an illustrative example, this change builds the following:
* the "Maybe" modality, classical
* a modal, non-unital double theory given by two objects (Entity, AttrType) and four pro-arrows (Attr, AttrMaybe, Rel, RelMaybe)
* models of this double theory are modally fibred virtual double categories, so in particular there are possibly discrete categories of objects e:Entity, ty:AttrTy, and concrete pro-arrow collections over the four pro-arrows of the theory. Because we're _modally_ fibred, if e:Entity then Maybe(e):Maybe(Entity), etc
* an _instance_ of this model is now a C-set essentially, it turns e:Entity into a set, ty:AttrTy into a set, r:Rel : e:Entity -|-> e':Entity into a collection of functions on the entity sets, etc. It does this in a way that is compatible with the modality.

The central claim is that this is what we probably wanted from database modelling: each ty:AttrTy becomes an actual set (or more) of terms of that type etc. The cost of this is essentially in upgrading from considering only functors F: T -> Set to spans T <- M -> Set (of which the former are surely a nice subcategory in the usual ways)

This appears to be a necessary extension of the formalism, which talks about functors Theory -> Set only. In that setting it does not appear to be possible to name a concrete pro-arrow with codomain Mabye(ty) when ty was an _element_ of the set AttrTy, we would be forced to pick Just(ty) or None. So to make sure that we had a theory whose every model was a semantically correct "database schema with nullable relations and attributes" we would be force to encode the mechanics of "Maybe" into our theory (and that would not strictly be possible because we would exhaust dimensions, so could not give laws of the "monad" thus encoded).

## Code
* `Modality::Maybe` and supporting plumbing for modal theory level things
* Refactor of `ModalOp::Concat` to use named arguments instead of a 3-tuple
* Model level code (including following the example of List to the letter, which implements flatten_list separately for ModalOb and ModalMor)
* `th_schema_maybe` which is the non-unital (see section) theory on two objects (Entity, AttrType) and four pro-arrows (Attr, AttrMaybe, Rel, RelMaybe)
* WASM ThSchemaMaybe, corresponding changes to allow non-unital boxed models/theories
* TypeScript registering of the `simple-schema-maybe` file and its related machinery, including widgets for nullables (dashed arrows)
* Generalising the ERD analysis to allow for nullable things, specifying what the relations should be called (instead of baking in assumptions about homs), adding odotcrowsfoot
* Rework the SQL analysis to be based on an intermediate representation, and support two paths to generate that: the old discrete category code (essentially unchanged), new (possibly too) generic non-unital modal double theory code which understands Maybe

## To do

* [X] Figure out how to plumb this into the TypeScript, maybe @kasbah and i can collaborate on that. (hacky because of tight coupling in FE)
* [X] allow pro-arrow to have maybe in codomain
* [X] Rework SQL analysis to support Modal stuff 

## Pictures

image

## Note on unitality
Non-unital was chosen because it wasn't obvious to me why `th_schema` was chosen to be "unital" (`d1c825b6` from @olynch maybe but that could have just been a file move?), as i understand it there's a horizontal FPCategory which means in particular that the semantics of what would have otherwise been a DB now has "Entity" and "AttrType" as _categories_ instead of mere sets.

This change is therefore based on #1162 and #1178


tslil-topos:
See the comment on #1201

PR #1168, opened 2026-03-27
Reduce duplication in analyses bindings and extract LaTeX helpers

### LINKED ISSUE #1050: Remove duplication in `mass_action_tab` and `mass_action_modal`
Following #1045 , in `catlog/src/theories.rs`, the functions `mass_action_tab` and `mass_action_modal` are essentially identical apart from the first two lines. This should be refactored in a sensible way to avoid so much reduplication.

tim-at-topos:
Closes #1050. Future work is to move most of the contents of (the new file) `catlog-wasm/src/latex.rs` into `catlog`, where I think it belongs, by some trait implementation.

### DIFF
--- a/packages/catlog/src/stdlib/analyses/ode/mass_action.rs
+++ b/packages/catlog/src/stdlib/analyses/ode/mass_action.rs
@@ -16,7 +16,7 @@ use serde::{Deserialize, Serialize};
 #[cfg(feature = "serde-wasm")]
 use tsify::Tsify;
 
-use super::ODEAnalysis;
+use super::{ODEAnalysis, Parameter};
 use crate::dbl::{
     model::{DiscreteTabModel, FpDblModel, ModalDblModel, TabEdge},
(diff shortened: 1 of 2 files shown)


packages/catlog/src/stdlib/analyses/ode/mass_action.rs
```diff
@@ -187,7 +187,16 @@ pub struct MassActionProblemData {
...
+pub type Parameter<Id> = Polynomial<Id, f32, i8>;
+
+/// The mass-action analysis is currently implemented for Petri nets and stock-flow
+/// diagrams, and we can avoid some code reduplication by making this explicit.
+pub enum MassActionAnalysisLogic {
+    /// The modal theory of Petri nets.
+    PetriNet,
+    /// The discrete tabulator theory of stock-flow diagrams.
+    StockFlow,
+}
```
epatters: Since this type is only used in `catlog-wasm` to control dispatch and doesn't seem to have utility beyond that, I'd just move it to `catlog-wasm` near where it is actually used.

Issue #161, opened 2024-09-10
Principled approach to keyboard focus management


epatters:
As noted in #159, the app does not handle keyboard focus appropriately when multiple notebooks are open. To make this at all reliable, we need a more principle approach to managing focus.

@olynch suggested something like that having top-level application state that tracks which component currently "owns" the focus.

PR #1175, opened 2026-03-31
History and rollback for notebooks

kasbah:
This adds an undo/redo side-bar ~and keyboard shortcuts (Ctrl-Z and Ctrl-Shift-Z)~. 

An undo-able change-set is a set of  changes made within 500ms of the last change: if you don't change anything for 500ms then the snapshot gets saved.

image

[ui-component demo](https://branch-kb-undo-redo--catcolab.netlify.app/dev/ui-components/?path=/docs/misc-historynavigator--docs)

Some things to consider:

- The DB migration takes a long time (3 minutes on the staging DB dump on my laptop)
- The undo history is *global* per-document meaning you will undo someone else's changes if you are working on the same doc. 
- The undo history is stored as a tree but we present a flat history of one branch to the user.
- Due to how we handle the new cell popups we do have a strange interaction where undoing/re-doing to an empty cell creation will show the menu unexpectedly.  We will need to fix this in the future, we should probably also not create the cell empty before the user picks what it's going to be.
- How this relates to the Automerge history is a little un-intuitive due to not being able to set an existing Automerge doc to particular "heads". I drew a picture:

[automerge_undo_redo](https://excalidraw.com/#json=A5xHD9sefTRqbfvbRJ9e-,luKBvx0wa06WoOhYxpU97Q)


epatters:
> The undo history is _global_ per-document meaning you will undo someone else's changes if you are working on the same doc.

That, along with the user interface, makes me think we should describe this feature as "history" a la Google docs instead of undo-redo. What do you think?

kasbah:
It is called that in the code already: "history" and "snapshots". Nothing is really calling this undo/redo except me and the undo & redo buttons on their tooltips (which are that and should definitely keep saying that).

To me the most important aspect of this feature is actually that I can Ctrl-Z when I mess up so that's why I describe it as undo/redo. I don't mind "history" or "undo history", I think they all describe what this does.

packages/backend/src/document.rs
```diff
@@ -140,65 +144,113 @@ pub async fn ref_deleted_at(
...
-    Ok(())
+        if let Err(e) = update_ref_for_users(&state, ref_id, vec![]).await {
+            tracing::error!(%ref_id, error = %e, "Failed to update user states after set_current_snapshot");
+        }
+
+        Ok(())
+    }
+    .await;
+
+    state.suppress_autosave.write().await.remove(&ref_id);
```
jmoggr: I'm fairly certain there is a race condition here: there is no guarantee that `changes.next().await` from `ensure_autosave_listener` runs before we remove the flag. If `changes.next().await` does not run first, then a snapshot will be recorded even though we meant to skip it.

Given that we are doing DB operations between modifying the doc and removing the flag, it seems highly likely that `changes.next().await` will always runs before this line, however the lack of guarantees means that this will probably break at some point.

packages/backend/src/autosave.rs
```diff
@@ -0,0 +1,57 @@
...
+                    }
+                    continue;
+                }
+                if let Some(handle) = snapshot_handle.take() {
+                    handle.abort();
+                }
+                snapshot_handle = Some(tokio::spawn({
+                    let state = state.clone();
+                    async move {
+                        tokio::time::sleep(Duration::from_millis(SNAPSHOT_DEBOUNCE_MS)).await;
```
jmoggr: I think there is a second race condition here: this task can wake up and call `create_snapshot` at basically the same time as `set_current_snapshot` is running. Which I think leads to some sort of race in the db `UPDATE`, I haven't fully traced this out, I think it's less concerning than the other race condition.

jmoggr (changes requested):
The asynchronous dance we do around debouncing changes and snapshots likely has at least one race condition and is otherwise really hard to reason about. The use of flags for asynchronous coordination is usually something that should be avoided in favor of primitives that are easier to follow and have better guarantees, this is not how asynchronous rust is meant to be written.

I admit I don't know a better way of writing this off the top of my head, debouncing makes this quite hard.

I could be wrong about the race conditions, but my inability to tell if there is a race condition should be enough of a concern.

---

> The undo history is global per-document meaning you will undo someone else's changes if you are working on the same doc.

This is unfortunate. This can degenerate very easily: one user hits undo, but sees no change so hits undo a few more times, another user sees their change undone and either thinks the app is broken or starts hitting undo themselves. Either way both users report the app as broken. 

I think that global undo/redo is better than no undo/redo, but I just want to flag that this will likely lead to occasional bug reports.

---

The intention has been to get rid of the JSON snapshots, we were fairly close to being able to remove them, but this turns them into a hard dependency. I think it's probably reasonable to keep them if they are a requirement for this feature, but I want to flag that we will no longer have the option of removing them.

 This is an improvement on snapshotting, since we're only creating snapshots every 500ms instead of every change.

---

I'll take a couple hours this afternoon to experiment with alternative ways of handling the asynchronous parts and will follow up here. It could be the case that there is no better solution, but that would feel fairly bad.

epatters:
Thanks for reviewing, Jason.

> The intention has been to get rid of the JSON snapshots, we were fairly close to being able to remove them, but this turns them into a hard dependency. I think it's probably reasonable to keep them if they are a requirement for this feature, but I want to flag that we will no longer have the option of removing them.

I'm not yet tracking the technical details in this PR, but I'm surprised that this feature would depend on JSON snapshotting (even if we might want JSON snapshots around for other reasons, mainly to allow DB queries that reach into documents). Is the undo/rollback performed using JSON snapshots rather than the Automerge doc's own history?

jmoggr:
> Is the undo/rollback performed using JSON snapshots rather than the Automerge doc's own history

Yes. iirc Kaspar mentioned something about having difficulty getting the API to do what was needed.

packages/backend/src/document.rs
kasbah: One immediate solution that springs to mind is to call our lock `modifying_current_snapshot` and both autosave and `set_current_snapshot` need to acquire it to modify the `current_snapshot` on `refs`. What do you think?

kasbah:
Please do check the code before you make assertions like that! I probably did say something like that before but I am using Automerge heads not the JSON snapshot now. 


https://github.com/ToposInstitute/CatColab/blob/4af3b122ac61775f25f5f427706f5e9da9536c28/packages/backend/src/document.rs#L226-L234 


https://github.com/ToposInstitute/CatColab/blob/4af3b122ac61775f25f5f427706f5e9da9536c28/packages/notebook-types/src/automerge_util.rs#L5-L36

epatters:
That's a relief! And yeah, echoing the doc comment, if we went through JSON, rich text would be busted.

kasbah:
> I think that global undo/redo is better than no undo/redo, but I just want to flag that this will likely lead to occasional bug reports. 

Yeah I suggest we shelve it for now. Happy to discuss it on this [Zulip thread](https://catcolab.zulipchat.com/#narrow/channel/474783-development/topic/Local.20vs.20global.20undo.20in.20collaborative.20documents) though where I started looking at a possible algorithm to use. @jmoggr

packages/backend/src/document.rs
kasbah: It's worth noting that it is not a problem if a snapshot gets created by autosave while we are undoing/redoing. That will currently just be a very short (one snapshot) branch that nobody sees.

The problem is if `current_snapshot` gets overridden by autosave when it is supposed to be set by undoing/redoing.

kasbah: So in addition to autosave needing to acquire the lock it would also need to ignore the changes coming in when it can't acquire the lock. 

So it's similar to the current design, but autosave _can_ lock it too and `set_current_snapshot` _will_ wait for the lock, and not ignore requests.

kasbah: It is a dance for sure, so curious if we can come up with something better, otherwise I think my proposed solution above avoids both races.

kasbah: I implemented my proposal: https://github.com/ToposInstitute/CatColab/pull/1175/changes/e72456fee0e2543d0f60cb2323c2e6eccf8e010a

packages/frontend/src/page/document_page.tsx
```diff
@@ -370,67 +404,102 @@ export function DocumentPane(props: {
...
+                    history.onRedo();
+                }
+            } else {
+                if (history.canUndo()) {
+                    evt.preventDefault();
+                    history.onUndo();
+                }
+            }
+        }
+    });
```
kasbah: I noticed there is an issue with this and the two-pane layout. It will undo/redo for both sides.

packages/backend/src/autosave.rs
```diff
@@ -0,0 +1,67 @@
...
+                        let lock = state.snapshot_lock(ref_id).await;
+                        let _guard = match lock.try_lock() {
+                            Ok(guard) => guard,
+                            Err(_) => {
+                                // we can't acquire the lock ignore this change
+                                return;
+                            }
+                        };
```
kasbah: We should couple this with a periodic check if we missed any changes that are currently on the doc and save them.

packages/frontend/src/page/document_page.tsx
epatters: Yeah, the situation with keyboard shortcuts has been a mess from the beginning. To work around this, we are currently disabling shortcuts like Shift-Enter in the right pane when in a two-pane layout.

See this longstanding issue:

- https://github.com/ToposInstitute/CatColab/issues/161

packages/backend/src/document.rs
jmoggr: my proposal: https://github.com/ToposInstitute/CatColab/pull/1184

The actor pattern is easier to reason about compared to locks and gives us the same semantics with high confidence that we won't deadlock ourselves in the future.

jmoggr:
> Please do check the code before you make assertions like that!

Apologies. I had read the code but my mental model got screwed up somewhere.

packages/backend/src/document.rs
kasbah: Merged, thanks!

packages/frontend/src/page/document_page.tsx
kasbah: ~My quick fix: https://github.com/ToposInstitute/CatColab/pull/1175/changes/e39998dbdf77b87bdd422bb4d85c15782bdeda56~ 

No that didn't fully solve it. I'm punting on it and not adding keyboard shortcuts here. https://github.com/ToposInstitute/CatColab/pull/1186

epatters: Very reasonable.

jmoggr (approved):
I think it's ready!

I think I saw some weird/surprising behavior, however:
- I was unable to reliably reproduce any of it
- It only happened when I was trying to break things
- the app never actually broke

It's possible that whatever I saw was due to local dev stuff. I don't think this is an issue at all, but I wanted to mention it incase we get other annoyingly vague reports.

epatters:
Thanks both! I will review ASAP.

packages/backend/src/document.rs
```diff
@@ -175,29 +141,91 @@ pub async fn create_snapshot(state: AppState, ref_id: Uuid) -> Result<(), AppErr
...
+    let mut db_tx = state.db.begin().await?;
+
+    let snapshot = sqlx::query!(
+        "SELECT heads FROM snapshots WHERE id = $1 AND for_ref = $2",
+        snapshot_id,
+        ref_id,
+    )
+    .fetch_optional(&mut *db_tx)
+    .await?
+    .ok_or_else(|| AppError::Invalid("Snapshot not found for this ref".to_string()))?;
```
epatters: Our current error handling in `rpc` converts sqlx `RowNotFound` errors into 404s, so it seems better to call `fetch_one(...).await?;` here than to return a 400 when the snapshot is not found.

Alternatively, if that seems too implicit and you want to include an informative error message, we could add a new error variant `NotFound` to `AppError`.

packages/backend/src/document.rs
```diff
@@ -175,29 +141,91 @@ pub async fn create_snapshot(state: AppState, ref_id: Uuid) -> Result<(), AppErr
...
+        })
+        .collect::<Result<_, _>>()?;
+
+    sqlx::query!("UPDATE refs SET current_snapshot = $2 WHERE id = $1", ref_id, snapshot_id,)
+        .execute(&mut *db_tx)
+        .await?;
+
+    doc_handle.with_document(|doc| {
+        doc.transact::<_, _, automerge::AutomergeError>(|tx| copy_doc_at_heads(tx, &target_heads))
+            .map_err(|e| AppError::Invalid(format!("Failed to update document: {e:?}")))
```
epatters: In this case, we should just use the `AppError::Automerge` variant, right?

packages/backend/src/rpc.rs
```diff
@@ -45,14 +45,21 @@ async fn get_doc(ctx: AppCtx, ref_id: Uuid) -> RpcResult<RefDoc> {
...
 
         if max_level >= Some(PermissionLevel::Write) {
-            let doc_id = doc::get_doc_id(ctx.state, ref_id).await?;
+            let doc_id = doc::get_doc_id(ctx.state.clone(), ref_id).await?;
+            let doc_handle = ctx
+                .state
+                .repo
+                .find(doc_id.clone())
+                .await?
+                .ok_or_else(|| AppError::Invalid("Document not found".to_string()))?;
```
epatters: Ditto: this should return a 404, not a 400.

packages/notebook-types/bindings/user_state.ts
epatters, line 1: We decided not to commit the generated bindings, right?

epatters (changes requested):
Looks great! Just a few small comments below, mostly related to error handling.

epatters (approved):

Issue #960, opened 2026-01-24
Flexible bindings of editors to object/morphism types


kasbah:
When we have a logics that share a common theory we would like more flexibility in switching out the editor that they use. We can e.g. switch the arrow styles at the moment but it's too rigid. 

One example would be wanting to switch out the delimiters in Petri net vs a reaction network (not yet implemented)

<img width="442" height="126" alt="Image" src="https://github.com/user-attachments/assets/d31b2d40-a0e5-4811-92e3-1203566c660e" />

<img width="442" height="175" alt="Image" src="https://github.com/user-attachments/assets/75818364-f807-44b2-aa61-d730c6457d1b" />

https://github.com/ToposInstitute/CatColab/blob/8f9dc9c7b19908d1561d03262fb60f7042614f8a/packages/frontend/src/model/object_list_editor.tsx#L23-L29

epatters:
To expand a bit: for each object type (resp. morphism type) bound in theory, it should be possible to specify a custom component to edit object declarations (rep. morphism declarations) of that type. The props to the component should include at least a getter and mutator for an `ObDecl` (resp. `MorDecl`) and the theory, as well as likely notebook-specific things like the available notebook actions.

This feature is intended to enable:

- changing settings of the default/existing editors (as in the motivating example above)
- using completely new editors (for example, as in #955, or to create a suitable interface for contributions in an ODE model, as in RFC-0001).

kasbah:
@epatters I just discussed with @tim-at-topos and I will take this on for before the workshop.

PR #1176, opened 2026-03-31
Logic for systems of polynomial ODEs

tim-at-topos:
This PR is going to completely ignore all UI issues (blocked by #960) and just:

- [x] Present the logic to the user
- [x] Implement equations analysis
- [x] Implement simulation analysis

Stacked on https://github.com/ToposInstitute/CatColab/pull/1162


tim-at-topos:
@tslil-topos @epatters you can see some `FIXME` throughout this PR, all relating to issues way over my head with `Unital` vs `NonUnital` issues. I fixed these in the bluntest way possible just to check that the rest of the code in this PR actually works (which it now does), but I would appreciate some help in fixing these before I mark this as ready for review!

packages/catlog-wasm/src/model.rs
```diff
@@ -261,8 +261,11 @@ pub enum DblModelBox {
     Discrete(Rc<dbl_model::DiscreteDblModel>),
     /// A model of a discrete tabulator theory.
     DiscreteTab(Rc<dbl_model::DiscreteTabModel>),
-    /// A model of a modal double theory.
+    /// A (unital) model of a modal double theory.
     Modal(Rc<dbl_model::ModalDblModel<Unital>>),
+    /// A (non-unital) model of a modal double theory.
+    // FIXME: This seems very not-correct.
+    ModalNU(Rc<dbl_model::ModalDblModel<NonUnital>>),
```
epatters: Unfortunately, I think this is correct :) The reason is that types exposed by wasm-bindgen do not support generics (type parameters), thus the two versions of modal models have to be explicitly present in different variants of the enum.

This can be considered a downside to the choice of encoding the theory kind at the type level.

packages/catlog-wasm/src/model.rs
```diff
@@ -371,15 +380,22 @@ impl DblModel {
...
         (&self.model).try_into().map_err(|_| "Model should be of a modal theory".into())
     }
 
+    /// Tries to get a non-unital model of a modal theory.
+    // FIXME: this should not be a whole new function, but I'm confused about how to do this
+    // without dependent types
+    pub fn modal_non_unital(&self) -> Result<&Rc<dbl_model::ModalDblModel<NonUnital>>, String> {
+        (&self.model).try_into().map_err(|_| "Model should be of a modal theory".into())
+    }
+
```
epatters, line 392: This one should be fixable. I haven't tried it, but does the following work?

```rust
pub fn modal<Kind: DblTheoryKind>(&self) -> Result<&Rc<dbl_model::ModalDblModel<Kind>>, String> {
  (&self.model).try_into().map_err(|_| "Model should be of a modal theory".into())
}
```

packages/catlog-wasm/src/theory.rs
```diff
@@ -230,8 +230,11 @@ pub enum DblTheoryBox {
     Discrete(Rc<theory::DiscreteDblTheory>),
     /// A discrete tabulator theory.
     DiscreteTab(Rc<theory::DiscreteTabTheory>),
-    /// A modal double theory.
+    /// A (unital) modal double theory.
     Modal(Rc<theory::ModalDblTheory<Unital>>),
+    /// A (non-unital) modal double theory.
+    // FIXME: This seems very not-correct.
```
epatters: This is also actually correct, for the reason given above.

epatters (commented):
Cool to hear that this PR is basically working! I commented on your FIXMEs as requested (but have otherwise not reviewed the PR).

tslil-topos:
@tim-at-topos i have worked through many of these in https://github.com/ToposInstitute/CatColab/pull/1167 

in particular see:
* https://github.com/ToposInstitute/CatColab/pull/1167/changes#diff-ccd09747286e67d752c93e2ad1121ba40f28e138a1addf9cd00f686c15916d0eR267
* https://github.com/ToposInstitute/CatColab/pull/1167/changes#diff-ccd09747286e67d752c93e2ad1121ba40f28e138a1addf9cd00f686c15916d0eR284
* https://github.com/ToposInstitute/CatColab/pull/1167/changes#diff-ccd09747286e67d752c93e2ad1121ba40f28e138a1addf9cd00f686c15916d0eR284
* https://github.com/ToposInstitute/CatColab/pull/1167/changes#diff-ccd09747286e67d752c93e2ad1121ba40f28e138a1addf9cd00f686c15916d0eR385

etc

it would be great if we could either separate this out from your or my change and or agree on naming etc. Happy to talk through this or whatever works

tim-at-topos:
@tslil-topos ah ok, I think I'll just follow your naming (it's better). It sounds like the cleanest thing to do would be a completely separate PR for just this one propagated change (introducing `ModalUnital` and `ModalNonUnital`) that we can both then rebase onto?

tslil-topos:
that would be great, happy to let you lead the charge since this change is basically already cleanly divisible, or i can have a go at doing the same?

tim-at-topos:
I'm on it! This is now https://github.com/ToposInstitute/CatColab/pull/1190

tim-at-topos:
@epatters composition doesn't work, in that I can instantiate a model and the dropdown knows about all the variables inside that model, but doesn't give me any options of variables in the larger model for assignment — the screenshot shows the left-hand side working, but the right-hand side is always empty

Screenshot 2026-04-07 at 18 02 11

I'm guessing that maybe something that currently just says `ModalUnital` (previously `Modal`) now needs to be updated to also include `ModalNonUnital`, but I don't know where — do you have an easy suggestion for a first place to look? Or is this just dependent on https://github.com/ToposInstitute/CatColab/issues/1181 ?

epatters:
@tim-at-topos, to make compositional models work with non-unital theories, you'll need to add a variant [here](https://github.com/ToposInstitute/CatColab/blob/8df5025318a5b940eddba4a63de0107134136b5f/packages/catlog/src/tt/modelgen.rs#L23), analogous to catlog-wasm.

I'd suggest adding a unit test for a non-unital theory in the text-based elaborator and only proceed with frontend testing once that works.

tim-at-topos:
@tslil-topos this is part of what you're doing in #1167 , right? if so, would you be able to copy that over to #1190 ?

edit: actually, this is https://github.com/ToposInstitute/CatColab/issues/1181 , right? I'm getting confused

tslil-topos:
unfortunately i have not yet wired up double tt for non-unital, i hacked around this in #1167 , here https://github.com/ToposInstitute/CatColab/pull/1167/changes#diff-ccd09747286e67d752c93e2ad1121ba40f28e138a1addf9cd00f686c15916d0eR753

there's a fallback path that you can use that doesn't pass through the DoubleTT code. We would also need to the stuff outlined here https://github.com/ToposInstitute/CatColab/pull/1190#issuecomment-4204912699 and likely more beside

tslil-topos:
ok looks like #1197 covered a fair portion of this

> unfortunately i have not yet wired up double tt for non-unital, i hacked around this in #1167 , here https://github.com/ToposInstitute/CatColab/pull/1167/changes#diff-ccd09747286e67d752c93e2ad1121ba40f28e138a1addf9cd00f686c15916d0eR753
> 
> there's a fallback path that you can use that doesn't pass through the DoubleTT code. We would also need to the stuff outlined here [#1190 (comment)](https://github.com/ToposInstitute/CatColab/pull/1190#issuecomment-4204912699) and likely more beside

tim-at-topos:
composition now works!
Screenshot 2026-04-08 at 12 58 54

tim-at-topos:
@epatters The following test is currently failing ("called `Option::unwrap()` on a `None` value"), and I'm trying to figure out whether it's #1197 or me (this is my first time interacting with ``tt::modelgen`, and the only example I can find is in `sql.rs`).

```rust
fn ode_system_from_text() {
    let th = Rc::new(th_polynomial_ode_system());
    let model = tt::modelgen::Model::from_text(
        &th.into(),
        "[
            X : State,
            Y : State,
            A : State,
            f : Contrib[[X, Y, Y], A],
            g : Contrib[[X, X], Y],
            h : Contrib[[A], X]
        ]",
    );
    let model = model.ok().and_then(|m| m.as_modal_non_unital()).unwrap();
    let sys = PolynomialODEAnalysis::default().build_system(&model);
    let expected = expect!([r#"
        dX = (h) A
        dY = (g) X^2
        dA = (f) X Y^2
    "#]);
    expected.assert_eq(&sys.to_string());
}
```

packages/catlog/src/stdlib/analyses/ode/polynomial_ode.rs
```diff
@@ -75,20 +75,18 @@ impl PolynomialODEAnalysis {
...
-            let output = model.get_cod(&mor).unwrap();
-            let term: Monomial<_, _> =
-                inputs.iter().map(|ob| (ob.clone().unwrap_generator(), 1)).collect();
 
+            let term: Monomial<_, _> =
+                inputs.iter().cloned().map(|ob| (ob.unwrap_generator(), 1)).collect();
             let term: Polynomial<_, _, _> =
                 [(Parameter::generator(mor), term.clone())].into_iter().collect();
 
-            // TODO: only a single output
```
epatters, line 90: @tim-at-topos, I wasn't sure what this TODO meant so I deleted it. If it's important, please file an issue.

PR #1180, opened 2026-04-01
Editor variants and a string diagram variant for Petri nets

### LINKED ISSUE #955: Editable widget for multiary morphism generators in string diagram style
We would like to have an alternative editing widget for the Petri net morphisms that could be called the "string diagram style".

This is related to https://github.com/ToposInstitute/CatColab/issues/244 in that it may share a similar interface on the left side.

This is an opportunity to assess and improve the flexibility of binding editors to objects/morphism types. i.e. #960 

Some more notes from the whiteboard about this feature:

Image

### LINKED ISSUE #960: Flexible bindings of editors to object/morphism types

kasbah:
closes #960  
closes #955 

image

hover:

image


packages/frontend/src/model/document.ts
```diff
@@ -57,37 +58,72 @@ export async function createModel(
     return api.createDoc(init);
 }
 
-/** Migrate a model document from one theory to another. */
-export async function migrateModelDocument(
+/** Migrate a model document to a different theory, or switch its editor variant.
+ */
+export async function switchTheoryOrEditor(
```
epatters: I really think these should be two separate functions since they're doing conceptually very different things (even if the so-called "trivial" migrations blur that a bit).

As my TODO comment says, the model migration code here is supposed to be temporary: non-trivial migrations involve changing the cell content of the notebook, and I aim for that to be done in Rust by a more principled mechanism. OTOH, toggling the editor variant is clearly the provenance of the frontend.

packages/frontend/src/model/document.ts
```diff
@@ -11,10 +11,11 @@ import type { ValidatedModel } from "./model_library";
 export type ModelDocument = Document & { type: "model" };
 
 /** Create an empty model document. */
-export const newModelDocument = (theory: string): ModelDocument => ({
+export const newModelDocument = (theory: string, editorVariant?: string): ModelDocument => ({
```
epatters: Not a big deal, but I'd be inclined to make these named arguments now:

```ts
(args: {theory: string, editorVariant?: string}): ModelDocument => ...
```

packages/frontend/src/model/model_editor.tsx
```diff
@@ -54,32 +56,60 @@ export function ModelNotebookEditor(props: { liveModel: LiveModelDoc }) {
...
-                        actions={props.actions}
-                        theory={theory()}
-                    />
-                )}
+                {(theory) => {
+                    const obDecl = () => props.content as ObDecl;
+                    const editor = () => {
+                        const override = editorOverrides()?.obEditors?.find((o) =>
+                            deepEqual(o.obType, obDecl().obType),
+                        );
```
epatters: The type `ObTypeIndex` in `catlog_wasm::theory` might make this nicer: no iteration with deep equality checking, just a hash map lookup. This helper is already used in the `Theory` class in the frontend.

As always, it's sad and bewildering that JS doesn't have a hash map with arbitrary JSON keys, but oh well.

packages/frontend/src/model/model_editor.tsx
```diff
@@ -54,32 +56,60 @@ export function ModelNotebookEditor(props: { liveModel: LiveModelDoc }) {
...
-                        isActive={props.isActive}
-                        actions={props.actions}
-                        theory={theory()}
-                    />
-                )}
+                {(theory) => {
+                    const morDecl = () => props.content as MorDecl;
+                    const editor = () => {
+                        const override = editorOverrides()?.morEditors?.find((o) =>
+                            deepEqual(o.morType, morDecl().morType),
```
epatters: Ditto: might be simplified by using `MorTypeIndex`.

packages/frontend/src/model/string_diagram_morphism_cell_editor.tsx
```diff
@@ -0,0 +1,340 @@
...
+    return null;
+}
+
+/** Wrap an `ob` with `App(op, ob)`. */
+function wrapApp(ob: Ob | null, applyOp: ObOp | undefined): Ob | null {
+    if (!ob || !applyOp) {
+        return ob;
+    }
+    return { tag: "App", content: { op: applyOp, ob } };
+}
```
epatters: These helpers are duplicative with the wrapping/unwrapping performed in `src/model/object_input`.

Behind this is the more important point that this component would ideally not bake in the assumption about a particular operation being applied (cf. similar comment below), which couples it unnecessarily closely to a particular theory. To be more specific, a Petri net can be viewed as free symmetric monoidal categories (SMCs) *or* free PROPs (or other things besides), and we might want to reuse this component for such situations. However, if it's not clear what to do, we can defer this to a future PR.

packages/frontend/src/model/string_diagram_morphism_cell_editor.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+/** Extract the list of objects from a domain/codomain `Ob`. */
+function getObList(ob: Ob | null, applyOp: ObOp | undefined): Array<Ob | null> {
```
epatters: Similarly, this is duplicative with what's in `src/model/object_list_editor`. Perhaps we should extract these helpers into a separate module and reuse them.

packages/frontend/src/model/string_diagram_morphism_cell_editor.tsx
```diff
@@ -0,0 +1,340 @@
...
+    const codApplyOp = () => morTypeMeta()?.codomain?.apply;
+
+    /** Rebuild a domain/codomain Ob from a list of objects. */
+    const makeObList = (objects: Array<Ob | null>, applyOp: ObOp | undefined): Ob | null => {
+        if (!applyOp) {
+            return null;
+        }
+        const domObjType = props.theory.theory.dom(applyOp);
+        const modality: Modality =
+            domObjType?.tag === "ModeApp" ? domObjType.content.modality : "SymmetricList";
```
epatters: We shouldn't have to bake `SymmetricList` in here, just like it's not baked in `ObjectListEditor`. But again, we can defer this to later if you'd like.

packages/frontend/src/stdlib/theories/petri-net.ts
```diff
@@ -9,6 +11,26 @@ export default function createPetriNetTheory(theoryMeta: TheoryMeta): Theory {
         ...theoryMeta,
         theory: thSymMonoidalCategory.theory(),
         onlyFreeModels: true,
+        editorVariants: [
+            {
+                id: "editor-variant-petri-net-string-diagram",
```
epatters: It's not important that this ID begin with `editor-variant-`, is it?

epatters (changes requested):
This is great, thanks Kaspar! I like the overall conception for "editor variants," and the string-diagram-style editor variant Petri nets is very cool.

To my questions/comments below I would add that I agree with @tim-at-topos that the editor variants shouldn't appear in the list of logics/theories but should be settable via some other widget in the model notebook. But we don't have to figure that out right now.

packages/frontend/src/stdlib/theories/petri-net.ts
kasbah: No, I added that just for clarity if it pops up anywhere.

kasbah:
I believe I've addressed all feedback and I've changed this to be behind a settings pane like the analysis settings. 

https://github.com/user-attachments/assets/47005a74-3ec6-4a38-b9fb-78a634b3b61f

epatters (approved):
Thanks, looks wonderful!

PR #1189, opened 2026-04-07
RFC-0004: Internal languages for models

epatters:
Preview: https://branch-rfc-pointful-syntax--catcolab.netlify.app/rfc/0004


epatters:
This RFC isn't finished, but for bureaucratic reasons I'm going to merge it now. We'll track it in #1212 and I'll follow up with at least one more PR.

PR #1192, opened 2026-04-07
Remove stem cells

kasbah:
- Doesn't create an empty "stem" cell when you unfocus without selecting anything
- Newly created cells grab focus


epatters (changes requested):
I'm afraid this one doesn't seem to work properly for me (running frontend locally, on Firefox, rebased off main). When I press Shift+Enter with the active cell a stock at the top of a stock-flow notebook, the cell constructor menu appears at the end of the notebook.

Beyond that, it seems that this PR effectively removes stem cells from the notebook in that it is no longer possible to create them. I'm not committed to the stem cell concept (which was modeled on Notion). However, if we are going to replace it, we should go further and remove it as fully as practical from the codebase, though a complete removal would require a migration filtering notebook cells, which is annoying. In its current form, this PR seems to leave significant vestigial code.

kasbah:
I believe I fixed the UI issues and pushed a commit with a document-types v2 and stem cells purged everywhere.

packages/frontend/src/notebook/notebook_editor.tsx
```diff
@@ -151,35 +133,108 @@ export function NotebookEditor<T>(props: {
...
+            setShiftEnterCellIndex(cellIndex);
+
+            // Position the popup at the bottom edge of the active cell.
+            const cellEls = document.querySelectorAll(".notebook-cells .cell");
+            const cellEl = cellIndex != null ? cellEls[cellIndex] : cellEls[cellEls.length - 1];
+            if (cellEl) {
+                const rect = cellEl.getBoundingClientRect();
+                setPopupLeft(rect.left);
+                setPopupTop(rect.bottom);
+            }
```
epatters: Seems a bit messy and unidiomatic to have to select the notebook element with a document query and then manually position the popover. Also, is this even correct when there are multiple notebook elements in the DOM?

Perhaps there are other problems that I'm not anticipating, but I believe the idiomatic approach is to have each `NotebookCell` contain its own `Popover` component to display the constructor menu below it. The popover can then be positioned automatically and we don't have to track all this state (i.e., all the new signals added above) in the notebook editor itself.

epatters (changes requested):
Thanks for propagating this change all the way through the data types. Looks much better!

Since I'm not currently set up to run the backend locally, I haven't tested it myself. I'm fine to merge this PR having only reviewed the code, about which I have just one comment below.

packages/frontend/src/notebook/notebook_editor.tsx
kasbah: I've pushed a cleaner solution. There is a bug that is present on next too: if you have two documents open side by side the new cell on "Shift-Enter" will always be on the left hand side document.

kasbah: Oops, there are some issues with my "cleaner solution". I'll look at this next week.

epatters: Sounds good, thanks.

And yep, that issue has always been there due our problems with focus management across the app. We should just avoid digging a deeper hole for ourselves :)

kasbah: Had to resort to [e62ceb7](https://github.com/ToposInstitute/CatColab/pull/1192/commits/e62ceb7e3eb11610a42a6eb0c437e05af5a4a6cc) which prompted me to look at focus overall. That's the best I can do for now. Ready for re-review.

epatters (approved):
Thanks Kaspar, this is much cleaner!

PR #1194, opened 2026-04-07
Logic for signed polynomial ODEs

tim-at-topos:
To-do:

- [x] validate coefficient values in simulation as non-negative iff in the *signed* logic
- [x] rebase onto https://github.com/ToposInstitute/CatColab/pull/1211 and implement UI changes


epatters:
You're getting fast at this! :)

packages/frontend/src/stdlib/theories/signed-polynomial-ode.ts
```diff
@@ -0,0 +1,59 @@
...
+                shortcut: ["V"],
+            },
+            {
+                tag: "MorType",
+                morType: { tag: "Basic", content: "Contribution" },
+                editor: ContributionCellEditor,
+                name: "Positive contribution",
+                description: "Additive monomial contribution to the system of ODEs",
+                shortcut: ["C"],
+                arrowStyle: "plus",
```
tim-at-topos: This use of `arrowStyle` is a little bit of a hack, since although it does technical describe the "arrow" style, it's only used to pass this information along to `ContributionCellEditor`. I think it would be cleaner to allow passing options into editors like this (or maybe there's already an easy way to do this? @kasbah)

Either way, I don't think this PR should stall on this point, but I wanted to mention it!

epatters: Yep, I raised a similar point in my review of #1200. Agreed that we don't need to worry about it at this moment.

epatters (approved):

PR #1200, opened 2026-04-08
Make default editors for objects and morphisms explicit

kasbah:
Building on #1180 this makes the default editors for a model/theory explicit so they can be changed for e.g. https://github.com/ToposInstitute/CatColab/pull/1211

### DIFF
--- a/packages/frontend/src/model/morphism_cell_editor.tsx
+++ b/packages/frontend/src/model/morphism_cell_editor.tsx
@@ -11,7 +11,7 @@ import arrowStyles from "../stdlib/arrow_styles.module.css";
 import "./morphism_cell_editor.css";
 
 /** Editor for a morphism declaration cell in a model. */
-export function MorphismCellEditor(props: MorphismEditorProps) {
+export default function MorphismCellEditor(props: MorphismEditorProps) {
     const liveModel = useContext(LiveModelContext);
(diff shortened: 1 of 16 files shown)


packages/frontend/src/model/morphism_cell_editor.tsx
```diff
@@ -11,7 +11,7 @@ import arrowStyles from "../stdlib/arrow_styles.module.css";
 import "./morphism_cell_editor.css";
 
 /** Editor for a morphism declaration cell in a model. */
-export function MorphismCellEditor(props: MorphismEditorProps) {
+export default function MorphismCellEditor(props: MorphismEditorProps) {
```
epatters, line 14: I wonder whether, now that we have multiple editors for morphism cells, we want to give this one a more descriptive name, such as `ArrowMorphismCellEditor`.

More substantively, there is now at least one option (`arrowStyle`) in `ModelMorTypeMeta` that only makes sense for *this* editor. In future work, we should go through the theory metadata carefully and extract any fields that would not make sense for *any* editor.

epatters (approved):
Thanks, this seems sensible!

epatters:
I'll merge this as-is and save for later the job of renaming the default morphism editor, as well as the more substantial refactoring.

PR #1201, opened 2026-04-08
RFC-0005: widening the definition of model, introducing the notion of instance

tslil-topos:
This is designed to be a more verbose description of what's going on in #1167

https://branch-tslil-push-ynzqnltmyznx--catcolab.netlify.app/rfc/0005

### DIFF
--- a/rfc/0005.md
+++ b/rfc/0005.md
@@ -1,8 +1,8 @@
 ---
 title: |
-  `0005` Separating models and instances of double theories
+  `0005` Fibrational definition of model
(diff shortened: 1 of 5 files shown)


tslil-topos:
After in-person discussions with @epatters , it would seem that the best path forward for now is to close this PR. Rejecting or accepting it would impose the same burden: both deciding on the larger shape of how we want the codebase and mathematics to evolve, and explaining why or why not this forms part of that vision.

PR #1209, opened 2026-04-09
Add "Last edited" document info, sort document lists by it

kasbah:
Based on #1175, if we are happy with this and want to merge then I should consolidate the migrations and actually merge into that branch before merging into main. 

image


packages/ui-components/src/relative_time.tsx
epatters, line 1: Ideally, we wouldn't have to put this stuff in our codebase. What about using this package instead? https://github.com/github/relative-time-element

kasbah: Good point, I used that but kept a thin wrapper around it in ui-components (for the type declaration mainly) and kept the story so that ppl perusing ui-components know we have something for this.

epatters (approved):

PR #1210, opened 2026-04-09
Add document picker to instantiations

kasbah:
based on #1209 

image

### DIFF
--- a/packages/frontend/src/components/document_picker.tsx
+++ b/packages/frontend/src/components/document_picker.tsx
@@ -258,6 +258,12 @@ function DocSearchInput(
                 showCompletionsOnFocus={true}
                 popupClass="document-picker-completions"
                 completionsEmptyText="No results"
+                hasBlurred={() => {
+                    const text = inputText();
+                    if (tryParseRefId(text) !== null) {
+                        onSubmit(text);
+                    }
+                }}
                 interceptKeyDown={(evt) => {
                     if (evt.key === "Escape" && props.onCancel) {
                         props.onCancel();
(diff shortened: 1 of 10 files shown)


packages/frontend/src/components/document_picker.tsx
```diff
@@ -109,80 +133,157 @@ export function DocumentPicker(
...
  */
-function RefInput(
+function DocSearchInput(
     allProps: TextInputOptions & {
         refId: Uuid | null;
         setRefId: (refId: Uuid | null) => void;
         onCancel?: () => void;
         docType?: Document["type"];
+        currentRefId?: string;
+        theoryId?: string;
```
epatters: I'd suggest factoring this differently since we're going to want to reuse the document picker component in different contexts. That's why this component isn't in `src/model/`. For example, we'll want to use it for picking instances when instances support composition (something Kevin is starting to think about).

One idea would be to replace the `theoryId` prop with a predicate `filterCompletions`, then have the model instantiation editor pass in a predicate similar to the one here.

packages/frontend/src/model/instantiation_cell_editor.tsx
```diff
@@ -19,11 +21,24 @@ export function InstantiationCellEditor(props: {
     actions: CellActions;
 }) {
     const api = useApi();
+    const params = useParams();
```
epatters: It's unfortunate to have a component this deep in the hierarchy have to reach all the way up to the router to pull out a parameter.

I'm not sure how best to avoid it, though. I agree with you that we don't want the active document to appear in the list of completions.

Somewhat better, IMO, would to have `DocumentPane` create a Solidjs Context containing the ref ID. Any other ideas?

epatters (changes requested):
Thanks Kaspar, this looks beautiful! I have a few architecture concerns though.

packages/frontend/src/model/instantiation_cell_editor.tsx
epatters: On further thought, I'd argue that the current implementation is incorrect since it assumes that this component appears in the left pane of a split view.

epatters (approved):

PR #1211, opened 2026-04-09
Improved UI for models of polynomial ODEs

tim-at-topos:
Stacked on #1203 (and thus #1180).


packages/frontend/src/model/contribution_cell_editor.css
epatters, line 1: Not a proper review, but noting that this CSS conflicts with that `morphism_cell_editor.css`.

Would suggest making this file a CSS module. (We should be using those more in general.)

epatters (approved):

PR #1240, opened 2026-04-26
Security: Unchecked `unwrap()` can panic on malformed model data (DoS risk)

tomaioo:
## Summary

Security: Unchecked `unwrap()` can panic on malformed model data (DoS risk)

## Problem

**Severity**: `Medium` | **File**: `packages/catlog/src/stdlib/analyses/ode/signed_coefficients.rs:L59`

In `build_matrix`, object index lookups for morphism domain/codomain use `.unwrap()`. If a morphism references an object not present in `ob_index` (e.g., inconsistent or maliciously crafted model data crossing trust boundaries), this causes a panic. In backend/WASM contexts, this can crash request handling or the module, resulting in denial of service.

## Solution

Replace `unwrap()` with checked handling (`if let Some(...)` / `ok_or(...)`), and return a structured validation error when model invariants are violated. Optionally validate model consistency before matrix construction.

## Changes

- `packages/catlog/src/stdlib/analyses/ode/signed_coefficients.rs` (modified)

### DIFF
--- a/packages/catlog/src/stdlib/analyses/ode/signed_coefficients.rs
+++ b/packages/catlog/src/stdlib/analyses/ode/signed_coefficients.rs
@@ -66,16 +66,22 @@ impl<ObType, MorType> SignedCoefficientBuilder<ObType, MorType> {
         let mut mat = DMatrix::from_element(n, n, zero());
         for mor_type in self.positive_mor_types.iter() {
             for mor in model.mor_generators_with_type(mor_type) {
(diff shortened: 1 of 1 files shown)


jmoggr:
I am fairly certain this is an AI slop PR. Forgive me if I'm wrong.

There are a lot of places we are doing unwraps in "mathy logic", this instance is not particularly noteworthy. While it would be good to systematically avoid the use of unwrap, that is currently out of scope.

The risk of this issue causing leading being the avenue for a denial of service attack is near zero, this code does not run on the server, and would just break a document loading/running for a client.

PR #1241, opened 2026-04-27
Replace `deepCopyJSON` with structured clones

kasbah:
Replaces our use of `JSON.parse(JSON.stringify(...))` workaround for cloning proxy objects.

Turned out more complicated than I expected because Automerge also uses proxies but doesn't provide a nice `unwrap` like Solid does.


epatters (approved):
Thanks Kaspar, this is indeed more involved than I was expecting (due to the need to handle both Automerge and Solid proxies), but it looks good!

PR #1242, opened 2026-04-27
 Extract handles for editing model objects and morphisms

kasbah:
This add some `*EditHandle` classes which help with common get/set operations on objects and morphisms


kasbah:
I'm not happy with the object oriented approach of this one but I haven't come up with anything better yet.

epatters:
Thanks for noting that. I'll think about it and hopefully say something more useful by Monday.

epatters (changes requested):
Thanks for taking a first pass at this refactor. I agree that with our palette of morphism editors having grown recently, there's now too much redundant boilerplate spread across them.

I also share your discomfort about this design. Besides the questionable use of old-fashioned inheritance, I overall feel that the new abstraction (the "edit handle") is not doing enough work too justifying adding a layer of indirection between the editors and document data.

A lot of the edit handle methods are simple one-liners. Am I correct that the two important things that this PR seeks to address are the following?

1. (Un)wrapping the (co)domain objects with operation applications
2. (Un)wrapping lists of objects that carry a modality

If so, what about this?

1. Lift the logic to wrap and unwrap (co)domains out of the individual editors. There's no reason that each individual editor should have to handle this. Refactor the interface for a morphism editor so that, rather taking getters/mutators for a whole `MorDecl`, it gets more granular getters/mutators (`dom`, `changeDom`, `cod`, `changeCod`, etc). Then the part of the frontend that dispatches on the choice of morphism editor should also read the theory metadata to appropriately unwrap the (co)domain in the getter and wrap it in the mutator.
2. Add some methods to manipulate lists of objects to `document-methods`. Unless I'm missing something, these shouldn't need to maintain any state, like the edit handle allows.

In this proposed design, no edit handle is needed.

kasbah:
Thanks, that makes sense, I'll make another PR.

PR #1243, opened 2026-04-28
Show completions of text inputs by default on focus

kasbah:
I'm not sure why we aren't defaulting to showing all the possible completions when clicking on an input.


epatters (approved):
I'm not sure why either but I agree that this behavior is a usability improvement. Thanks.

I thought it cleaner to change the default upstream than to pass `inputProps` by spread but then override one of the props. I went ahead and did that.

PR #1244, opened 2026-04-28
Improved UI for instantiation specializations

### LINKED ISSUE #891: Not obvious how to make multiple assignments in an instantiation

kasbah:
Improves instantiation specialization selection UI. Similar behavior to the string diagram transition editor. closes https://github.com/ToposInstitute/CatColab/issues/891 

Also prefixed the left-hand side with a "." because I think it's clearer. 

https://github.com/user-attachments/assets/24413f3a-8e34-4d02-8454-af8c4ff3fab2

### DIFF
--- a/packages/frontend/src/model/instantiation_cell_editor.tsx
+++ b/packages/frontend/src/model/instantiation_cell_editor.tsx
@@ -184,7 +184,20 @@ export function InstantiationCellEditor(props: {
                 <span class="is-a" />
                 <DocumentPicker
                     refId={refId() ?? null}
-                    setRefId={setRefId}
+                    setRefId={(newRefId) => {
+                        setRefId(newRefId);
+                        // After picking a model, move focus out of the picker
+                        // and into the specializations area so the user can
+                        // keep going on the keyboard. If there are no rows
+                        // yet, add an empty one to focus.
+                        if (newRefId) {
+                            if (props.instantiation.specializations.length === 0) {
+                                insertSpecializationAtTop();
+                            } else {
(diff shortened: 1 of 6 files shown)


packages/frontend/src/model/instantiation_cell_editor.tsx
```diff
@@ -231,7 +339,16 @@ function SpecializationEditor(
                 {...inputProps}
             />
             <span class="specialize-as" />
-            <Show when={obType()} fallback={<span class="placeholder">{"..."}</span>}>
+            <Show
+                when={obType()}
+                fallback={
+                    <div class="id-input">
```
epatters: I noticed that this snippet of code appears in several places. More importantly, it feels like breaking an encapsulation boundary to use the DOM structure of the `InlineInput` component here. Instead, I would suggest adding a documented `PlaceholderInlineInput` component to `inline_input` in `ui-components`.

epatters (changes requested):
Thanks, I really like how this feels!

In addition to the code comment below, I noticed a usability issue that we didn't have before: after selecting a model to instantiate using the document picker, focus is lost and the mouse must be used to get into the specializations. It would be good to fix this as I like being able to drive the app solely by keyboard.

epatters (approved):

PR #1246, opened 2026-05-05
Instances in DoubleTT

KevinDCarlson:
Here are some instances in doublett. Runnable for text elaborator, notebook elaborator TBD pending #1257


packages/catlog/examples/tt/text/test_instances.dbltt
```diff
@@ -0,0 +1,41 @@
...
+    we : Edge,
+    wf : Edge,
+    src(we.e) := v1,
+    tgt(we.e) := v2,
+    src(wf.e) := v2,
+    tgt(wf.e) := v1,
+    weight(we.e) := w,
+    weight(wf.e) := w
+]
+
```
KevinDCarlson, line 28: As the syntax suggests, an instance is now simply a term of the type of which it's an instance. We can declare generators explicitly with lists of new names, or by importing them from another instance (semantically, getting an instance map in from there.) We can specify mappings explicitly, as below, but the mappings need not be total, and are in that case implicitly filled out via the dopf property.

KevinDCarlson: OK, this changed a bit again: we needed to keep the standard `Cons` terms of dependent records for, primarily, internal reasons. These correspond semantically to tight transformations of models. I've thus reintroduced a new keyword for instance-type terms. These are expected to always be closed, thus corresponding to lax loose transformations from the point into a model--the open analogue doesn't seem obviously useful to me. So the semantics at the moment is in the category of instances of models, fibered over the category of models and tight transformations.

PR #1247, opened 2026-05-05
UI for editing path equations in schemas and ologs

### LINKED ISSUE #63: Editor for path equations in models
Path equations in models of discrete double theories are already implemented in the core. We need to add an editor for them in the frontend. Applications include adding equations to ologs and schemas.

kasbah:
Following on from #1061 this is the editing UI for equations. Closes https://github.com/ToposInstitute/CatColab/issues/63 .

### DIFF
--- a/packages/frontend/src/model/equation_cell_editor.tsx
+++ b/packages/frontend/src/model/equation_cell_editor.tsx
@@ -1,0 +1,197 @@
+import { createEffect, createMemo, createSignal, useContext } from "solid-js";
+import invariant from "tiny-invariant";
+import { P, match } from "ts-pattern";
+
+import {
+    type FocusHandle,
+    InlineListEditor,
+    NameInput,
+    useChildFocus,
+} from "catcolab-ui-components";
+import type { Mor, QualifiedName } from "catlog-wasm";
+import { MorIdInput } from "../components";
+import { removeProxyAndCopy } from "../util/remove_proxy_and_copy";
+import { LiveModelContext } from "./context";
+import type { EquationEditorProps } from "./editors";
+
+import styles from "./equation_cell_editor.module.css";
+
+type EquationCellInput = "name" | "lhs" | "rhs";
+
+/** Editor for an equation cell in a model. */
+export default function EquationCellEditor(props: EquationEditorProps) {
+    const liveModel = useContext(LiveModelContext);
+    invariant(liveModel, "Live model should be provided as context");
+
+    // oxlint-disable-next-line solid/reactivity -- Focus handles are stable for a mounted cell.
+    const focus = useChildFocus<EquationCellInput>(props.focus, { default: "name" });
+
+    const completions = (): QualifiedName[] | undefined =>
+        liveModel().elaboratedModel()?.morGenerators();
+
+    const lhs = createMemo(() => morToList(props.equation.lhs));
+    const rhs = createMemo(() => morToList(props.equation.rhs));
+
+    const setLhs = (mors: Array<Mor | null>) =>
+        props.modifyEquation((eqn) => {
+            eqn.lhs = removeProxyAndCopy(listToMor(mors));
+        });
+
+    const setRhs = (mors: Array<Mor | null>) =>
+        props.modifyEquation((eqn) => {
+            eqn.rhs = removeProxyAndCopy(listToMor(mors));
+        });
+
+    const setName = (name: string) =>
+        props.modifyEquation((eqn) => {
+            eqn.name = name;
+        });
+
+    return (
+        <div class={`formal-judgment ${styles.decl}`}>
+            <div class={styles.name}>
+                <NameInput
+                    placeholder="Unnamed"
+                    name={props.equation.name}
+                    setName={setName}
+                    focus={focus.childFocus("name")}
+                    deleteBackward={props.actions.deleteBackward}
+                    deleteForward={props.actions.deleteForward}
+                    exitBackward={props.actions.activateAbove}
+                    exitForward={() => focus.setActiveChild("lhs")}
+                    exitUp={props.actions.activateAbove}
+                    exitDown={() => focus.setActiveChild("lhs")}
+                    exitRight={() => focus.setActiveChild("lhs")}
+                    createBelow={() => focus.setActiveChild("lhs")}
+                />
+            </div>
+            <div class={styles.nameSeparator}>:</div>
+            <div class={styles.equation}>
+                <MorListEditor
+                    mors={lhs()}
+                    setMors={setLhs}
+                    completions={completions()}
+                    focus={focus.childFocus("lhs")}
+                    exitBackward={() => focus.setActiveChild("name")}
+                    exitForward={() => focus.setActiveChild("rhs")}
+                    exitLeft={() => focus.setActiveChild("name")}
+                    exitRight={() => focus.setActiveChild("rhs")}
+                    deleteBackward={() => focus.setActiveChild("name")}
+                    deleteForward={() => focus.setActiveChild("rhs")}
+                />
+                <div class={styles.equals}>{"="}</div>
+                <MorListEditor
+                    mors={rhs()}
+                    setMors={setRhs}
+                    completions={completions()}
+                    focus={focus.childFocus("rhs")}
+                    exitBackward={() => focus.setActiveChild("lhs")}
+                    exitForward={props.actions.activateBelow}
+                    exitLeft={() => focus.setActiveChild("lhs")}
+                    exitRight={props.actions.activateBelow}
+                    deleteBackward={() => focus.setActiveChild("lhs")}
+                    deleteForward={props.actions.activateBelow}
+                />
+            </div>
+        </div>
+    );
+}
+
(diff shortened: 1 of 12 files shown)


kasbah:
I could use some help formulating some brief statements about this to put into the logics docs.

KevinDCarlson (dismissed):
Very awesome. The bit of Rust looks unobjectionable and I'm not going to read the TypeScript, just user-side comments: 

- The widget is really big. I'd reduce whitespace in the blue-scribbled places. 

Image

- It seems inconsistent with the formatting of other notebook elements to have everything center-aligned. 

Image

- It looks like, once I say one branch of the equation, you're searching for things to equate to it that type check, since as here, `h ; k` and `f ; g` float to the top of the list, but then all the other simple paths are still displayed. But none of the other options would type-check, so I think it would be better to filter them out entirely.

Image

- The list of all simple paths tends to be really big, even for tiny models with no loops. I think I would add an explicit user interaction to generate the list from nothing, only generating automatically once a user types `f` or something else? 

- I can't think of any time when a fact would need a name. Even if it's in the data structures, as I guess it probably is in the type theory, I might hide it in the UI for now.

- Need an issue for making the visualization know about facts.

kasbah:
Thanks @KevinDCarlson 

## 1. 
> I'd reduce whitespace
> ... center-aligned. 

I pushed some commits to reduce spacing. I do want to mention that the spacing center alignment come from trying to make this look consistently good and easy to parse visually, so e.g.:

image




---
or
image




--- 
vs currently:
image


## 2.
>  none of the other options would type-check, so I think it would be better to filter them out entirely. 

I tried this before actually but I settled on a more dumb/lax editor because filtering like this makes the RHS act differently to the LHS. If filtering, the behavior feels unexpected: the user is used to seeing the full list of all paths and then suddenly can't find a path that they know exists. I tried some other stricter checking of valid paths but it all seemed to get in the way a bit. It feels useful to be able to have an invalid "draft" equation while you are editing.


## 3.

> The list of all simple paths tends to be really big, even for tiny models with no loops. I think I would add an explicit user interaction to generate the list from nothing, only generating automatically once a user types f or something else?

This wouldn't really work because without the completion there is no indication to the user what they should be typing.

Furthermore, I am not sure what this gets us unless there are real performance concerns. I haven't run into any though the most complicated model I tried is the CatColab DB Schema, so we should stress test it.

It also reminds me that it might make sense to limit the path length to 1000 or something just as a safe guard.

## 4. 

> I can't think of any time when a fact would need a name. Even if it's in the data structures, as I guess it probably is in the type theory, I might hide it in the UI for now.

Do you mean for "facts" for ologs specifically but "equations" should have names in schemas? I'm guessing you are just using "facts" as the catch-all but wasn't sure because you switched the term. I'm not sure if/why they would need names, it might be nice for the visualization or referring to them in other ways? I'll defer to others on that decision.

## 5.

> Need an issue for making the visualization know about facts

I opened https://github.com/ToposInstitute/CatColab/issues/1262, maybe we should add another issue for visualizing them as commutative diagrams?

KevinDCarlson:
Thanks, Kaspar.

> I do want to mention that the spacing center alignment come from trying to make this look consistently good and easy to parse visually, 

I see your point on the centering. I still don't love the compromise but maybe it's the best option available. 

> I settled on a more dumb/lax editor... If filtering, the behavior feels unexpected...

Maybe. I found it unexpected to see impossible options, personally; we don't autocomplete object names in the morphism boxes! I think I would still prefer filtering out invalid options, though it's a reasonable point that you might want to allow for impossible equations during drafting. 

> This wouldn't really work because without the completion there is no indication to the user what they should be typing.

If that's really the objection, then I would think that including an explicit prompt "type morphism names, separated by semicolons" would teach the interface more easily than the current implicit method, though I did really like that when you click back in to an existent equation branch the interface shows you how to type it. 

> Do you mean for "facts" for ologs specifically but "equations" should have names in schemas?

Sorry, yes, I meant equations in any theory. Basically, you'll never have two equations between the same two paths, so it's probably unnecessary to same them. Will let @epatters weigh in when he can. (Evan: any reason for a user-accessible name box for equations?)

> Maybe we should add another issue for visualizing them as commutative diagrams?

Interestingly, there is no traditional visual language for distinguishing between commutative and noncommutative diagrams; you just draw the graph and figure out what commutes from context. So I think this is probably subsumed in #1262 .

epatters (changes requested):
Thanks Kaspar! While this isn't a code review, I'm recording a few problems I noticed while testing.

1. Cosmetic issue: the vertical alignment of "E" in the bottom path is off:

    Image

2. Another cosmetic issue: the vertical sizing of "E" in the popover is off:

    Image

2. Logic issue: it is not currently possible to build path with repeat edges:

    Image

This is important because in practice a path in a path equation often uses the same generating morphism twice. In this example (the schema for symmetric graphs), I am trying to express that "inv" is indeed an involution, i.e., composing it with itself gives the identity.

kasbah:
I changed the approach completely to allow for repetition. It now uses the list editor to form paths and is based on #1326 .

epatters:
Thanks for the ping, and sorry that this sat for so long. I'd love to get this in as it will pair nicely with the new data features. I'll re-review this PR tomorrow.

packages/frontend/src/model/equation_cell_editor.tsx
```diff
@@ -0,0 +1,284 @@
...
+/** Build a basic object from its generator id. */
+function basicOb(id: Uuid): Ob {
+    return { tag: "Basic", content: id };
+}
+
+/** Domain of a morphism in the model, or `null` if it cannot be computed. */
+function morDom(model: DblModel, mor: Mor): Ob | null {
+    try {
+        return model.dom(mor);
+    } catch {
```
epatters: The blanket try-catch in `morDom` here and `morCod` below is smelly to me.

IIRC, the safe way to write this code if you're not sure the morphism is valid/contained in the model (e.g., because the user could have deleted a morphism generator that appears in an equation) is:

```typescript
if (model.hasMor(mor)) {
    return model.dom(mor)
} else {
   return null;
}
```

If this will be a common need, we should put it somewhere else, perhaps as a helper method in `catlog-wasm`.

packages/frontend/src/model/equation_cell_editor.tsx
```diff
@@ -0,0 +1,284 @@
...
+        return null;
+    }
+}
+
+/** Structural equality of objects, sufficient for comparing model objects. */
+function obsEqual(a: Ob | null, b: Ob | null): boolean {
+    if (a === null || b === null) {
+        return a === b;
+    }
+    return JSON.stringify(a) === JSON.stringify(b);
```
epatters: This is even smellier to me :) Do we have to stringify? The `frontend` package depends on a deep equality lib; can we use that instead, and then delete this helper?

packages/frontend/src/model/path_mor_input.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+/** Render a qualified label as dotted display text. */
+function labelText(label: QualifiedLabel | undefined): string {
+    return label && label.length > 0 ? label.join(".") : "";
+}
+
+/** Parse a single label segment, treating all-digit text as an anonymous index. */
+function parseSegment(text: string): LabelSegment {
+    return /^\d+$/.test(text) ? Number.parseInt(text, 10) : text;
+}
```
epatters: The functions `labelText` and `parseSegment` presumably shouldn't be here.

In lieu of the first one, we usually write `label?.join(".") ?? ""` (see almost any analysis), which is kinda annoying to write everywhere but so short that we so far haven't extracted any helper.

packages/frontend/src/model/path_mor_input.tsx
```diff
@@ -0,0 +1,207 @@
...
+/** Build an identity morphism on a basic object generator. */
+function identityMor(obId: Uuid): Mor {
+    return {
+        tag: "Composite",
+        content: { tag: "Id", content: { tag: "Basic", content: obId } },
+    };
+}
+
+/** Parse the object name out of `id(<name>)` text, if it matches. */
+function parseIdentityName(text: string): string | null {
```
epatters, line 191: Hmm, I don't love this but I understand why you're doing it and I don't currently have another suggestion. So no action needed, but I'll mull this over.

epatters (changes requested):
This is awesome. Check out my schemas:

Image

Image

A few small comments below related to the helper functions.

Relatedly, I wonder about the helpers to (de)construct identities and composites: I didn't comment on these specifically but I'm imagining that these will be subsumed by the TS API for models. Is that right?

kasbah:
Thanks, I believe I addressed everything. 

> identities and composites: I didn't comment on these specifically but I'm imagining that these will be subsumed by the TS API for models. Is that right?

I haven't thought about these in that context yet but I probably should.

epatters (approved):

packages/frontend/src/model/path_mor_input.tsx
epatters: My proposed solution is in #1411.

Issue #1177, opened 2026-03-31
Refactor Lotka–Volterra and Linear ODE analyses (for LaTeX equations)


tim-at-topos:

epatters:
> We should possibly consider whether we really need both `ODEResult` and `ODEResultWithEquations`, since all ODE analyses should now come with equations.

All the data types in `catlog-wasm` should be considered disposable glue code, so we should remove anything that we don't actively need.

> Note that this is currently blocked by [#1176](https://github.com/ToposInstitute/CatColab/pull/1176)

Fun fact that I think you'll appreciate: you can mark issues as blocking/blocked by others in the Relationships section on the right side bar.

tim-at-topos:
ah but #1176 is a PR, not an issue, and that makes GitHub unhappy when I try to add it as a blocker sadly

PR #1251, opened 2026-05-07
Refactor ODE analyses to build models in the theory of signed polynomial ODE systems

tim-at-topos:
**GitHub closed this and I cannot reopen it; see #1363**

---
This PR includes a massive redesign of the way that we handle ODE semantics, which has the bonus of introducing some new capabilities (#1177) as well as (hopefully) making adding new ODE semantics in the future considerably simpler.

For those reading (or reviewing) this code, my suggested reading order would be:

1. `ode::polynomial_ode::build_system_custom_parameters` to see how we interpret models of `th_signed_polynomial_ode_system` as systems of polynomial ODEs
2. `ode::lotka_volterra` to see an example of the main chunk of defining an ODE semantics: giving an implementation for `ODESemanticsAnalysis`
3. `ode::ode_semantics` to see where all the actual new code is
4. `ode::mass_action` to see where things are still messy and need some extra finessing
5. The comments that contain the string `FEEDBACK`, where I have some specific questions / highlight specific problems with this design.

p.s. I know this PR adds quite a lot of code for something called a "refactor", but (a) I added a lot of documentation and some new tests, and (b) subsequent PRs focussing on the front-end parts should now be possible, allowing for much more code removal.

---

Subsequent PRs:

- #1319
    N.B. this PR also tidies up some messy things in this current PR.

### DIFF
--- a/packages/frontend/src/util/serialize_document.test.ts
+++ b/packages/frontend/src/util/serialize_document.test.ts
@@ -4,7 +4,7 @@ import { Repo } from "@automerge/automerge-repo";
 import { assert, describe, test } from "vitest";
 
 import { Model, type ModelDocument, Nb } from "catcolab-document-methods";
-import { serializeAutomergeDocument, type Document } from "catcolab-document-types";
+import { serializeAutomergeDocument } from "catcolab-document-types";
 
 // Dummy Automerge repo with no networking or storage.
 const repo = new Repo();
@@ -38,8 +38,8 @@ describe("serializeAutomergeDocument", () => {
         });
 
         const serialized = serializeAutomergeDocument(
-            Automerge.save((docHandle as DocHandle<Document>).doc()),
-        ) as Document;
+            Automerge.save(docHandle.doc()),
+        ) as ModelDocument;
 
         const cell = serialized.notebook.cellContents[cellId];
         assert(cell?.tag === "rich-text");
@@ -85,9 +85,7 @@ describe("serializeAutomergeDocument", () => {
             Automerge.splice(doc, contentPath, 1, 0, "hello world");
         });
 
-        const json = JSON.stringify(
-            serializeAutomergeDocument(Automerge.save((docHandle as DocHandle<Document>).doc())),
-        );
+        const json = JSON.stringify(serializeAutomergeDocument(Automerge.save(docHandle.doc())));
         assert(
             !json.includes("\ufffc"),
             "serialized JSON must not contain U+FFFC placeholders from a flattened Text object",


tim-at-topos:
@epatters Here's a problem: we cannot push the mass-action analysis for *signed* stock-flow diagrams through the theory of polynomial ODEs because we interpret negative links as negative exponents, meaning we'd need *rational* ODEs. For now, since I essentially made up the theory of signed stock-flow diagrams, I've just removed its mass-action analysis. I see two options:

1. Wait until we think about rational ODEs (probably not that far into the future) and then come back to fix this.
2. Implement mass-action as it was before just for the special case of signed stock-flow diagrams.

I'm in favour of 1, but would appreciate your thoughts.

epatters:
Hmm, good question. What would be the code footprint of option (2), i.e., leaving the special-cased ODE analysis code for signed stock-flow as is?

tim-at-topos:
For now it would be reasonably small, essentially just duplicating half of the current `mass_action.rs`. But it would get worse in future PRs, since one goal of this is to refactor a lot of duplicated code between Lotka–Volterra, linear ODE, and mass-action, so mass-action for signed stock-flow diagrams would have entirely unique chunks of code for each part of the analysis (core, wasm, and frontend).

epatters:
> For now it would be reasonably small, essentially just duplicating half of the current `mass_action.rs`. But it would get worse in future PRs, since one goal of this is to refactor a lot of duplicated code between Lotka–Volterra, linear ODE, and mass-action, so mass-action for signed stock-flow diagrams would have entirely unique chunks of code for each part of the analysis (core, wasm, and frontend).

Thanks for explaining. Though I'm not sure how to weigh the tradeoff between keeping existing functionality and maintaining a clean codebase, I tend to value the latter a lot and since you also seem inclined in that direction, I'd say go ahead and remove it.

tim-at-topos:
@epatters and @tslil-topos this isn't quite ready for review yet, but all that's missing is some extra documentation and then some extra refactoring that I think I might move to a future PR because this one is already quite hefty. (Note, however, that the subsequent PR should hopefully remove as many lines as this one adds!)

Would it be possible for one of you to to start taking a look through this when you get a chance please? I think it's very likely that there will be substantial changes requested, and I'd rather put off writing documentation for things that are liable to change 😎 I've included a suggested path through reading the code in the opening comment of this PR!

epatters:
Sounds good. I've marked this for review despite your caveats because that will help me remember it needs a review.

tim-at-topos:
@epatters Something else to consider when thinking about how to encode the parameters (on the type level, as in this PR, or on the data level, as in your example) is how we deal with displaying them. The nice thing about the type level is that we can write

```rust
impl fmt::Display for LotkaVolterraParameter {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match &self {
            Self::Growth { variable } => {
                write!(f, "Growth({})", variable)
            }
            Self::Interaction { link } => {
                write!(f, "Interaction({})", link)
            }
        }
    }
}

impl ToLatex for LotkaVolterraParameter {
    fn to_latex(&self) -> Latex {
        match self {
            Self::Growth { variable } => Latex(format!("g_{{{variable}}}")),
            Self::Interaction { link } => Latex(format!("k_{{{link}}}")),
        }
    }
}
```

where `ToLatex` is a trait that just says you come with a `to_latex()` function. I'm not sure how simple this would be to do in the data layer.

However, I'm still not convinced that the type way is actually the best way... happy to hear honest feedback!

tim-at-topos:
I've drafted https://github.com/ToposInstitute/CatColab/pull/1319/ which shows what a refactor of the LaTeX code could look like, based on this PR

PR #1254, opened 2026-05-11
Update TypeScript to v6, and update oxlint & stylelint

kasbah:
Was interested in updating TS to 6.0 since it is the "make sure you are compatible with TS Go" release and we are using TS Go a little via oxlint-tsgo.

### DIFF
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -75,6 +75,12 @@ export default defineConfig({
         // Ternary-as-statement (`condition ? doA() : doB()`) is idiomatic in
         // this codebase for concise branching in callbacks.
         "no-unused-expressions": ["warn", { allowTernary: true }],
+        // Requires adding explicit `return undefined` in places seems to add too much noise
+        "consistent-return": "off",
+        // We use _ prefix in places
+        "no-underscore-dangle": "off",
+        // We have some tests that just see if anything breaks when things are run
+        "vitest/expect-expect": "off",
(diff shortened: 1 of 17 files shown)


packages/frontend/oxlint.config.ts
```diff
@@ -10,6 +10,9 @@ export default defineConfig({
     jsPlugins: ["eslint-plugin-solid"],
     rules: {
         ...baseConfig.rules,
+        "consistent-return": "off",
```
kasbah: I had a look at keeping this one enabled but we would need to add a lot of `return undefined` in places that don't add clarity IMHO.

https://oxc.rs/docs/guide/usage/linter/rules/typescript/consistent-return.html

epatters: Did you see the `treatUndefinedAsUnspecified` option? Haven't tried, but it sounds like setting it to `true` would avoid the need to add `return undefined`.

epatters (approved):
Thanks for doing this chore, looks good. I only had one question/comment about the option you mentioned.

packages/frontend/oxlint.config.ts
kasbah: It doesn't seem to help, it makes it more strict. The whole rule is about having to add explicit returns which we haven't been doing.

epatters:
@kasbah, due to #1258, the pnpm hash will need to be updated again. Sorry about that. Feel free to merge after updating.

PR #1255, opened 2026-05-11
Evaluating `rush` monorepo build tool for TS

kasbah:
https://rushjs.io/


kasbah:
Not going for this, just some simpler dedupe instead #1290

PR #1257, opened 2026-05-11
RFC-0009: Instances in DoubleTT

KevinDCarlson:
Here is a short RFC on instances as pseudo-terms of types in DoubleTT. See #1246 for draft PR of implementation.

https://branch-instances-rfc--catcolab.netlify.app/rfc/0009


epatters (changes requested):
Thanks, this is a nice start! The idea of seeing the fibration of instances over the category of models as a comprehension category is cool.

Perhaps due to my lack of experience with comprehension categories and type theory, the gap between the categorical semantics and the WIP implementation feels rather large. What's missing for me is a presentation of the type theory.

Some questions I'd like to see answered in the RFC:

- Are new kinds of judgments added? If so, what are they?
- What new rules are added compared to RFC-0002 on DoubleTT?

Having clear presentation of the type theory should help me and others understand the implementation.

PR #1263, opened 2026-05-12
Use pnpm2nix-nzbr

kasbah:
Uses [FliegendeWurst/pnpm2nix-nzbr](https://github.com/FliegendeWurst/pnpm2nix-nzbr) for our pnpm in Nix. This improves the workflow of updating deps (no longer have to manualy create hash mismatch and copy in the hash). 

Additionally I looked at supporting MacOS better by not hardcoding `x86_64-linux`.


packages/frontend/default.nix
```diff
@@ -2,15 +2,67 @@
...
+    "packages/document-methods" = ../document-methods/pnpm-lock.yaml;
+    "packages/backend/pkg" = ../backend/pkg/pnpm-lock.yaml;
+  };
+
+  processedLocks = lib.mapAttrs (_: processLock) lockfilesToProcess;
+
+  patchedLockfiles = lib.mapAttrs (
+    relPath: processed:
+    pkgs.writeText "pnpm-lock-${builtins.replaceStrings [ "/" ] [ "-" ] relPath}.yaml" (
+      builtins.toJSON processed.patchedLockfile
```
jmoggr: We're calling `toJSON` on content written to a .yaml file?

This whole statement is a bit hard to read, but that's possibly a skill issue.

flake.nix
```diff
@@ -194,90 +217,121 @@
...
-          };
+              frontend =
+                (pkgs.callPackage ./packages/frontend/default.nix {
+                  inherit inputs self;
+                  rustToolchain = rustToolchainFor system;
+                  pnpm2nix = inputs.pnpm2nix-nzbr;
+                }).package;
+
+              frontend-tests =
+                (pkgs.callPackage ./packages/frontend/default.nix {
```
jmoggr: This `callPackage` block seems cut and paste from above. Possibly worth factoring out.

jmoggr (approved):
Looks good!

The weirdness with the yaml file containing json should be fixed.

The lockfile alchemy is hella hacky, but it wasn't that hard to follow and let's us ditch the hash update, so I think it's worth it. I'm hoping this code doesn't become a maintenance problem, though I don't think it will.

packages/frontend/default.nix
kasbah: I am not going to claim this is a very clean solution but: this works because YAML is a superset of JSON. This was taken from [what pnpm2nix-nzbr does](https://github.com/FliegendeWurst/pnpm2nix-nzbr/blob/35f88a41d29839b3989f31871263451c8e092cb1/derivation.nix#L186) but adapted for our monorepo setup.

Maybe we just add a comment for now? I can't think of a clean solution.

flake.nix
kasbah: [4b5b014](https://github.com/ToposInstitute/CatColab/pull/1263/commits/4b5b014750b895119af402dd19700707ca744c34)

packages/frontend/default.nix
jmoggr: > YAML is a superset of JSON

TIL (apparently there are caveats to this)

We could try `pkgs.formats.yaml`. Otherwise a comment indicating that JSON in a yaml file is technically correct would be nice, I thought it was just a mistyped extension.

kasbah: I assumed such a convenience didn't exist since ppl are resorting to `toJSON` hacks! It seems to work. [cd67a35](https://github.com/ToposInstitute/CatColab/pull/1263/commits/cd67a355c4ca259441b24d71be6c988f8c69f26d)

jmoggr: I was unable to find any documentation on it besides the source... claude found it after some poking.

PR #1268, opened 2026-05-12
Make optional deps be ordinary deps in AlgJulia interop

kris-brown:
To simplify the Julia interop, we now avoid the package extension feature and dynamic loading of dependencies. 

The downside of this simplicity is that someone who just wants to run `endpoint.jl` with a subset of the dependencies cannot (and therefore may have to precompile for a long time). However, the primary intended usage of this package is for it to be running on a central instance which is only precompiled once.


epatters (approved):
Thanks Kris!

PR #1274, opened 2026-05-13
Remove Project.toml/manifest for AlgJulia interop tests

quffaro:
Tests pass.


kasbah:
AHA! You might be the right person to tell me what's going on with these lock files and why they caused our staging deploy to try and install a bunch of Julia packages into a non writable folder.

EDIT: Wait no, it was trying to write the lock file which is in a read-only dir. That makes sense, and I just saw https://github.com/ToposInstitute/CatColab/pull/1268/ moved the Manifest to `test/` which I am guessing was a mistake.

.gitignore
```diff
@@ -112,3 +112,7 @@ CHANGELOG.html
 
 # VSCode
 .vscode/*
+
+# Julia
+/packages/algjulia-interop/test/Manifest.toml
+/packages/algjulia-interop/test/Project.toml
```
epatters: Rather than putting in these in the top-level `.gitignore`, can you create `packages/algjulia-interop/.gitignore` and add them there?

epatters (changes requested):
Thanks Matt!

epatters:
> EDIT: Wait no, it was trying to write the lock file which is in a read-only dir. That makes sense, and I just saw #1268 moved the Manifest to `test/` which I am guessing was a mistake.

Yeah, that was a mistake. I should have caught that in the review.

packages/algjulia-interop/.gitignore
```diff
@@ -22,3 +22,7 @@ docs/site/
 
 # Local preferences
 LocalPreferences.toml
+
+# Julia
+/packages/algjulia-interop/test/Manifest.toml
+/packages/algjulia-interop/test/Project.toml
```
epatters: Please use relative rather than absolute paths.

epatters (changes requested):

epatters (approved):

PR #1186, opened 2026-04-02
Keyboard shortcuts for undo/redo

kasbah:

PR #1276, opened 2026-05-15
Refactor and fix focus management

### LINKED ISSUE #159: Focus seems wrong when opening analysis tab
The analysis tab claims you can add analyses using shift+enter but at least on Safari it seems impossible to get the focus of shift+enter outside the model notebook.

### LINKED ISSUE #161: Principled approach to keyboard focus management

### LINKED ISSUE #1256: Keyboard navigation of cells interrupted by completion popup
Since #1243, the up/down arrow keys can't reliably be used to navigate the notebook: when a field that allows completions is hit (e.g., a specialization in an instantiation), the completions pop up and grab control of the arrow keys.

kasbah:
This is largely a refactor of focus and keyboard shortcut management (closes #161). It also:

- Implements shortcuts for undo/redo (supersedes #1186)
- Fixes #1256 as best we can without changing completions to not to use up/down: when doing keyboard navigation you might sometimes have to `Esc` out of a completion popup. 
- Fixes https://github.com/ToposInstitute/CatColab/issues/159 
- Adds focus indication to the side-bar:
image

### DIFF
--- a/packages/frontend/src/analysis/analysis_editor.tsx
+++ b/packages/frontend/src/analysis/analysis_editor.tsx
@@ -3,6 +3,7 @@ import { Dynamic } from "solid-js/web";
 import invariant from "tiny-invariant";
 
 import { Nb } from "catcolab-document-methods";
+import { type FocusHandle } from "catcolab-ui-components";
 import { type CellConstructor, type FormalCellEditorProps, NotebookEditor } from "../notebook";
 import type { AnalysisMeta, DiagramAnalysisMeta, ModelAnalysisMeta } from "../theory";
 import { LiveAnalysisContext } from "./context";
@@ -16,7 +17,10 @@ import type { Analysis } from "./types";
 
 /** Notebook editor for analyses of models of double theories.
  */
-export function AnalysisNotebookEditor(props: { liveAnalysis: LiveAnalysisDoc }) {
+export function AnalysisNotebookEditor(props: {
+    liveAnalysis: LiveAnalysisDoc;
+    focus: FocusHandle;
+}) {
     const liveDoc = () => props.liveAnalysis.liveDoc;
 
     const cellConstructors = () => {
@@ -39,7 +43,7 @@ export function AnalysisNotebookEditor(props: { liveAnalysis: LiveAnalysisDoc })
                 changeNotebook={(f) => liveDoc().changeDoc((doc) => f(doc.notebook))}
                 formalCellEditor={AnalysisCellEditor}
                 cellConstructors={cellConstructors()}
-                noShortcuts={true}
+                focus={props.focus}
             />
         </LiveAnalysisContext.Provider>
     );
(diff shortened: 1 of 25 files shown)


packages/frontend/src/analysis/analysis_editor.tsx
```diff
@@ -39,7 +39,6 @@ export function AnalysisNotebookEditor(props: { liveAnalysis: LiveAnalysisDoc })
                 changeNotebook={(f) => liveDoc().changeDoc((doc) => f(doc.notebook))}
                 formalCellEditor={AnalysisCellEditor}
                 cellConstructors={cellConstructors()}
-                noShortcuts={true}
```
epatters, line 42: Exciting to finally be rid of this hack!

epatters (commented):
Thanks for undertaking this refactor. Focus management has been a longstanding problem, and I'm glad we're thinking through a design for it instead of piling on hacks.

There may well be subtleties that I'm overlooking, but the present design is more centralized and specific to notebooks that I would have expected. As the comments in `type.ts` explain, the application-level focus state says not only which notebook is focused, but which cell and even which popover in the notebook is focused. I have several worries about this design:

1. It is specialized to notebooks, but we will have documents that are not notebooks
2. It leaks information deep in the component hierarchy up to the application top-level
3. It conflates data and layout: focus is attached to notebooks by ID, rather than to panes

To focus the mind, consider the concrete scenario of integrating a Petri net editor into CatColab (cf. #1279). Such an editor will need to get and lose focus coherently with the rest of the app.

I'm trying to imagine a more compositional/hierarchical approach, where a component tracks which of its *immediate* children has focus. We're already kind of doing this in the notebook and object/morphism editors, e.g., a notebook keeps track of which cell is focused. I propose doing the same thing one level up: the document page should keep track of which of its 1-2 panes has focus. In this approach, focus changes bubble up the component hierarchy, e.g., when a name field in a morphism editor gets focus, it notifies its cell that it has focus, which notifies its notebook, which notifies the document page. In this way, a component pushes its focus events upward.

To handle keyboard shortcuts, components must also be able to "pull" the latest focus state, since another component may have grabbed it. To do so, one must walk up the component hierarchy again, asking if each parent has focus. I would imagine this pulling and pushing could be handled by passing a getter and a setter for focus state.

What do you think?

kasbah:
I've introduced a `FocusHandle` and a `useChildFocus` that handles the recursion into parent components. I believe this implements your idea as described and seems to mesh better with the rest of the code than my previous design.

packages/frontend/src/model/morphism_cell_editor.tsx
```diff
@@ -16,14 +16,8 @@ export default function MorphismCellEditor(props: MorphismEditorProps) {
...
     invariant(liveModel, "Live model should be provided as context");
 
-    const [activeInput, setActiveInput] = createSignal<MorphismCellInput>("name");
-
-    // Reset to default on deactivation so re-entry lands on the name input.
-    createEffect(() => {
-        if (!props.isActive) {
-            setActiveInput("name");
-        }
-    });
```
epatters, line 26: It's great that the new design avoids having to handle this tricky logic across a bunch of individual components.

epatters (approved):
Thanks Kaspar! The way you've set up the new `FocusHandle` abstraction is quite elegant IMO.

PR #1277, opened 2026-05-15
Reset active input in editors when focus is lost

kasbah:
Make keyboard navigation more predictable. The focused input no longer depends on the last input that was previously focused in the cell.


epatters (approved):
Thanks! I've found the current behavior annoying/counterintuitve too.

PR #1281, opened 2026-05-15
Update wasm-pack to 0.15 and remove patch

kasbah:
Fixes the upstream issue we needed the patch for.


epatters (approved):
Glad to see this fixed!

PR #1284, opened 2026-05-18
Modelica export

kasbah:
WIP: Was interested to see how feasible a Modelica export is

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 4
 
 [[package]]
+name = "addr2line"
+version = "0.25.1"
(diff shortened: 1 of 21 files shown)


tim-at-topos:
This looks interesting! I think it makes sense that this should be something downstream of the mass-action semantics (or any other ODE semantics)analysis though, right? Since it wants to use values like `duration` and `initial_values`.

After https://github.com/ToposInstitute/CatColab/pull/1251 I'd quite like to take a look at this because it should follow quite nicely on from that.

tim-at-topos:
@kasbah should we take a quick look at this sometime this coming week? I'd like to!

PR #1285, opened 2026-05-19
Code view component

quffaro:
Replaces #1252.

### DIFF
--- a/packages/frontend/src/stdlib/analyses/sql.tsx
+++ b/packages/frontend/src/stdlib/analyses/sql.tsx
@@ -4,7 +4,7 @@ import Copy from "lucide-solid/icons/copy";
 import Download from "lucide-solid/icons/download";
 import { For, Match, Show, Switch } from "solid-js";
 
-import { BlockTitle, ErrorAlert, IconButton } from "catcolab-ui-components";
+import { CodeView, BlockTitle, ErrorAlert, IconButton } from "catcolab-ui-components";
 import type { ModelAnalysisProps } from "../../analysis";
 import * as SQL from "./sql_types.ts";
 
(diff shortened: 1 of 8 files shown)


packages/ui-components/src/code_view.tsx
```diff
@@ -0,0 +1,23 @@
...
+    const [html] = createResource(() =>
+        codeToHtml(props.text, {
+            lang: props.language,
+            theme: "min-light",
+        }),
+    );
+
+    // oxlint-disable-next-line solid/no-innerhtml
+    return <div innerHTML={html()} />;
+};
```
kasbah: Is the html we get from shiki sanitized?

packages/frontend/src/stdlib/analyses/sql.tsx
```diff
@@ -105,7 +105,9 @@ export default function SQLSchemaAnalysis(
                                         actions={SQLHeader(sql())}
                                         settingsPane={BackendConfig()}
                                     />
-                                    <pre>{sql()}</pre>
+                                    <Show when={props.content.backend} keyed>
+                                        {(_) => <CodeView language={"sql"} text={sql()} />}
```
kasbah:
```suggestion
                                        {(_) => <CodeView language="sql" text={sql()} />}
```

(thought our auto-formatter would handle things like this)

kasbah (commented):
Thanks Mat, looks much cleaner than the Prismjs version!

packages/ui-components/src/code_view.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const CodeView = (props: CodeViewOptions) => {
+    const [html] = createResource(() =>
+        codeToHtml(props.text, {
+            lang: props.language,
+            theme: "min-light",
+        }),
+    );
+
+    // oxlint-disable-next-line solid/no-innerhtml
```
kasbah:
```suggestion
    // oxlint-disable-next-line solid/no-innerhtml -- shiki uses hast-util-to-html which escapes html entitites so no need for extra sanitization and setting innerHTML should be safe
```

epatters (approved):
Thanks both! Looks nice and clean.

PR #1286, opened 2026-05-20
Clean up `createResource` reactivity

kasbah:
I was a bit suspicious of these lint exceptions so dug a bit deeper and cleaned them up.


epatters (approved):
Thanks for cleaning this up!

PR #1288, opened 2026-05-20
Upgrade ECharts from v5.5 to v6.1

epatters:
ECharts v6 has been out for nearly a year. The recent release of v6.1.0 was a good reminder that we should upgrade.

However, `echarts-solid` appears to be unmaintained (see https://github.com/alxnddr/echarts-solid/pull/2), so I've switched the dependency to our fork.


pnpm-workspace.yaml
```diff
@@ -2,6 +2,9 @@
 minimumReleaseAge: 12960
 minimumReleaseAgeExclude:
   - wasm-pack
+  - echarts
```
epatters, line 5: Like `wasm-pack` above, this addition is temporary and we should remember to remove it later.

pnpm-workspace.yaml
```diff
@@ -2,6 +2,9 @@
...
+onlyBuiltDependencies:
+  - echarts-solid
```
epatters: This is only necessary because we're now depending on a Git repo instead of a pre-built package from npm.

epatters:
@kasbah, given the suspicious hacks that Claude invented to make the dependency on a Git repo work (see commit below) and how tiny the dependency `echarts-solid` actually is, we might be better off just copying it directly into our monorepo. I'd appreciate your help finding a clean way to proceed.

https://github.com/ToposInstitute/echarts-solid/commit/7191b18a556747b8f1df0d9bd002ecf9b3dacf30

(This does work for me locally, even if the CI is currently failing.)

kasbah:
I pushed a commit that switches to my fork (I don't have write permissions on Topos' fork) where I built the package and committed the `dist/` folder. That seems to work without extra hacks so that's an option.

I guess including the package into our monorepo would be a little cleaner but a bit more work, plus a little more annoying to switch back if the previous maintainer does decide to maintain echarts-solid again (though that doesn't seem that likely given the lack of commits for 2 years). 

Let me know how you want to proceed.

epatters:
Thanks! Your proposed solution (committing the `dist` folder) seems best. Let's go with that.

We should use the [fork](https://github.com/ToposInstitute/echarts-solid) in the Topos org though. I've given you permissions on that repo. Would be great if you could push your changes to that repo and then update this PR accordingly.

kasbah (approved):

kasbah:
Done

PR #1289, opened 2026-05-21
Colors: stop using green, alias blue to "indicator"

kasbah:
I always felt the green clashes with the teal so this cuts it out replacing it with "Topos primary" teal. 

image

↓

image

image
image

---

I also renamed the blue we use for indication to `--color-indicator`

image
image

---

These two colors are aliased for our "question" and "note" alerts replacing a similar teal we were using previously.

image


epatters (approved):
Thanks, I like it!

PR #1290, opened 2026-05-22
Add a `vite-plugin-monorepo-dedupe` tool

kasbah:
This replaces our existing dedupe with a more thorough plugin that recursively goes through all deps of local packages. We can then drop this into any vite.config of any new monorepo package.

Rush actually has some [good docs on why this is needed](https://rushjs.io/pages/advanced/npm_doppelgangers/). I am opting for this solution instead of a more comprehensive monorepo tool like Rush though so this supersedes #1255.

I also opted to put this into a new `tools/` dir rather than `packages/` just to make clear that this isn't really anything to do with CatColab, just a dev tool we use. I gave it a `@catcolab-dev-tools/` package name prefix. I have some work on another branch that will also use the `tools/` dir so it won't be the only package there.

### DIFF
--- a/packages/frontend/default.nix
+++ b/packages/frontend/default.nix
@@ -40,6 +40,7 @@ let
     "packages/ui-components" = ../ui-components/pnpm-lock.yaml;
     "packages/document-methods" = ../document-methods/pnpm-lock.yaml;
     "packages/backend/pkg" = ../backend/pkg/pnpm-lock.yaml;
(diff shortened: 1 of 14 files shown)


epatters (approved):
Thanks Kaspar!

- While I haven't reviewed the code in detail, it seems a clear improvement to consolidate the deduplication functionality in one place instead of copy-pasting it around.
- Again, while I have no experience with tools like Rush, I'm happy enough to avoid introducing another heavy opinionated tool on top of Nix into our infrastructure

PR #1294, opened 2026-05-26
Run Julia CI conditionally

### LINKED ISSUE #1293: Conditionally run Julia in CI workflow
Testing Julia code in CI is always slow, as can be seen since #1292. We should cache the results of Julia tests, like we do for other build actions, so that we can avoid running them when possible.

quffaro:


quffaro:
The first image shows that the Julia step skips when no change was made to any file in the subdirectory. The second image shows the Julia step running because the `algjulia/README.md` was changed.

image
image

.github/workflows/ci.yml
```diff
@@ -20,13 +20,26 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check if packages/algjulia-interop changed
+        id: changes
+        run: |
```
kasbah, line 28: I don't think this is right because on a PR it didn't necessarily change on the last commit even if there are algjulia-interop changes on the PR.

kasbah (changes requested):
Are you going to add caching too?

.github/workflows/ci.yml
kasbah: See https://github.com/ToposInstitute/CatColab/pull/1295

quffaro: I think your suggestion is cleaner.

quffaro: On the issue #1293, Evan and I decided to not cache this time around. I should have changed the branch name.

kasbah: I guess I'll close this then and take that one out of draft.

kasbah:
Superseded by https://github.com/ToposInstitute/CatColab/pull/1295

PR #1296, opened 2026-05-27
Move ui-component tests to own GitHub Actions workflow

kasbah:
So that they don't run on unrelated PRs. Also fixes an issue with installing chromium for these tests affecting CI runs on all PRs currently.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,31 +101,6 @@ jobs:
         run: |
           nix build .#rust-docs-check
 
-  ui_components_tests:
-    name: ui-components tests
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-
-      - name: Setup pnpm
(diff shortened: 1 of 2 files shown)


jmoggr (approved):
Possible future footgun: in the future something in the `ui-components` package or the CI tests for it could become dependent on something in the root `package.json`, then the root package.json could change without the tests re-running, possibly causing a false positive on tests passing. I'm somewhat inclined to adding the root package.json now to head off that risk, however the risk is really quite small so idk.

kasbah:
Thanks, I think that was mostly covered by the root pnpm-lock already being included but I added the package.json for good measure, I don't think there are serious downsides to being explicit like that.

jmoggr:
> Thanks, I think that was mostly covered by the root pnpm-lock already being included but I added the package.json for good measure, I don't think there are serious downsides to being explicit like that.

The only example I could think of was a dependency on scripts in the root package.json, since changing that wouldn't impact the lockfile.

PR #1303, opened 2026-05-29
MutDblModelMapping trait for discrete doctrine

quffaro:


packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -21,13 +21,48 @@
...
+    type MorGen;
+
+    /// Constructs a model mapping from a pair of hash maps.
+    fn new(
+        ob_pairs: impl IntoIterator<Item = (QualifiedName, Self::ObGen)>,
+        mor_pairs: impl IntoIterator<Item = (QualifiedName, Self::MorGen)>,
+    ) -> Self;
+
+    /// Assigns an object generator, returning the previous assignment.
+    fn assign_ob(&mut self, x: QualifiedName, y: Self::ObGen) -> Option<Self::ObGen>;
```
epatters, line 54: The types of `y` and the return type are not general enough.

packages/catlog/src/dbl/model_morphism.rs
```diff
@@ -21,13 +21,48 @@
...
+
+    /// Assigns a morphism generator, returning the previous assignment.
+    fn assign_mor(&mut self, e: QualifiedName, n: Self::MorGen) -> Option<Self::MorGen>;
```
epatters, line 57: Similar comment about `n` and the return type.

epatters (changes requested):
Thanks for starting this, Matt.

On looking at this further, I realized that we already have an `FgCategoryMap` trait to build on, analogous to how `(Fp)DblModel` extends `FgCategory`. We should start by having `DiscreteDblModelMap` implement that trait.

We could then have add a trait `MutDblModelMap` extending `FgCategoryMap`, analogous to how `MutDblModel` extends `FgCategory` (among other traits). For that you could reuse some of what you have below (but see my comments).

quffaro:
After discussion with @epatters, we have decided to punt on this for now.

PR #1304, opened 2026-05-29
Morphisms of Modal Double Models

### LINKED ISSUE #1282: Large difference in size of `DblModelDiagramBox` variants
Image

I'm marking this as a task because it's not blocking or causing a feature to break, but ignoring this error is probably nose-blindness to code smell.

quffaro:
Dependent on #1303


quffaro:
After discussion with @epatters, we have decided to punt on this for now.

PR #1305, opened 2026-05-30
Frontend for Diagrams of Modal Double Models

quffaro:
Dependent on #1304


packages/frontend/src/diagram/morphism_cell_editor.tsx
```diff
@@ -45,16 +45,16 @@ export function DiagramMorphismCellEditor(props: {
...
+            <ObInput
                 placeholder="..."
                 ob={props.decl.dom}
                 setOb={(ob) => {
                     props.modifyDecl((decl) => {
                         decl.dom = ob;
                     });
                 }}
-                obType={domType()}
                 generateId={v7}
```
quffaro: Build breaks at this point since this field does not exist. However, this is needed for anonymous variables to work.

PR #1309, opened 2026-06-03
List picker component for morphisms of modal diagrams

quffaro:
This is the list picker cherry-picked from #1305

PR #1316, opened 2026-06-08
List picker component for morphisms of modal morphisms

quffaro:
This is the list picker cherry-picked from #1305. This is a new PR for a branch renamed from the branch in #1309. This component uses the generic component in #1321 for morphisms of modal models. This PR also introduces the theory of the DEC, which used for testing the frontend feature.

image


quffaro:
I've rebased onto #1326 as well

epatters:
Thanks, #1326 has also been merged now.

kasbah (commented):
I can't speak too much on the theory changes so would good if you could have a quick look @epatters. 

I was going to comment that the ObListEditor is convoluted but it's pretty much copied from the existing models ObListEditor so I can't really complain. RFC-0006 should really help refactoring this in the future.

PR #1319, opened 2026-06-15
Refactor LaTeX methods in catlog-wasm and catlog

### LINKED ISSUE #1171: Stop using `fmt::Display` for LaTeX output
We should continue to use `fmt::Display` but only for unicode-style outputs, not for anything LaTeX. We should likely consider a new trait (`DisplayLaTeX`, or `DisplayRich`, or something like this)  and start using this.

### LINKED ISSUE #1177: Refactor Lotka–Volterra and Linear ODE analyses (for LaTeX equations)

### LINKED ISSUE #1215: Polynomial ODE equations and simulation both show unassigned variables in compositional models
If another polynomial ODE model is instantiated but not all of its variables are assigned to ones in the larger notebook, all the unassigned variables of the inner model still show in both the equations and simulation analyses.

The behaviour is slightly different depending on whether or not the unassigned variable has been explicitly mentioned.

Image

Image

tim-at-topos:
_Stacked on #1251_

Now that more ODE analyses have LaTeX equations, we should be stricter about the use of `fmt::Display` only for debugging.

This PR introduces some new types and traits (`Latex`, `ToLatex`, `ToLatexWithMap`) for dealing with LaTeX expressions and rendering object/morphism names nicely in the frontend. It also includes a bunch of tests for ODE semantics in the wasm layer. Finally, as a bonus, it removes a lot of repetitive code from `catlog-wasm::src::analyses` by defining `ode_semantics_simulation` and `ode_semantics_equations`, enabled by #1251, as well as removing some unnecessary types introduced there (such as `LotkaVolterraEquationsData`, which is just `null`).

Again, this refactor adds a lot of lines, but that's almost entirely due to how many nice new tests it adds 👼 

As a bonus, this PR fixes the following issues:
- #1171
- #1215

---

Although further refactoring is possible (e.g. `lotka_volterra.tsx` and `linear_ode.tsx` share a lot of code), I think that the "correct" solution to this would be to start looking at generating that code from the definitions of the analyses rather than manually refactoring it. But this is now a much larger, rather separate project,  so I think it's a bit out of scope.

With that in mind, **I would suggest that this PR finally completes the last item in #1177 , and thus closes the RFC-0001 parent issue #1098.**


tim-at-topos:
@epatters Although this PR is separate from #1251 (it deals with LaTeX things), it does simplify (i.e. reduce) the code of #1251. If it makes things easier for me to merge this in so there's only a single PR then let me know and I'll fix all these merge conflicts and do so 👍

PR #1326, opened 2026-06-24
List editor improvements

kasbah:
- Make list editor show an empty input at end

- Make clicking anywhere remove focus from list editor

- Move to next list editor item when clicking completion
- More tweaks

https://github.com/user-attachments/assets/8a17d142-101b-47d3-922c-c7c5cdb74e22


epatters (approved):
Thanks for all these improvements, looks great!

PR #1332, opened 2026-06-30
OBEE-6 Document type for (tabular) instances

tim-at-topos:
To-do:
- [x] write a backend integration test

### DIFF
--- a/packages/document-types/src/v2/document.rs
+++ b/packages/document-types/src/v2/document.rs
@@ -1,6 +1,7 @@
+use crate::current::instance::Table;
 use crate::v0::AnalysisType;
 use crate::v1;
-pub use crate::v1::DocumentType;
+use std::{collections::HashMap, str::FromStr};
 
 use super::analysis::Analysis;
 use super::api::Link;
@@ -9,6 +10,7 @@ use super::notebook::Notebook;
 
 use serde::{Deserialize, Serialize};
 use tsify::Tsify;
+use uuid::Uuid;
 
 /// This is the content of a model document. For legacy reasons, we reserve
 /// the name "ModelDocument" for `Document & { type: "model" }`.
@@ -37,6 +39,16 @@ pub struct DiagramDocumentContent {
     pub version: String,
 }
 
+#[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct InstanceDocumentContent {
+    pub name: String,
+    #[serde(rename = "instanceOf")]
+    pub instance_of: Link,
+    pub tables: HashMap<Uuid, Table>,
+    pub version: String,
+}
+
 #[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Tsify)]
 #[tsify(into_wasm_abi, from_wasm_abi)]
 pub struct AnalysisDocumentContent {
@@ -49,7 +61,7 @@ pub struct AnalysisDocumentContent {
     pub version: String,
 }
 
-#[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Tsify)]
+#[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
 #[serde(tag = "type")]
 #[tsify(into_wasm_abi, from_wasm_abi)]
 pub enum Document {
@@ -59,8 +71,50 @@ pub enum Document {
     Diagram(DiagramDocumentContent),
     #[serde(rename = "analysis")]
     Analysis(AnalysisDocumentContent),
+    #[serde(rename = "instance")]
+    Instance(InstanceDocumentContent),
     #[serde(rename = "llmconversation")]
     LLMConversation(LLMConversationDocumentContent),
+}
+
+/// The type/kind of a document, without any associated content.
+#[derive(PartialEq, Eq, Debug, Clone, Copy, Serialize, Deserialize, Tsify)]
+#[cfg_attr(
+    feature = "backend",
+    derive(autosurgeon::Reconcile, autosurgeon::Hydrate, ts_rs::TS)
+)]
+#[cfg_attr(
+    feature = "backend",
+    ts(export_to = "user_state.ts", rename_all = "lowercase")
+)]
+#[serde(rename_all = "lowercase")]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub enum DocumentType {
+    #[cfg_attr(feature = "backend", autosurgeon(rename = "model"))]
+    Model,
+    #[cfg_attr(feature = "backend", autosurgeon(rename = "diagram"))]
+    Diagram,
+    #[cfg_attr(feature = "backend", autosurgeon(rename = "instance"))]
+    Instance,
+    #[cfg_attr(feature = "backend", autosurgeon(rename = "analysis"))]
+    Analysis,
+    #[cfg_attr(feature = "backend", autosurgeon(rename = "llmconversation"))]
+    LLMConversation,
+}
+
+impl FromStr for DocumentType {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "model" => Ok(DocumentType::Model),
+            "diagram" => Ok(DocumentType::Diagram),
+            "instance" => Ok(DocumentType::Instance),
+            "analysis" => Ok(DocumentType::Analysis),
+            "llmconversation" => Ok(DocumentType::LLMConversation),
+            other => Err(format!("unknown document type: {other}")),
+        }
+    }
 }
 
 impl Document {
--- a/packages/document-types/src/v2/document.rs
+++ b/packages/document-types/src/v2/document.rs
@@ -40,7 +40,7 @@ pub struct DiagramDocumentContent {
 }
 
 #[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
-#[tsify(into_wasm_abi, from_wasm_abi)]
+#[tsify(into_wasm_abi, from_wasm_abi, hashmap_as_object)]
 pub struct InstanceDocumentContent {
     pub name: String,
     #[serde(rename = "instanceOf")]
--- a/packages/document-types/src/v2/instance.rs
+++ b/packages/document-types/src/v2/instance.rs
@@ -1,0 +1,43 @@
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use tsify::Tsify;
+use uuid::Uuid;
+
+/// The value of a single "cell" (i.e. field) in a table row. If the column corresponds to an
+/// attribute morphism then we provide the value of the type; if the column corresponds to a
+/// mapping morphism then we provide the uuid of the row.
+#[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+pub enum FieldValue {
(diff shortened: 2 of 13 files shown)


packages/document-types/src/v2/document.rs
```diff
@@ -36,6 +36,16 @@ pub struct DiagramDocumentContent {
     pub version: String,
 }
 
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct InstanceDocumentContent {
+    pub name: String,
+    #[serde(rename = "instanceIn")]
+    pub instance_in: Link,
+    pub notebook: Notebook<super::instance_judgment::InstanceJudgment>,
```
kasbah: Why are we pretending this is a notebook? Beyond confusing names we are adding unnecessary data by doing this.

I believe this should just be something like `pub tables: Vec<Table>` and we should get rid of the judgment until Evan tells us what we need judgments for.

kasbah:
I pushed a commit removing the fake instance editor from frontend.

packages/document-types/src/v2/instance.rs
```diff
@@ -0,0 +1,35 @@
...
+#[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+pub enum CellValue {
+    // If the column corresponds to an attribute morphism then we provide the value of the type.
+    Null,
+    Bool(bool),
+    Int(i32),
+    Float(f32),
+    String(String),
+    // If the column corresponds to a mapping morphism then we provide the uuid of the entity.
+    EntityRef(Uuid),
```
kasbah: This needs to hold the uuid of a TableRow and should probably be called something like RowRef.

packages/document-types/src/v2/instance.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+#[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+pub struct Table {
+    // The uuid of the table.
+    id: Uuid,
+    // The uuid of the entity to which this table corresponds.
+    entity: Uuid,
+    // The rows of the table.
+    rows: Vec<TableRow>,
```
kasbah: All the fields need to be pub to actually use them.

packages/document-types/src/v2/document.rs
tim-at-topos: The problem with `pub tables: Vec<Table>` is that somewhere along the line a trait will require that we implement `Eq`, and we cannot derive this once we have `Float(f32)` as a possible `CellValue`...

kasbah: What trait requires `Eq`? I think you would have run into this issue if you actually put your `Table` into the `Notebook` too (via `InstanceJudgment` if you really want). It seems unrelated to the field name and its type. At some point a f32 has to go into the document so we cannot implement `Eq` for tabular instances.

tim-at-topos: `pub enum Document` currently derives `Eq`, so `InstanceDocumentContent` also requires it, and thus so too does `tables: Vec<Table>`

kasbah: Oh this may be my fault. I may have added it for property based testing.

kasbah: It really doesn't seem to need `Eq`. https://github.com/ToposInstitute/CatColab/pull/1342

tim-at-topos: cool, free free to merge #1342 into this PR :-)

kasbah: It was as easy as 1342!

packages/document-types/src/v2/instance.rs
```diff
@@ -0,0 +1,35 @@
...
+    // column and the value.
+    pub content: HashMap<Uuid, CellValue>,
+}
+
+#[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+pub struct Table {
+    // The uuid of the table.
+    pub id: Uuid,
+    // The uuid of the entity to which this table corresponds.
+    pub entity: Uuid,
```
kasbah: I think if this is an entity then we really shouldn't call a row an entity even just in comments. 😬

packages/document-types/src/v2/document.rs
```diff
@@ -36,6 +37,16 @@ pub struct DiagramDocumentContent {
     pub version: String,
 }
 
+#[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct InstanceDocumentContent {
+    pub name: String,
+    #[serde(rename = "instanceIn")]
+    pub instance_in: Link,
```
epatters: "Instance in" is not really idiomatic. This should be "instance of" or "instance over". I have a slight preference for the former, though I've seen both in use.

packages/ui-components/src/document_type_icon.tsx
```diff
@@ -2,11 +2,12 @@ import ChartSpline from "lucide-solid/icons/chart-spline";
 import File from "lucide-solid/icons/file";
 import FileX from "lucide-solid/icons/file-x";
 import Network from "lucide-solid/icons/network";
+import Table from "lucide-solid/icons/table";
 import { Match, Switch } from "solid-js";
 
 import { ModelFileIcon } from "./model_file_icon";
 
-export type DocumentType = "model" | "diagram" | "analysis";
+export type DocumentType = "model" | "diagram" | "analysis" | "instance";
```
epatters: We've discussed this [previously](https://github.com/ToposInstitute/CatColab/pull/865#discussion_r2593125503) in #865, and while it's not the biggest deal, I continue to feel that we shouldn't redefine `DocumentType` here.

The argument against re-using the type in #865 was that it would require `ui-components` to depend on `catlog-wasm`. Today it would only have to depend on `document-types`, but IMO what this really points to is that `DocumentTypeIcon` is not a generic UI component but is inherently tightly coupled to CatColab's ontology and thus should be moved from `ui-components` to `frontend`.

packages/document-types/src/v2/document.rs
```diff
@@ -58,6 +69,45 @@ pub enum Document {
...
+    feature = "backend",
+    derive(autosurgeon::Reconcile, autosurgeon::Hydrate, ts_rs::TS)
+)]
+#[cfg_attr(
+    feature = "backend",
+    ts(export_to = "user_state.ts", rename_all = "lowercase")
+)]
+#[serde(rename_all = "lowercase")]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub enum DocumentType {
```
epatters, line 92: Not an issue with this PR, but while I'm on the topic of `DocumentType` duplication, I noticed that we have [this typedef](https://github.com/ToposInstitute/CatColab/blob/bcaa311d0bc08a500a7f6cd34cf5bb52daccbd84/packages/frontend/src/api/document.ts#L51) too in `api/document.ts`. I think we can just delete that and replace it with an import of this type?

packages/document-types/src/v2/instance.rs
```diff
@@ -0,0 +1,34 @@
...
+    Int(i32),
+    Float(f32),
+    String(String),
+    // If the column corresponds to a mapping morphism then we provide the uuid of the row.
+    RowRef(Uuid),
+}
+
+#[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+pub struct TableRow {
+    // The row "number".
```
epatters: Upgrade all these comments (`//`) to doc comments (`///`).

In fact, our linter should be flagging the absence of doc comments on public fields, but perhaps it's not running against `document-types`. We should look into that.

packages/document-types/src/v2/instance.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub id: Uuid,
+    // The content of the row, i.e. the column and the value.
+    pub content: HashMap<Uuid, CellValue>,
```
epatters: I'd suggest giving this field a more descriptive name like `fields` or `cells`. (I kind of like `fields` since it's useful to think of a row as a struct/record, but `cells` goes better with `CellValue` assuming we keep that name.)

packages/document-types/src/v2/document.rs
```diff
@@ -36,6 +37,16 @@ pub struct DiagramDocumentContent {
     pub version: String,
 }
 
+#[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct InstanceDocumentContent {
+    pub name: String,
+    #[serde(rename = "instanceIn")]
+    pub instance_in: Link,
+    pub tables: Vec<Table>,
```
epatters: Having a `Notebook` consist of `Vec<NotebookCell>` turned out to be inconvenient because Automerge doesn't have "move"/"reorder" as an operation: you have to delete and then add back, which is bad for version control.

So if the order of the tables should not be persisted, then this should be `tables: HashMap<Uuid, Table>` from entity types to tables, while if it should be persisted, then we should have `tables: HashMap<...>` plus `table_order: Vec<Uuid>`.

packages/document-types/src/v2/instance.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+#[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+pub struct Table {
+    // The uuid of the table.
+    pub id: Uuid,
+    // The uuid of the entity to which this table corresponds.
+    pub entity: Uuid,
```
epatters: Confused about why there are two UUIDs here: is a table not uniquely identified by the entity type to which it corresponds?

packages/document-types/src/v2/instance.rs
```diff
@@ -0,0 +1,34 @@
...
+    // The rows of the table.
+    pub rows: Vec<TableRow>,
```
epatters: Same comment about `rows` as about `tables`: we should have `rows: HashMap<Uuid, TableRow`> and `row_order: Vec<Uuid>`.

epatters (changes requested):
Thanks, this is shaping up! Some comments, mostly minor, below.

packages/document-types/src/v2/instance.rs
tim-at-topos: enabling the linting for `document-types` gives a *lot* of warnings... I'll see how many of them are just from this PR

packages/ui-components/src/document_type_icon.tsx
tim-at-topos: gonna ping @kasbah about this 😉

packages/document-types/src/v2/instance.rs
epatters: Yeah, probably we should enable that linting in a future PR because it will require adding new doc comments all over the place.

epatters (approved):
This is looking good!

I will approve as is and leave it for @kasbah to decide, when he gets back, whether to address the document type duplication as part of this PR or in the future. In the latter case I would prefer that we create an issue to track this so we don't have the same conversation every time we add a new document type.

packages/ui-components/src/document_type_icon.tsx
kasbah: https://github.com/ToposInstitute/CatColab/issues/1369

tim-at-topos:
We've opened #1369 and #1370 to track these two reduplications.

PR #1333, opened 2026-07-07
Protect against some previewing errors I noticed this morning

KevinDCarlson:
The problem this was responsive to is that it was possible to get a Netlify preview up that was totally incapable of opening a model, if you have a stale session from the same host, due to some interactions with Firebase attempting to do some auth that it can't do on a non-whitelisted host. It's a bit hard to repro the bug situation itself because it happens entirely client-side, but the changes here are pretty chill. The time-outs should never happen on the happy path.

### DIFF
--- a/packages/frontend/src/api/rpc.ts
+++ b/packages/frontend/src/api/rpc.ts
@@ -8,6 +8,26 @@ import type { QubitServer, RpcResult } from "catcolab-api";
 /** RPC client for communicating with the CatColab backend. */
 export type RpcClient = QubitServer;
 
+/** Resolve after `ms` milliseconds. Used to bound an await that could otherwise hang. */
+const delay = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms));
+
+/** Reject if `promise` doesn't settle within `ms` milliseconds. */
+function withTimeout<T>(promise: Promise<T>, ms: number): Promise<T> {
+    return new Promise<T>((resolve, reject) => {
+        const timer = setTimeout(() => reject(new Error(`Timed out after ${ms}ms`)), ms);
+        promise.then(
+            (value) => {
+                clearTimeout(timer);
+                resolve(value);
+            },
+            (err) => {
+                clearTimeout(timer);
+                reject(err);
+            },
+        );
+    });
+}
+
 /** Create a fetch function that automatically attaches Firebase auth tokens. */
 export function createFetchWithAuth(firebaseApp?: FirebaseApp): typeof fetch {
     let currentUser: User | null = null;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/api/rpc.ts
```diff
@@ -23,15 +43,21 @@ export function createFetchWithAuth(firebaseApp?: FirebaseApp): typeof fetch {
...
         if (currentUser) {
-            const token = await currentUser.getIdToken();
-            const headers = new Headers(init?.headers);
-            headers.set("Authorization", `Bearer ${token}`);
-            init = {
-                ...init,
-                headers,
-            };
+            try {
+                const token = await withTimeout(currentUser.getIdToken(), 5000);
```
KevinDCarlson, line 50: this can sometimes hang forever if there's a stale session in the browser *and* your current domain isn't Firebase-authorized, i.e. if you're futzing around a lot in Netlify previews.

epatters:
Thanks Kevin! @kasbah, can you please review whether this is the right approach to solve this?

packages/frontend/src/api/rpc.ts
kasbah: What do you mean by "futzing around"? You tried to login via Google/Github on the preview?

KevinDCarlson: No, I don't think you have to have tried to log in to make this break, it's something about the mere request for a firebase token that then gets turned down. I wouldn't be able to fully diagnose it myself but I could send you some robot's description if you like.

KevinDCarlson: Here is what the robot says: 

Symptom: On the sirius Netlify branch preview (branch-sirius--catcolab.netlify.app, built in staging → backend-next + Firebase project catcolab-next), any backend call hung forever — "creating…" spinner never resolved, import silently reset, zero RPC requests in the Network tab, nothing logged. Incognito worked fine.

Root cause: The frontend attaches auth per-request in createFetchWithAuth (packages/frontend/src/api/rpc.ts):
```
if (currentUser) {
    const token = await currentUser.getIdToken();  // <-- stalls, never settles
    headers.set("Authorization", `Bearer ${token}`);
}
return await fetch(...);   // never reached
```
`currentUser.getIdToken()` never resolved, so the fetch was never dispatched. It stalls because the preview origin isn't in catcolab-next's Authorized domains, and Firebase's token-refresh path coordinates through the cross-origin gapi iframe hosted at the authDomain — on an unauthorized origin that handshake never posts back (it doesn't error, it just hangs).

KevinDCarlson: Actually, we're confused about how `currentUser` could ever have gotten to be non-null on a non-Firebase-whitelisted domain, and the forensics trail is cold. But the incognito check seems to prove that's what somehow happened. Might be some deeper mystery in here, but cutting off these infinite hangs does at least fix the symptom. Or, if this is ever observed again, whitelisting that preview on Firebase also fixes it. Sorry I can't actually explain how it happened.

epatters:
Closing this out for now due to lack of clarity about what is being fixed.

PR #1334, opened 2026-07-08
RFC-0006: Notebook API for Typescript

kasbah:
> Catlog and its WASM interface expose a rich and flexible API surface for formal modeling. The full flexibility of this is available to our frontend via Typescript, with types largely generated from Rust, but the ergonomics and
type-safety of this API are sub-par.
>
> Colocation of the Typescript handling `catlog-wasm` and our `catcolab-document-types` with the frontend has led to entanglement of UI framework concerns with the modeling API.
>
> We want to improve this API surface with the aim of helping three main use-cases:
>
> 1. Our use in the CatColab frontend
> 2. Other programmatic use from Javascript and Typescript
> 3. Users instructing LLMs
> 
> We will focus on notebook documents, since that is the only document currently in use, but this API should be expandable to other types of documents in the future.
> 

https://branch-kb-rfc-0006--catcolab.netlify.app/rfc/0006

### DIFF
--- a/rfc/0006.md
+++ b/rfc/0006.md
@@ -669,10 +669,13 @@ if (wrong.tag === "Err") {
 issues: Cannot load document with theory "petri-net" using a shape with theory "simple-olog".
 ```
 
-We can also load a notebook from a handle.
+We can also load a notebook from a `DocumentRef`.
 
 ```ts
-const loaded = await binder.loadNotebookFromHandle(PetriNet, notebook.handle);
+const loaded = await binder.loadNotebookFromRef(PetriNet, {
+    id: "some-document-id",
+    version: null,
+});
 ```
 
 ### Migrating between logics
@@ -1233,12 +1236,19 @@ plug in custom backends. We offer the `createBinder` method and the
 <!-- verifier:reset -->
 
 ```ts
-import type { Document, Link } from "catcolab-document-types";
+import type { Document } from "catcolab-document-types";
+import type { Result } from "catcolab-documents";
+interface DocumentRef {
+    id: string;
+    version: string | null;
+    server?: string;
+}
 interface DocumentStore<Handle> {
     // An async function to create a document handle from inital data
     createHandle(initialDoc: Document): Promise<Handle>;
-    // An async function to get a document handle from a `Link`
-    getHandle(link: Link): Promise<Handle | undefined>;
+    // An async function to get a document handle from a `DocumentRef`,
+    // as a `Result` (`Ok` with the handle, or `Err` with issues).
+    getHandle(ref: DocumentRef): Promise<Result<Handle>>;
     // Apply modifications to a handle.
     changeDocument(handle: Handle, fn: (doc: Document) => void): void;
     // Subscribe change callbacks to our store for `onChange`. Returns a
@@ -1246,8 +1256,8 @@ interface DocumentStore<Handle> {
     subscribe(handle: Handle, callback: () => void): () => void;
     // Copy values (with any proxies removed)
     copyValue<T>(handle: Handle, value: T): T;
-    // Get the link for a handle
-    linkForHandle(handle: Handle): Omit<Link, "type"> | undefined;
+    // Get the reference for a handle
+    getDocumentRef(handle: Handle): DocumentRef;
     // Get a document view from a handle
     getDocumentView(handle: Handle): Readonly<Document>;
 }
@@ -1274,6 +1284,18 @@ type SolidStoreHandle = {
     listeners: Set<() => void>;
 };
 
+// Every store mints a stable reference for its handles; this one assigns an id
+// on demand and keeps it in a WeakMap.
+const solidStoreIds = new WeakMap<SolidStoreHandle, string>();
+const solidStoreIdFor = (handle: SolidStoreHandle): string => {
+    let id = solidStoreIds.get(handle);
+    if (!id) {
+        id = crypto.randomUUID();
+        solidStoreIds.set(handle, id);
+    }
+    return id;
+};
+
 const solidStore: DocumentStore<SolidStoreHandle> = {
     async createHandle(initialDoc) {
         const draftDoc = structuredClone(initialDoc as Document);
@@ -1295,9 +1317,12 @@ const solidStore: DocumentStore<SolidStoreHandle> = {
     },
     copyValue: (_handle, value) => structuredClone(unwrap(value)),
     getDocumentView: (handle) => handle.docView,
+    getDocumentRef: (handle) => ({ id: solidStoreIdFor(handle), version: null, server: "" }),
     // Link resolution omitted for brevity.
-    getLinkForHandle: () => undefined,
-    getHandle: async () => undefined,
+    getHandle: async () => ({
+        tag: "Err",
+        content: [{ message: "This store cannot resolve references." }],
+    }),
 };
 
 const solidBinder = createBinder(solidStore);
@@ -1342,9 +1367,12 @@ const automergeStore: DocumentStore<DocHandle<Document>> = {
         return getBackend(doc).materialize(objId!) as typeof value;
     },
     getDocumentView: (handle) => handle.doc(),
+    getDocumentRef: (handle) => ({ id: handle.documentId, version: null, server: "" }),
     // Link resolution omitted for brevity.
-    getLinkForHandle: () => undefined,
-    getHandle: async () => undefined,
+    getHandle: async () => ({
+        tag: "Err",
+        content: [{ message: "This store cannot resolve references." }],
+    }),
 };
 
 const automergeBinder = createBinder(automergeStore);
@@ -1421,30 +1449,33 @@ const backendStore: DocumentStore<StoreHandle> = {
         const objId = getObjectId(value as object);
         return getBackend(doc).materialize(objId!) as typeof value;
     },
-    getLinkForHandle: (handle) => {
+    getDocumentRef: (handle) => {
         const refId = refByDocId.get(handle.docHandle.documentId);
         if (!refId) {
-            return undefined;
+            throw new Error("handle is not registered with this store");
         }
-        return { _id: refId, _version: null, _server: backend.serverHost };
+        return { id: refId, version: null, server: backend.serverHost };
     },
-    getHandle: async (link) => {
-        const refId = link._id;
+    getHandle: async (ref) => {
+        const refId = ref.id;
         const cached = handleByRefId.get(refId);
         if (cached) {
-            return cached;
+            return { tag: "Ok", content: cached };
         }
 
         const result = await backend.get_doc(refId);
         if (result.tag !== "Ok") {
-            return undefined;
+            return {
+                tag: "Err",
+                content: [{ message: `Cannot resolve reference "${refId}".`, path: ["id"] }],
+            };
         }
 
         const docHandle = await repo.find<Document>(result.content.docId as DocumentId);
         const handle = makeHandle(docHandle);
         handleByRefId.set(refId, handle);
         refByDocId.set(docHandle.documentId, refId);
-        return handle;
+        return { tag: "Ok", content: handle };
     },
 };
 
@@ -2137,6 +2168,16 @@ type SolidStoreHandle = {
     listeners: Set<() => void>;
 };
 
+const solidStoreIds = new WeakMap<SolidStoreHandle, string>();
+const solidStoreIdFor = (handle: SolidStoreHandle): string => {
+    let id = solidStoreIds.get(handle);
+    if (!id) {
+        id = crypto.randomUUID();
+        solidStoreIds.set(handle, id);
+    }
+    return id;
+};
+
 const solidStore: DocumentStore<SolidStoreHandle> = {
     async createHandle(initialDoc) {
         const draftDoc = structuredClone(initialDoc as Document);
@@ -2158,9 +2199,12 @@ const solidStore: DocumentStore<SolidStoreHandle> = {
         };
     },
     copyValue: (_handle, value) => structuredClone(unwrap(value)),
+    getDocumentRef: (handle) => ({ id: solidStoreIdFor(handle), version: null, server: "" }),
     // Link resolution omitted for brevity.
-    getLinkForHandle: () => undefined,
-    getHandle: async () => undefined,
+    getHandle: async () => ({
+        tag: "Err",
+        content: [{ message: "This store cannot resolve references." }],
+    }),
 };
 
 const solidBinder = createBinder(solidStore);
@@ -2654,8 +2698,15 @@ function makeAutomergeRichTextStore(): DocumentStore<StoreHandle> {
             const objId = getObjectId(value as object);
             return getBackend(handle.docHandle.doc()).materialize(objId!) as typeof value;
         },
-        getLinkForHandle: () => undefined,
-        getHandle: async () => undefined,
+        getDocumentRef: (handle) => ({
+            id: handle.docHandle.documentId,
+            version: null,
+            server: "",
+        }),
+        getHandle: async () => ({
+            tag: "Err",
+            content: [{ message: "This store cannot resolve references." }],
+        }),
         // This is needed for the RichTextCellEditor to bypass the notebook interface.
         getRichTextRef: (handle, cellId) => ({
             docHandle: handle.docHandle,
--- a/rfc/0006.md
+++ b/rfc/0006.md
@@ -3064,6 +3064,110 @@ for (const mapping of model.judgmentsOf(Mapping)) {
 }
 ```
 
+### Transactions
+
+It would be convenient for scripting to be able to abort a set of changes on
+errors and generally revert change sets. A transaction ability could also allow
+us to make our undo history feature more intuitive, keeping track of local
+edits in addition to the global snapshots.
+
+<!-- verifier:reset -->
+
+```ts
+import { createBinder } from "catcolab-documents";
+import { Entity, SimpleSchema } from "catcolab-logics/simple-schema";
+
+const binder = createBinder();
+const schema = await binder.createNotebook(SimpleSchema, { title: "Example schema" });
+
+const tx = schema.beginTransaction();
+tx.add(Entity, { label: "Person" });
+tx.add(Entity, { label: "Company" });
+
+console.log("schema:", schema.cellsOf(Entity).length);
+console.log("tx:", tx.cellsOf(Entity).length);
+
+const commit = tx.commit();
+console.log(
+    "schema:",
+    schema
+        .cellsOf(Entity)
+        .map((cell) => cell.label)
+        .join(", "),
+);
+
+schema.revertCommit(commit);
+console.log("reverted:", schema.cellsOf(Entity).length);
+```
+
+```
+schema: 0
+tx: 2
+schema: Person, Company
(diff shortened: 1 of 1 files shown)


tim-at-topos:
Not a very technical review, but this looks pretty good to me, and I'm excited to see some design work around APIs (getting one step closer to emacs `catcolab-mode` 😈 ). Below are the few minor thoughts I had while reading.

- [2.9] Is there a reason to prefer in-place rewriting for migrations, rather than creating a new copy? I know that this mirrors how we currently do it, but I'm not sure if we want to really commit to that. Something like `migrateTo` and `migrateTo` (pretending this is Ruby, where the `!` means that this edits and returns the success, rather than returning an edited fresh copy) could be nice.
- [2.14.3 and 2.14.4] I'm very excited about how what you call "Shapes" seem (to me) to be a step towards what I was calling ["builder schemas"](https://next.catcolab.org/rfc/nonactionable/0003), where we separate the shapes of things that a notebook can build from the way that we interpret it. To be a bit more concrete, I can imagine something like `myCausalLoopDiagram.evaluateAs(PetriNet)` working, as long as somewhere an "interpretation" (i.e. map) of shapes `CausalLoopDiagram->PetriNet` has been given.
- [2.14.5 and 2.15] I think these code examples should have syntax highlighting turned on?

kasbah:
> [2.14.5 and 2.15] I think these code examples should have syntax highlighting turned on?

Looks like we are missing highlighting for `tsx` code blocks on the RFC Quarto. It is working on Github: e.g. [2.14.5 Shape consumer](https://github.com/ToposInstitute/CatColab/blob/kb/rfc-0006/rfc/0006.md#shape-consumer). 

> [2.9] Is there a reason to prefer in-place rewriting for migrations, rather than creating a new copy? I know that this mirrors how we currently do it, but I'm not sure if we want to really commit to that. Something like migrateTo and migrateTo (pretending this is Ruby, where the ! means that this edits and returns the success, rather than returning an edited fresh copy) could be nice.

The whole RFC is a rationalization and disentanglement of our current approach. Beyond the suggested "scripting" capability no new functionality will be added. The change should be invisible to the user. 

While I don't disagree that making a copy on migration would be better in a lot of ways (some downsides too), proposing that change in this RFC feels like the wrong place. Ideally we can adopt this API and then make changes like that after that, reflecting them both in the API and in our frontend.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+const intro = notebook.add(RichText, { content: "We define a simple olog." });
+```
+
+We can create objects and morphisms in the notebook.
+
+<!-- verifier:prepend-to-following -->
+
+```ts
+const source = notebook.add(Type, {
+    name: "A",
```
epatters: For historical reasons it's not consistent across the codebase, but I'm now trying to use "name" or "id" for IDs (in notebooks, UUIDs; in plain text, symbolic identifiers) and "label" for human-readable labels. See, e.g., `QualifiedName` vs `QualifiedLabel` in `catlog`.

I understand why you're using "name" here: because that's what the document types are already doing. Likely now is not the right time to perform this migration. But I did want to flag it for general awareness.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+```
+a: A
+b: B (copy of A)
+```
+
+</details>
+
+#### Re-ordering cells
+
+Every cell handle can move itself within the notebook. Moves locate the cell
```
epatters, line 379: Slick API for re-ordering and deleting.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+</details>
+
+### A `Result` type
+
+In cases where where functions take in unknown and user generated inputs we want a
+standardized way to return the result on success and diagnostics on failure.
+
+Our `Result` type is a tagged union structurally identical to the `JsResult`
+type in our Rust bindings (`catlog-wasm`). Results coming out of the Rust layer
+can therefore pass through untouched.
```
epatters, line 557: Nice. I like this approach. As we've discussed, `Result`-based error handling is less ergonomic in TS than in Rust, but it still seems preferable to the default approach of throwing and catching exceptions whose values aren't even properly typed.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+<!-- verifier:prepend-to-following -->
+
+```ts
+import { PetriNet } from "catcolab-logics/petri-net";
+import { binder } from "catcolab-documents";
+
+const notebook = await binder.createNotebook(PetriNet, { name: "Example Petri-net" });
+```
+
+We can dump a notebook.
```
epatters, line 705: I'm assuming that `dump` means "dump to JSON" and comes with the usual caveats about losing rich text annotations. Since this API abstracts over whether Automerge is being used, I don't see an alternative. I suppose we just have to accept that the abstraction is a little leaky.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+if (wrong.tag === "Err") {
+    console.log("issues:", wrong.content.map((issue) => issue.message).join("; "));
+}
+```
+
+```
+issues: Cannot load document with theory "petri-net" using a shape with theory "simple-olog".
+```
+
+We can also load a notebook from a handle.
```
epatters: I'm confused about the terminology here. Is a "handle" an Automerge document handle or something abstracted from that?

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+
+```
+name: Olog analysis
+type: model
+```
+
+<!-- verifier:prepend-to-following -->
+
+```ts
+import { Visualization } from "catcolab-analyses";
```
epatters, line 944: Interesting. So will the visualization analyses for, say, CLDs, Petri nets, and stock-flow diagrams all get grouped into the `Visualization` type, despite IIRC having similar but slightly different configuration options? Maybe I'll find out later.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+Notebook API makes sense and is free of misspellings or other issues
+that would throw errors at runtime.
+
+Since the API can also be used from Javascript directly without type checking
+we want runtime validation that inputs match our expected schema.
+
+At runtime invalid inputs:
+
+- Should _throw_ an exception, ideally carrying a Standard Schema compatible `issues` array, _if_ it is something Typescript would have caught
+- Should return a `Result` type otherwise
```
epatters, line 1083: Yes, I agree with this dichotomy.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+
+<!-- verifier:prepend-to-following -->
+
+```ts
+import { createEffect, createRoot } from "solid-js";
+import { createStore, reconcile, type SetStoreFunction, unwrap } from "solid-js/store";
+import { binder, createBinder, type DocumentStore } from "catcolab-documents";
+import { type Document } from "catcolab-document-types";
+
+type SolidStoreHandle = {
```
epatters, line 1363: Alright, this answers my question above. Your concept of `Handle` abstracts over Automerge's. Seems like a good approach to me.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+`FakeBackend`).
+
+<!-- verifier:reset -->
+
+<!-- verifier:prepend-to-following -->
+
+```ts
+import { getBackend, getObjectId } from "@automerge/automerge";
+import { type DocHandle, type DocumentId } from "@automerge/automerge-repo";
+import { makeDocumentProjection } from "@automerge/automerge-repo-solid-primitives";
```
epatters, line 1482: Glad to see us adopting this library now.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+
+</details>
+
+### Defining notebook shapes
+
+#### A logic
+
+We provide some utilities for defining a logic: `defineMorphism`,
+`defineObject` and `defineShape`. Here is what the definition of the `PetriNet`
+logic looks like:
```
epatters, line 1596: Let me check that I understand what's going on.

Currently, we define a logic/theory in the frontend by passing one big blog of JSON to `TheoryLibrary.add`. You replacing that API with a more granular one involving helpers `defineObject`, `defineMorphism`, and `defineShape`. This has the advantage that you can create side constants along the way like `Place` and `Transition`, which form part of the custom logic for this API. And that resolves my worry about previous designs that adding a new logic was already enough work and we shouldn't need a new, separate chunk of code to define the TS API: in this approach, you write the same amount of TS and get the API for free.

Is that right? If so, I think that's wonderful. I also think you should expand the text in the RFC a bit to clarify those implications of the design.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+let thSymMonoidalCategory: Promise<ThSymMonoidalCategory> | undefined;
+
+function getCachedTheory(): Promise<ThSymMonoidalCategory> {
+    if (thSymMonoidalCategory === undefined) {
+        thSymMonoidalCategory = import("catlog-wasm").then(
+            ({ ThSymMonoidalCategory }) => new ThSymMonoidalCategory(),
+        );
+    }
+    return thSymMonoidalCategory;
+}
```
epatters, line 1699: The helper `getCachedTheory` seems a slightly annoying piece of boilerplate to include for each analysis. Can this be avoided?

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+const BasicObj = defineObject({ tag: "Basic", content: "Object" });
+
+const OnlyBasicObj = defineShape({
+    objects: [BasicObj],
+});
+```
+
+Because `OnlyBasicObj` is a subset of both the olog and petri-net logics (they
+share the basic `Object`), we can write a generic function against it that
+works with notebooks from either logic.
```
epatters, line 1746: Very cool, a nice implication of your design.

rfc/0006.md
```diff
@@ -0,0 +1,2966 @@
...
+rhs: employer
+equations: 1
+valid: true
+```
+
+</details>
+
+#### Instances
+
+Non-diagram instances are being added to CatColab. We can envision an API for
```
epatters: These "non-diagram instances" can be called "tabular instances" instead. (In the API, I think it's good to just say `Instance` as you've done below.)

epatters (commented):
Thanks Kaspar, this is an excellent RFC, both in the content itself and in the clear and thorough presentation. I've got a few small questions/comments below, but, assuming I've understood things correctly, we should be able to move rapidly toward merging.

rfc/0006.md
kasbah: Since this API is a wrapper I think it is the right place to lead in adopting this terminology. I have changed `name` to `label` throughout.

rfc/0006.md
kasbah: Thanks, I just want to point out that the strategy is a combination of `Result` and throwing `Exceptions`. https://branch-kb-rfc-0006--catcolab.netlify.app/rfc/0006#type-errors-runtime-errors

(EDIT: GH was hiding some comments from me, so just noticed you have actually commented on that too)

Not highlighted there are that the line between exception and control-flow is blurred on the boundaries when using `Promise` in combination with http requests.

rfc/0006.md
kasbah: Yes, where "JSON" is a JS object that can be serialized to JSON. At least our existing JSON import / export has the same problems so we can justify it for `dump` and `loadNotebook`.

For interfacing to the editor I added a explicit leak in the form of [`getRichTextRef` and `cell.editorRef`](https://branch-kb-rfc-0006--catcolab.netlify.app/rfc/0006#rich-text-handling). 

Thinking about it more though, programmatic edits being able to clobber the rich text cells is really not good, we should probably do something about that. One abstraction I had though of is to convert from and into markdown for this API but I haven't looked at the feasibility.

```ts
notebook.add(RichText, {markdown: "# Title"});
```

rfc/0006.md
kasbah: It is the `Handle` type parameter in `DocumentStore<Handle>` defined when creating the binder. 

This does raise the question of how we get our custom handle when we have a Solid+Automerge handle that looks like:

```ts 
import type { DocHandle } from "@automerge/automerge-repo"
import type { Document } from "catcolab-document-types";
import type { DocumentStore } from "catcolab-documents";

type AutomergeSolidHandle = {
    docHandle: DocHandle<Document>;
    docView: Document;
};

const documentStore = DocumentStore<AutomergeSolidHandle> = {
    ...
}
```

I've omitted to have examples of this used with `loadNotebookFromHandle` and overlooked that it is no longer just an Automerge `DocHandle` (which it was at one point in my design). 

Since we will need to define something like this in the binder for instantiation resolution: 
```ts
        getHandle: async (link: Link) => {
            const refId = link._id;

            const result = await backend.get_doc(refId);
            if (result.tag !== "Ok") {
                throw new Error("Could not get refId:", refId)
            }

            const docHandle = await repo.find<Document>(result.content.docId as DocumentId);
            const handle = { docHandle, docView: makeDocumentProjection(docHandle) };

            return handle;
        },
```

We should have a way make use of this from the binder. I will replace `loadNotebookFromHandle` with:

```ts
binder.loadNotebookFromLink(PetriNet, link)
```

rfc/0006.md
kasbah: I guess you didn't find out later because I haven't fully thought it through. I do think one analysis e.g. `Visualization` should take consistent parameters. If it takes different parameters then we should give it another name.

One thing I did look at for visualization is to see if we can in principle disentangle the SVG rendering from SolidJS which does seem feasible if a little bit fiddly.

rfc/0006.md
kasbah: The reason I did it this way as opposed to caching "in the document store" is that there doesn't seem to be a consistent thing to cache for the analyses (unlike for elaboration and validation of the model notebooks -- which we do cache in the document store). An analysis could in principle want to load and cache anything so it seems like the responsibility of the analysis implementer to take care of caching.

It feels like there could be a better design here but I haven't thought of it.

rfc/0006.md
kasbah: I didn't realize I hadn't made that clear before! It is sort of fundamental to the whole approach that I actually don't even know what to say about it, which might be part of the problem.

rfc/0006.md
kasbah: I changed it to

```ts
binder.loadNotebookFromRef(shape: Shape, ref: DocumentRef)
```

Where `DocumentRef` is a `StableRef`-like type I added.

It also prompted me to change the `DocumentStore` definition around the `DocumentRef` type (it used `Link` before) and to use `Result` in ref-resolution.

kasbah:
I added a section on ["Validated models"](https://branch-kb-rfc-0006--catcolab.netlify.app/rfc/0006#validated-models) to the "Future extensions".

rfc/0006.md
kasbah: I am now wondering what to do about the notebook name. Should it become `title` or also `label`?

```ts
binder.createNotebook(PetriNet, {label: "My petri net"})
```

epatters: I'd go with `title` myself, because saying that documents have titles is so common, even if using `label` would be more consistent. No really strong opinion myself, though.

rfc/0006.md
epatters: Conversion from/to Markdown seems worth exploring, in particular for having LLMs generate rich text. I'm happy to defer this to future work though if it seems like a lot of extra work.

rfc/0006.md
epatters: Ok, glad I'm understanding correctly. With that in mind, here's what I suggest to improve the exposition.

It seems that the idea of a "shape" is key to the whole design. However, almost nothing about shapes is said until Sec 2.14, which is about 2/3rds of the way through a long document. Moreover, the introductory part of the RFC (Sec 1) is very short, just a few sentences before launching into the technical details (Sec 2). I'd suggest:

- Rename Sec 1 from "Summary and motivation" to "Motivation"
   - You could also expand the terse remark "...has led to entanglement of UI framework concerns with the modeling API" to clarify that we haven't had *any* API for document manipulation (outside the type defs), leading the UI components to perform lots of direct JSON manipulation
- Add a new Sec 2 called "Technical approach" with a couple paragraphs about the high-level approach (shapes), the pitfalls it avoids (requiring more boilerplate to manually define APIs for each logic), and the advantages it confers (e.g., type-safe manipulation of documents with varying degrees of type narrowing). The aim is to give the reader some idea of the big picture before launching into the details.
- Move current Sec 2 to Sec 3 and call it "Technical specification"

rfc/0006.md
epatters: > I do think one analysis e.g. `Visualization` should take consistent parameters. If it takes different parameters then we should give it another name.

Ok, good. That was the point I was hung up on. I don't need further clarity about the specific visualization analyses at this stage.

> One thing I did look at for visualization is to see if we can in principle disentangle the SVG rendering from SolidJS which does seem feasible if a little bit fiddly.

That would be nice as a matter of separation of concerns (though presumably out of scope for this RFC). OTOH, having the SVG outside of Solid would seem to limit our ability to have interactions between the visualization and the rest of the app: something we're not currently doing, but which I've long thought would be nice. E.g., it would be cool to be able to click on a node or arrow in a visualization and get taken to the place in the model where it is defined.

rfc/0006.md
kasbah: I changed it to title.

tslil-topos:
@kasbah would you mind adding `OBEE-5 ` as a prefix to this PR?

rfc/0006.md
kasbah: That makes sense. I have re-organized as you suggested though I couldn't think of more than a paragraph for sec 2.

rfc/0006.md
```diff
@@ -0,0 +1,3258 @@
...
+accompanied by an ergonomic and type-safe API for its models and instances.
+
+Let's first review the current mechanism for extending the frontend with a new
+theory. To add a theory, one calls a method on `TheoryLibrary` and passes a
+large JSON object that configures, among other things, how each object and
+morphism type in the theory will be labeled and displayed in the frontend. This
+specification is indeed declarative (although that is not currently exploited,
+say by a plugin system, as all the JSON constants are defined explicitly in the
+CatColab standard library).
+
```
kasbah: Unless I am misunderstanding something, I don't think JSON is the right word. `TheoryLibrary.add`, in addition to the `meta`,  takes a `Promise<TheoryConstructor>` that is not JSON serializable.

rfc/0006.md
```diff
@@ -0,0 +1,3258 @@
+---
+title: |
+    `0006` Document API for TypeScript
```
kasbah: I had named this "Notebook" specifically because that's what the actual spec covers. We have added some speculative extensions but they are all in "Future extensions" and more fluid in my eyes.

rfc/0006.md
kasbah: With https://github.com/ToposInstitute/CatColab/pull/1341 the `.dump` caveats would be gone at least.

rfc/0006.md
epatters: That's fair. It's not pure data, though it's pretty close. I'll edit this passage to avoid confusion.

epatters (approved):
Alright, I'm ready to merge this one if you are, @kasbah!

PR #1336, opened 2026-07-09
Syntax highlighting for TSX code in RFCs

kasbah:


tim-at-topos (approved):
happy this was such a simple fix :-)

PR #1338, opened 2026-07-14
OBEE-8 user db: add nullable inference_{key,hash} to user's table, create migration

tslil-topos:
Part of the [LLM Integration](https://app.notion.com/p/toposinstitute/LLM-Integration-38269ce798648055aa78db0eef3c95dc) outcome, this is the first of ~5 PR's which adds the columns to the db that will see future use.

Forthcoming changes:
* [part two](https://app.notion.com/p/toposinstitute/LLM-Integration-2-5-Backend-service-39d69ce7986480a8b685d87e1aa8335a?source=copy_link)
* [part three](https://app.notion.com/p/toposinstitute/LLM-Integration-3-5-secret-addition-39d69ce7986480cd8865defaf0d547ec?source=copy_link)
* [part four](https://app.notion.com/p/toposinstitute/LLM-Integration-4-5-front-end-token-acquisition-39d69ce79864805d8b4fd9a7ac4a8b27?source=copy_link)
* [part five](https://app.notion.com/p/toposinstitute/LLM-Integration-5-5-FE-inferenc-39d69ce7986480fcad11d1c55129e60e?source=copy_link)

### DIFF
--- a/packages/migrator/src/migrations/m20260714000000_add_inference_keys.rs
+++ b/packages/migrator/src/migrations/m20260714000000_add_inference_keys.rs
@@ -1,0 +1,48 @@
+use sqlx::{PgConnection, Postgres};
+use sqlx_migrator::Operation;
+use sqlx_migrator::error::Error;
+use sqlx_migrator::migration;
+use sqlx_migrator::vec_box;
+
+pub(crate) struct AddInferenceKeys;
+
+migration!(
+    Postgres,
+    AddInferenceKeys,
+    "backend",
+    "20260714000000_add_inference_keys",
+    vec_box![],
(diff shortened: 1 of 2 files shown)


epatters (approved):
LGTM. Happy to merge unless @kasbah has anything he'd like to add.

kasbah:
I was just going to wait if the follow up reveals anything that should go into the migration.

On thing we could think about is making the columns `NOT NULL` and generating keys for every user in the migration. It's a bit more work upfront but lets us not worry about NULL fields down the line.

kasbah:
On second thought: scratch that. If we fail to generate a key for a new user we still want them to have an account so they should indeed be nullable.

PR #1339, opened 2026-07-14
Correct behaviour for maximum path length in submodel search

### LINKED ISSUE #564: Panic when setting "Maximum length of path" to negative number
If I create a causal loop diagram, add a balancing/reinforcing loops analysis, and set the "Maximum length of path" value to be a negative number (by just typing one into the box) then I get a generic panic error. This should be validated in the same way that e.g. the input for Lotka–Volterra simulation is (`validate: (_, data) => data >= 0,`), but the structure of `submodel_graphs.tsx` is different enough from that of `lotka_volterra.tsx` that I can't quickly figure out how to fix this.

tim-at-topos:
Fixes #564 as well as an unreported bug (namely that enabling and then disabling the checkbox didn't actually disable it at all).

### DIFF
--- a/packages/frontend/src/stdlib/analyses/checker_types.ts
+++ b/packages/frontend/src/stdlib/analyses/checker_types.ts
@@ -15,10 +15,10 @@ export type MotifFindingAnalysisContent = {
     activeIndex: number;
 
     /** Whether or not to search up to a maximum length of paths. */
-    enableMaxPathLength: boolean;
+    enableMaxPathLength?: boolean;
 
     /** Maximum length of paths used in morphism search. */
-    maxPathLength: number | null;
+    maxPathLength?: number | null;
 };
 
 export type ReachabilityChecker = (model: DblModel, data: ReachabilityProblemData) => boolean;
--- a/packages/frontend/src/stdlib/analyses/checker_types.ts
+++ b/packages/frontend/src/stdlib/analyses/checker_types.ts
@@ -14,8 +14,11 @@ export type MotifFindingAnalysisContent = {
     /** Index of active submodel. */
     activeIndex: number;
 
+    /** Whether or not to search up to a maximum length of paths. */
+    enableMaxPathLength: boolean;
+
     /** Maximum length of paths used in morphism search. */
-    maxPathLength?: number | null;
+    maxPathLength: number | null;
 };
 
 export type ReachabilityChecker = (model: DblModel, data: ReachabilityProblemData) => boolean;
--- a/packages/frontend/src/stdlib/analyses/submodel_graphs.tsx
+++ b/packages/frontend/src/stdlib/analyses/submodel_graphs.tsx
@@ -107,15 +107,14 @@ export default function SubmodelGraphs(
                                 min="0"
                                 label="Maximum length of path"
                                 value={props.content.maxPathLength ?? ""}
-                                onChange={(evt) =>
-                                    props.changeContent((content) => {
-                                        content.maxPathLength =
-                                            evt.currentTarget.valueAsNumber > 0 &&
-                                            Number.isInteger(evt.currentTarget.valueAsNumber)
-                                                ? evt.currentTarget.valueAsNumber
-                                                : null;
-                                    })
-                                }
+                                onChange={(evt) => {
+                                    const value = evt.currentTarget.valueAsNumber;
+                                    if (value > 0 && Number.isInteger(value)) {
+                                        props.changeContent((content) => {
+                                            content.maxPathLength = value;
+                                        });
+                                    }
+                                }}
                             />
                         </Show>
                     </FormGroup>
--- a/packages/frontend/src/stdlib/analyses/submodel_graphs.tsx
+++ b/packages/frontend/src/stdlib/analyses/submodel_graphs.tsx
@@ -11,10 +11,6 @@ import { modelToGraph } from "./model_graph";
 
 import "./submodel_graphs.css";
 
-function isNumber(num: number | null | undefined): num is number {
-    return (num as number) !== undefined;
-}
-
 /** Find submodels of a model and visualize them as graphs. */
 export default function SubmodelGraphs(
     props: {
@@ -27,7 +23,7 @@ export default function SubmodelGraphs(
         // If `enableMaxPathLength` has not been set, then set it to be true iff `maxPathLength`
         // already has a (numerical) value.
         if (props.content.enableMaxPathLength === undefined) {
-            if (isNumber(props.content.maxPathLength)) {
+            if (props.content.maxPathLength != null) {
                 props.content.enableMaxPathLength = true;
             } else {
                 props.content.enableMaxPathLength = false;
@@ -44,7 +40,7 @@ export default function SubmodelGraphs(
             }
             return props.findSubmodels(validated.model, {
                 maxPathLength:
-                    enableMaxPathLength() && isNumber(props.content.maxPathLength)
+                    enableMaxPathLength() && props.content.maxPathLength != null
                         ? props.content.maxPathLength
                         : null,
             });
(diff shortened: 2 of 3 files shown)


packages/frontend/src/stdlib/analyses/submodel_graphs.tsx
```diff
@@ -80,19 +82,22 @@ export default function SubmodelGraphs(
...
                                 min="0"
                                 label="Maximum length of path"
                                 value={props.content.maxPathLength ?? ""}
                                 onChange={(evt) =>
                                     props.changeContent((content) => {
-                                        content.maxPathLength = evt.currentTarget.valueAsNumber;
+                                        content.maxPathLength =
+                                            evt.currentTarget.valueAsNumber > 0
+                                                ? evt.currentTarget.valueAsNumber
+                                                : null;
```
epatters: While we're at it improving the validation: should we also check that the value is an integer?

packages/frontend/src/stdlib/analyses/submodel_graphs.tsx
```diff
@@ -80,19 +82,22 @@ export default function SubmodelGraphs(
                             checked={props.content.maxPathLength != null}
```
epatters: With the new field, this line is no longer correct, right?

packages/frontend/src/stdlib/analyses/checker_types.ts
```diff
@@ -14,8 +14,11 @@ export type MotifFindingAnalysisContent = {
     /** Index of active submodel. */
     activeIndex: number;
 
+    /** Whether or not to search up to a maximum length of paths. */
+    enableMaxPathLength: boolean;
```
epatters: We have to be a bit careful here since we're introducing a new field. As it stands, this PR changes the behavior of existing notebooks because `enableMaxPathLength` being `undefined` is "false-y," hence if a max path length was previously set it will now be ignored, which is bad because that can crash the app on moderately large models :)

Actually, I don't quite understand why we're introducing the new field at all. I guess this was to fix the bug you noticed? "enabling and then disabling the checkbox didn't actually disable it at all"

epatters (changes requested):
Thanks for trying to fix these things up! A few questions/comments.

packages/frontend/src/stdlib/analyses/checker_types.ts
tim-at-topos: Yeah, the previous logic was that `maxPathLength` was of type `number | null | undefined`, and the (seemingly intended) way of disabling maximum path lengths was by setting this to be `null`, but this means that unchecking and then checking the checkbox would forget the previously set number. This new field lets us remember the previously set `maxPathLength` and still enable / disable using it.

For fixing this for existing notebooks (I always forget about this... thanks for catching it again!), I think it would suffice to have something like

```typescript
if (props.content.enableMaxPathLength === undefined) {
    if (typeof props.content.maxPathLength === "number") {
        props.content.enableMaxPathLength = true;
    } else {
        props.content.enableMaxPathLength = false;
    }
}
```

right at the top of `submodel_graphs.tsx`. Does that seem right to you?

Edit: the literal code as written above gives solid reactivity warnings, but I think I've figured out how to do this previously (with Kaspar's help) so should be able to do the same tricks here

packages/frontend/src/stdlib/analyses/submodel_graphs.tsx
```diff
@@ -11,21 +11,38 @@ import { modelToGraph } from "./model_graph";
...
     } & ModelAnalysisProps<MotifFindingAnalysisContent>,
 ) {
+    // For compatibility with notebooks created before the `enableMaxPathLength` field was introduced.
+    function enableMaxPathLength(): boolean {
+        if (props.content.enableMaxPathLength === undefined) {
+            if (isNumber(props.content.maxPathLength)) {
+                return true;
+            } else {
+                return false;
+            }
```
epatters: I would remove the helper `isNumber` and replace this whole if-else block with `return props.content.maxPathLength != null`, which is equivalent to checking that `maxPathLength` is neither `null` nor `undefined`.

packages/frontend/src/stdlib/analyses/submodel_graphs.tsx
```diff
@@ -77,22 +94,26 @@ export default function SubmodelGraphs(
...
                                 label="Maximum length of path"
                                 value={props.content.maxPathLength ?? ""}
                                 onChange={(evt) =>
                                     props.changeContent((content) => {
-                                        content.maxPathLength = evt.currentTarget.valueAsNumber;
+                                        content.maxPathLength =
+                                            evt.currentTarget.valueAsNumber > 0 &&
+                                            Number.isInteger(evt.currentTarget.valueAsNumber)
+                                                ? evt.currentTarget.valueAsNumber
+                                                : null;
```
epatters: To avoid clearing out the old input when the new input is invalid, I'd do something like:

```ts
(evt) => {
  const value = evt.currentTarget.valueAsNumber;
  if (value > 0 && Number.isInteger(value)) {
    props.changeContent((content) => { content.maxPathLength = value; } );
  }
}
```

packages/frontend/src/stdlib/analyses/checker_types.ts
epatters: Since this field can be undefined and the code below handles that, the type def should be modified to account for that, i.e., become `enableMaxPathLength?: boolean`

packages/frontend/src/stdlib/analyses/checker_types.ts
```diff
@@ -14,8 +14,11 @@ export type MotifFindingAnalysisContent = {
...
+
     /** Maximum length of paths used in morphism search. */
-    maxPathLength?: number | null;
+    maxPathLength: number | null;
```
epatters: The same applies here: this field was added without doing a migration, so the honest type def is `maxPathLength?: ...`.

epatters (changes requested):
Thanks for fixing this up. A few more comments aiming to make this as clean as it can be given the situation.

The reason it's so annoying is that in principle we should be doing a migration here, but writing migrations for Automerge docs is itself a pain, so we end up playing these games. Perhaps in the future we should just bite the bullet and write the migration, I dunno.

packages/frontend/src/stdlib/analyses/submodel_graphs.tsx
tim-at-topos: I would love to do this but when I did this I got a lot of warnings about Solid reactivity... maybe they can be ignored?

tim-at-topos: Note that this is the same trick I had to pull in the ODE refactor analysis: https://github.com/ToposInstitute/CatColab/pull/1319/changes#diff-4e0d4bcce26cf4ddb6615d25a2b32688b183f27e0be2996582621363e83ab98bR9 . This was the only way that I could get the reactivity warnings to disappear.

epatters: Does it work if you use the form

const enableMaxPathLenth = () => ...

That would be a standard pattern that I'd expect to work without reactivity warnings.

If not, I'll look at it when I'm back at my computer. Or you could ask Kaspar.

tim-at-topos: It fixes reactivity, but now
```typescript
return props.findSubmodels(validated.model, {
    maxPathLength: enableMaxPathLength() ? props.content.maxPathLength : null,
});
```
gives an error, because the compiler can't infer that `enableMaxPathLength() = true` implies that `props.content.maxPathLength` is definitely not `undefined` 🙃

This is the type of issues I keep on coming up against with the TypeScript compiler, so I'll see if I can find a way to convince it to believe me and then tag Kaspar to see if he can simplify it :-)

tim-at-topos: @kasbah consider yourself tagged

kasbah: I pushed what @epatters suggested and it seems fine :grimacing:

tim-at-topos: ok, well that's good enough for me then!

epatters (approved):
I made one final change since props are intended to be read-only and should never be mutated.

PR #1340, opened 2026-07-15
OBEE-9 inference key management: backend RPC and CLI subcommand

tslil-topos:
This change depends on #1338.

If there's a way to have github only show the relative diff between this and #1338  i don't know it, other than making the target of this PR that branch which is not desirable.

(this change is large because of sqlx cache regeneration and Cargo changes for pulling in sha2)

## Design decisions

* OpenRouter is in charge of controlling the spending cap and refresh period for this cap, per key
* We hand users their raw inference keys, this does mean that a malicious user could extract these keys from their browser
  * alternatives here are drastically more complex, and would involve us standing up an inference forwarding service so that we could interpose our own keys
* Keys are valid for eternity
  * alternatives here for key roll-over introduce more front-end complexity to do with session invalidation
* Functionality is gated behind an environment variable `OPENROUTER_PROVISIONING_KEY` which is currently unset, so RPC/cli interaction will fail with appropriate errors.

## Implementation details

Add functionality to populate the columns `inference_{key, hash}` in the `user` table through `get_inference_key` endpoint. This reads the key out from the table if it exists, or otherwise reaches out to OpenRouter to create a new one. The constants controlling this are present in `inference.rs`:

```rust
/// Spending limit per user key, in USD.
const KEY_LIMIT: f64 = 5.0;
/// Reset period for the speding limit.
const KEY_LIMIT_RESET: &str = "daily";
/// The base URL for the OpenRouter key-management API.
pub const OPENROUTER_API_URL: &str = "https://openrouter.ai/api/v1";
```

Additionally add "invalidateinferencekeys" subcommand for the backend binary which takes usernames or firebase UIDs and invalidates the keys on OpenRouter, and nulls the column in the DB.

## Tests

This change introduces mock tests using `wiremock` with (censored) transcripts recorded by me. To run them do `cargo test --features integration-tests --test inference_tests`.

### DIFF
--- a/packages/backend/src/inference.rs
+++ b/packages/backend/src/inference.rs
@@ -1,0 +1,210 @@
+//! OpenRouter inference key management.
+//!
+//! Creates OpenRouter child keys per user and persists them in the `users`
+//! table. The stable key `hash` is stored alongside the secret so that keys can
+//! be invalidated via the OpenRouter API, `DELETE /keys/{hash}`. We leverage
+//! the OpenRouter API to put credit cap and refresh windows on the keys.
+
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+
+use crate::app::{AppCtx, AppError};
+
+/// Spending limit per user key, in USD.
+const KEY_LIMIT: f64 = 5.0;
+/// Reset period for the spending limit.
+const KEY_LIMIT_RESET: &str = "daily";
+
+/// The base URL for the OpenRouter key-management API.
+pub const OPENROUTER_API_URL: &str = "https://openrouter.ai/api/v1";
+
+/// Request body for `POST /keys` (create a child key).
+#[derive(Serialize)]
+struct CreateKeyRequest {
+    name: String,
+    limit: f64,
+    limit_reset: String,
+}
+
+/// The `data` object nested inside a `POST /keys` response.
+#[derive(Deserialize)]
+struct CreateKeyData {
+    hash: String,
+}
+
+/// Response from `POST /keys`.
+#[derive(Deserialize)]
+struct CreateKeyResponse {
+    key: String,
+    data: CreateKeyData,
+}
+
+/// Return the user's existing inference key, or create a new one.
+pub async fn get_inference_key(ctx: &AppCtx) -> Result<String, AppError> {
+    let Some(user) = &ctx.user else {
+        return Err(AppError::Unauthorized);
+    };
+
+    let provisioning_key = ctx
+        .state
+        .openrouter_provisioning_key
+        .as_ref()
+        .ok_or(AppError::InferenceUnavailable)?;
+
+    get_or_create_key(
+        &ctx.state.db,
+        &ctx.state.http_client,
+        &ctx.state.openrouter_base_url,
+        provisioning_key,
+        &user.user_id,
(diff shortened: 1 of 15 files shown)


packages/backend/tests/common/test_utils.rs
```diff
@@ -69,6 +71,8 @@ pub fn create_test_firebase_user(user_id: &str) -> FirebaseUser {
     .expect("Failed to create test FirebaseUser")
 }
 
+// This function is not used in all consuming binaries of test_utils.
+#[allow(dead_code)]
```
kasbah, line 75: Is it possible to feature gate this correctly instead?

kasbah (commented):
We already discussed the design decision of using RPC response vs UserState for the key and I'm still leaning towards keeping it as an RPC response as you have done.

Appreciate the testing strategy! Only one minor comment otherwise.

packages/backend/tests/common/test_utils.rs
tslil-topos: I'm not sure how, i think you might know more about feature gates than i do here? Searching around didn't provide an answer, but i could also be misunderstanding the situation. AFAICT:

* there are a bunch of different test modules
* these modules all import this file
* these modules themselves are gated in various ways
* these modules represent separate compilation units
* the test module that i added which calls on this file doesn't use this function

it looks like the thing we're looking for here is "only consider this piece of code when looking at these compilation units"?

kasbah: No, you are right. We might even consider allowing dead code for anything in `tests/common` because of this.

packages/backend/src/inference.rs
```diff
@@ -0,0 +1,210 @@
...
+//! the OpenRouter API to put credit cap and refresh windows on the keys.
+
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+
+use crate::app::{AppCtx, AppError};
+
+/// Spending limit per user key, in USD.
+const KEY_LIMIT: f64 = 5.0;
+/// Reset period for the speding limit.
```
epatters: Typo: `speding` -> `spending`

packages/backend/src/inference.rs
```diff
@@ -0,0 +1,210 @@
...
+    .fetch_optional(db)
+    .await?;
+
+    let Some(row) = row else {
+        return Ok(());
+    };
+
+    let Some(hash) = row.inference_hash else {
+        return Ok(());
+    };
```
epatters, line 85: A matter of preference, so feel to free to ignore, but it's more concise to replace the two blocks above with:

```rust
let Some(hash) = row.and_then(|row| row.inference_hash) else {
  return Ok(());
}
```

epatters (approved):
Thanks for the clear description in the PR as well as the integration tests. Looks great.

I agree that it makes sense to use RPC rather than user state since we don't expect a key to change during a session.

Just a few trivial comments below.

packages/backend/src/inference.rs
tslil-topos: thanks, i think i'll leave it this way otherwise later we have to unwrap row. It's an obvious invariant, but there's a sense in which we could make it structurally true (this form) and not have to introduce that.

PR #1343, opened 2026-07-15
Fixed random seed for stochastic mass-action

tim-at-topos:
Allow for a fixed random seed to be set for stochastic mass-action analyses. (I thought that this closed an old issue, but it turns out that we never actually made a GitHub issue for this, instead only writing it down in a DM)

to-do:
- [x] fix reactivity

### DIFF
--- a/packages/catlog/src/stdlib/analyses/stochastic/mass_action.rs
+++ b/packages/catlog/src/stdlib/analyses/stochastic/mass_action.rs
@@ -22,7 +22,7 @@ use tsify::Tsify;
 #[cfg_attr(feature = "serde-wasm", derive(Tsify))]
 #[cfg_attr(
     feature = "serde-wasm",
-    tsify(into_wasm_abi, from_wasm_abi, hashmap_as_object)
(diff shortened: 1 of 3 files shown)


packages/catlog/src/stdlib/analyses/stochastic/mass_action.rs
```diff
@@ -32,6 +32,9 @@ pub struct StochasticMassActionProblemData {
     #[cfg_attr(feature = "serde", serde(rename = "initialValues"))]
     pub initial_values: HashMap<QualifiedName, u32>,
 
+    /// Optional fixed seed, for reproducibility.
+    pub seed: Option<u64>,
+
```
kasbah, line 37: A cleaner way could be to tell serde to serialize this the `Option` as `number | undefined` rather than our implicit `number | null | undefined` and then the rest of the handling becomes cleaner since we only have to worry about `undefined` not `null` and don't have to "migrate" anything.

kasbah (approved):

epatters (approved):
Thanks, it's important to have this option for stochastic simulations! I took the liberty of cleaning up a few lingering issues (see commits).

PR #1344, opened 2026-07-16
OBEE-11 FE: inference key provider and gated RPC tests

tslil-topos:
This changes depends on #1340 .

If there's a way to have github only show the relative diff between this and #1340 i don't know it, other than making the target of this PR that branch which is not desirable.

## Design decisions

* Following #1340, the key is not written into the user-state Automerge doc. It is session-scoped state delivered by the RPC end-point. We may choose to revisit this
decision in the future.

* The inference key is held per-session in a Solid resource keyed on the Firebase user id, following the user state pattern.

## Implementation details

Add `InferenceKeyProvider` / `InferenceKeyContext` to the `MultiProvider` in `App.tsx`, mirroring the existing `UserStateProvider` / `UserStateContext` pair. The only
notable features here are treating a back-end running without inference availability as a normal value, and explicit sign-out clearing.

The intended usage is through `useInferenceKey()` by matching on the resource's `state` / `tag` to distinguish `loading`, `ready`, `unavailable`, and `errored`.

## Tests

Add `src/api/inference_rpc.test.ts`, an integration test for the `get_inference_key` RPC. It is gated by `VITE_RUN_INFERENCE_TESTS=1` so it runs locally (with a stood-up backend + DB + OPENROUTER_PROVISIONING_KEY) but not in CI. There is future work here to discuss how, if at all, we would want this to run in the CI.

I have verified this manually, to obtain an `OPENROUTER_PROVISIONING_KEY` login/signup for a personal account and create one at https://openrouter.ai/settings/management-keys .

Run with: `VITE_RUN_INFERENCE_TESTS=1 pnpm test`

### DIFF
--- a/packages/frontend/src/api/inference_rpc.test.ts
+++ b/packages/frontend/src/api/inference_rpc.test.ts
@@ -1,0 +1,63 @@
+import { type FirebaseOptions, initializeApp } from "firebase/app";
+import { deleteUser, getAuth, signInWithEmailAndPassword, signOut } from "firebase/auth";
+import invariant from "tiny-invariant";
+import { afterAll, assert, describe, test } from "vitest";
+
+import { initTestUserAuth } from "../util/test_util.ts";
+import { createFetchWithAuth, createRpcClient, unwrap, unwrapErr } from "./rpc.ts";
+
+// These tests exercise the live `get_inference_key` RPC, which creates
+// OpenRouter child keys when a provisioning key is configured. They are skipped
+// unless `VITE_RUN_INFERENCE_TESTS=1` is set, so they run locally but not in
+// CI. To run them, ensure you have a backend running; they work both with and
+// without `OPENROUTER_PROVISIONING_KEY` set.
+const enabled = import.meta.env.VITE_RUN_INFERENCE_TESTS === "1";
+
+const serverUrl = import.meta.env.VITE_SERVER_URL;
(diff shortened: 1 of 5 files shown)


packages/frontend/src/api/inference_rpc.test.ts
```diff
@@ -0,0 +1,63 @@
...
+import { afterAll, assert, describe, test } from "vitest";
+
+import { initTestUserAuth } from "../util/test_util.ts";
+import { createFetchWithAuth, createRpcClient, unwrap, unwrapErr } from "./rpc.ts";
+
+// These tests exercise the live `get_inference_key` RPC, which creates
+// OpenRouter child keys when a provisioning key is configured. They are skipped
+// unless `VITE_RUN_INFERENCE_TESTS=1` is set, so they run locally but not in
+// CI. To run them, ensure you have a backend running; they work both with and
+// without `OPENROUTER_PROVISIONING_KEY` set.
```
kasbah, line 13: Why not run them in CI without `OPENROUTER_PROVISIONING_KEY`?

tslil-topos: we could, they exercise very little in that case (handling a 503), but the whole surface area of this change is tiny either way

tslil-topos: i'll change this in the next PR that is based on-top of this one, where the 503 actually does something impactful (disables inference)

epatters (approved):

PR #1345, opened 2026-07-16
OBEE-12 FE: chat and context execution functionality for inference

tslil-topos:
This change depends on #1350 

## Design decisions

* Represent inference context as OpenAI user, assistant, and tool messages. `runOpenAIChatTurn` treats this as an ephemeral request transcript and returns the newly generated assistant/tool-message delta; persistence is delegated to its caller.

* Expose only one tool, `contextExec`, which executes JavaScript with a supplied set of local bindings. Code runs in an `AsyncFunction`; it does not close over application locals, but is not a sandbox.

## Implementation details

This adds an OpenRouter client and `runOpenAIChatTurn`, which prepends the system prompt, runs the tool loop, and returns the generated assistant and tool-message delta. We also introduce `contextExec` and its scope and result types. While not a sandbox, this runs JS code with access to particular things (though `globalThis` etc. are still there). Successful return values are stringified, while syntax and runtime errors are returned as tagged errors.

Conversation-specific transcript derivation, persistence, and application scope are left to callers.

## Tests

Add unit tests for context execution and live chat integration tests using the free `openai/gpt-oss-20b:free` model.

Changes previous integration tests for inference key to run under all circumstances. Backends without inference configured are expected to return 503; follow-up behaviour assertions are skipped.

### DIFF
--- a/packages/frontend/src/inference/chat.ts
+++ b/packages/frontend/src/inference/chat.ts
@@ -1,0 +1,136 @@
+import OpenAI from "openai";
+import type {
+    ChatCompletionAssistantMessageParam,
+    ChatCompletionMessageParam,
+    ChatCompletionSystemMessageParam,
+    ChatCompletionToolMessageParam,
+    ChatCompletionUserMessageParam,
+} from "openai/resources/chat/completions";
+
+import { type ContextExecScope, type EvalResult, contextExec } from "./context_exec";
+
+const OPENROUTER_BASE_URL = "https://openrouter.ai/api/v1";
+const MODEL = "z-ai/glm-5.2";
+
+/** Maximum model completions, including tool-use continuations, in one user turn. */
+export const MAX_CHAT_COMPLETIONS_PER_TURN = 10;
+
+const EMPTY_FILES = Object.freeze(Object.create(null)) as Readonly<Record<string, string>>;
+
+const SYSTEM_PROMPT =
+    "You are an assistant embedded in CatColab. Use `contextExec` when you need to inspect, compute with, or act on the current CatColab context. It executes JavaScript with the available context values as local bindings. The read-only `files` binding maps available filenames to their UTF-8 content; use `Object.keys(files)` to inspect its keys. Use `return` when you need to observe a value; `await` can be used directly. Only use bindings and APIs explicitly described as available. Answer the user's request clearly and concisely, using tool results when relevant.";
+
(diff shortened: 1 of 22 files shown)


epatters:
Naive question from someone who has never done this stuff: why use the OpenAI API instead of OpenRouter's own client SDK?

tslil-topos:
> Naive question from someone who has never done this stuff: why use the OpenAI API instead of OpenRouter's own client SDK?

good question, this deserved to have been in the design decision, apologies.

* openai because it does the minimum we need: multi-turn chats with tools
* openrouter bare sdk would see us implementing this logic from scratch + couples us to openrouter afaict?
* openrouter agents sdk would suffice, but pulls in zod and we really have only one tool so that seemed too heavy

other frameworks were discounted for similar tradeoffs (large bundle size, shipping features we don't care for, or too simplistic). I'm not an expert here --- especially when it comes to typescript + Solid---, i think i only put in about an hour looking at the options so i'd be happy to take things under advisement if you had something in mind.

packages/frontend/src/inference/chat.ts
```diff
@@ -0,0 +1,124 @@
...
+    return content;
+}
+
+type ParsedContextExecArgs = { tag: "Ok"; code: string } | Extract<EvalResult, { tag: "Err" }>;
+
+function parseContextExecArgs(rawArgs: string): ParsedContextExecArgs {
+    try {
+        const value: unknown = JSON.parse(rawArgs);
+        if (
+            typeof value !== "object" ||
```
tslil-topos: this is quite awkward, i reworked it here https://github.com/ToposInstitute/CatColab/pull/1351/changes#diff-db9c40619a3b4d4858a45dda3d7ff5dd3c5cdd063b543543e5458caeeebe06feR126

tslil-topos:
in #1351 i refactored some of this code, i wasn't able to quite correctly plan the shape of things (and my understanding of TypeScript improved somewhat). Apologies if that makes review more complicated

kasbah:
Does it make sense to roll this into #1351 and review as one?

tslil-topos:
It was unfair of me to leave things in this state, i'm busy rewriting history so that we can pretend that the design was the way we wanted it all along

tslil-topos:
i re-ordered the prs, and shuffled history, it should now read coherently: #1353 -> #1345 -> #1351

epatters (approved):
Thanks for explaining the rationale behind the choice of API. Looks good to me!

PR #1346, opened 2026-07-17
OBEE-14 Action API: Initial tests for model notebooks of schema & olog

kasbah:
These are machine ported from RFC-0006 covering the documents package and the `SimpleOlog` and `SimpleSchema` logics. I included `PetriNet` because I had a lot of `PetriNet` examples in the RFC. No other logics or analyses notebooks expected to be implemented for these to pass. I expect we may want to tweak these here and there as we implement. 

I did confirm these _can_ pass, by letting them run against my very rough initial implementation that is not included here, but they are not only not run but also marked as skipped. Once we have some basic implementation going we can actually run them and unskip feature by feature.

The main thing to review here is probably the folder structure and naming for the new packages and just to glance over the tests to see if anything jumps out. I don't expect anyone to read all the tests but they should all look familiar from the RFC.

### DIFF
--- a/.oxfmtrc.json
+++ b/.oxfmtrc.json
@@ -11,7 +11,7 @@
         "customGroups": [
             {
                 "groupName": "workspace",
-                "elementNamePattern": ["catcolab-ui-components", "catcolab-api", "catlog-wasm", "catcolab-document-types", "catcolab-document-methods"],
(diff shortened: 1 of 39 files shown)


tslil-topos:
i didn't read any of the tests, but the top-level organisation makes sense to me :+1:

tslil-topos (approved):
We discussed this pairwise (Evan and i; Kaspar and i): treating these like a sealed box in this corner of the repo seems fair as long as we understand that we're not transferring the responsibility for maintenance to the team.

If something goes wrong or we change our minds about policy/behaviour, we're empowered to delete the offending tests instead of investing time in understanding and adjusting them.

PR #1349, opened 2026-07-20
OBEE-17 Action API: Property based tests for model notebooks

kasbah:
Based on #1346 

This implements property based tests, generating "arbitrary" edits that are executed on existing notebooks retrieved via a DB dump from next. 

"arbitrary" in quotation marks because we categorise them into edits that should preserve the validity of the models and ones that are destructive.

### DIFF
--- a/infrastructure/scripts/dump-model-fixtures
+++ b/infrastructure/scripts/dump-model-fixtures
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+# Dump all active model documents from a CatColab deployment.
+#
+# Usage: dump-model-fixtures <ssh-target> <document-server> <output-file>
+# Example: dump-model-fixtures catcolab@backend-next.catcolab.org next.catcolab.org dumps/next-model-fixtures.json
+
+set -euo pipefail
+umask 077
+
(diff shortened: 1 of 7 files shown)


tslil-topos (approved):
If our criteria for acceptance here are like the other tests (if something is wrong/not working, then team members are not saddled with spending large amounts of time understanding these tests) then this seems sensible to me.

It does however appear that dump-model-fixtures was a bit of a one-off script? Are we likely to use this again in the future? If not, or it was the result of a prompt, i don't think we should be keeping this around.

kasbah:
The script is LLM generated but needs to be run before running the test. The test operates on the dump but we likely don't want to commit a DB dump as a fixture.

I adapted this from the approach on https://github.com/ToposInstitute/CatColab/pull/1075 (which was never merged due to the existing TS API being bad for writing tests like that).

tslil-topos:
@kasbah and i had a conversation about the future of this kind of property based testing. I think we should proceed as-is for now, but when it comes to wiring up the CI we may wish to re-asses. I'll open a separate issue to track the synthetic data gen proposal.

PR #1351, opened 2026-07-22
OBEE-24 LLMConversation document connected to inference in FE

tslil-topos:
This change depends on #1350 and #1345.

## Context

#1350 introduced the `LLMConversation` document type and its interaction schema. #1345 introduced inference in the front end. This change supplies the frontend machinery that makes that document usable as an inference history. We stop (hopefully) right at the boundary of UI/UX questions for this functionality.

## Design decisions

* The Automerge `LLMConversation` document is the sole source of truth. We do not store an OpenAI transcript alongside it. When a turn runs, the frontend derives a provider transcript and its `files` binding from the current document, combines that with a caller-supplied `contextExec` scope, sends the ephemeral request to OpenRouter, and writes the completed assistant/tool output back to the document. These two directions may be found in:
    + `src/inference/llm_conversation.ts` document -> OpenAI conversion
    + `src/llm_conversation/document.ts` OpenAI -> document conversion

* The upcoming UI has one principal entry point: `runLLMConversationTurn`. It accepts, among other things, an optional streaming callback for UI purposes. A failed inference leaves its submitted user message in the document, allowing a later UI to offer retry.

* Inline files are decoded into a cumulative, read-only `files` map available to the JS tool. This allows the LLMs to perform non-trivial computation and inspection using JS. Questions about file size caps and enforcement are deferred to future UI layers.

* Starting a new user turn automatically records unresolved feedback requests as rejected, and the normal derived transcript informs the model of that rejection before the new message.

* Where possible errors are treated as first-class values, instead of state, so that the UI doesn't have to worry about catching and recovering. In particular unavailable inference, deleted conversations, failed requests, and malformed generated output are returned as values.

## Implementation details

* Add live conversation creation/loading alongside the existing model and diagram document modules.

## Tests

Add turn-level tests using in-memory Automerge and a mocked OpenAI boundary. They verify that a completed user/tool/assistant turn is persisted in the conversation document, and that a failed inference still retains the submitted user message.

### DIFF
--- a/packages/frontend/src/inference/llm_conversation_adapter.ts
+++ b/packages/frontend/src/inference/llm_conversation_adapter.ts
@@ -1,10 +1,10 @@
 import type { LLMConversationDocument } from "catcolab-document-methods";
-import type { FileType, UserMessage } from "catcolab-document-types";
+import type { UserMessage } from "catcolab-document-types";
 import { assertExhaustive } from "../util/assert_exhaustive.ts";
 import type { OpenAITranscript } from "./chat.ts";
 
 /** Files available to `contextExec`, indexed by their filenames. */
-export type ConversationFiles = Readonly<Record<string, string>>;
+export type ConversationFiles = Readonly<Record<string, string | number[]>>;
 
 /** The OpenAI request data derived from a persisted LLM conversation. */
 export type LLMConversationInferenceContext = {
@@ -17,14 +17,17 @@ export function prepareLLMConversationInference(
     conversation: LLMConversationDocument,
 ): LLMConversationInferenceContext {
     const transcript: OpenAITranscript = [];
-    const files: Record<string, string> = Object.create(null) as Record<string, string>;
+    const files: Record<string, string | number[]> = Object.create(null) as Record<
+        string,
+        string | number[]
+    >;
 
     for (const interaction of conversation.interactions) {
         switch (interaction.tag) {
             case "user-message":
                 transcript.push({ role: "user", content: userMessageToOpenAIContent(interaction) });
                 for (const file of interaction.files) {
-                    files[file.filename] = decodeFile(file.fileType, file.content);
+                    files[file.filename] = decodeFile(file.mediaType, file.content);
                 }
                 break;
             case "llm-message":
@@ -58,7 +61,8 @@ export function prepareLLMConversationInference(
                     case "unresolved":
                         transcript.push({
                             role: "system",
-                            content: "The user rejected your proposed changes.",
+                            content:
+                                "The user has a follow up message and so your proposed changes were not applied.",
                         });
                         break;
                     case "approved":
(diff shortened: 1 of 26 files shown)


tslil-topos:
to test this functionality i had an llm slopcode a UI

image

---

image

---

image

---

and it looks like this pr chain is sufficient basis for MVP requirements around the UI.

packages/frontend/src/inference/llm_conversation_adapter.ts
```diff
@@ -0,0 +1,108 @@
...
+                            content: "The user rejected your proposed changes.",
+                        });
+                        break;
+                    case "approved":
+                        transcript.push({
+                            role: "system",
+                            content: "The user accepted your proposed changes.",
+                        });
+                        break;
+                    case "rejected":
```
epatters, line 74: A small matter, but since the bodies for the "unresolved" and "rejected" cases are identical, it would be briefer and IMO clearer of intent to restructure the conditional to avoid the duplication.

You might also add a comment explaining why they're identical. This is mentioned in the PR text but is a possibly confusing aspect of the code.

packages/frontend/src/inference/llm_conversation_adapter.ts
epatters, line 1: What's the organizational principle behind having a folder `inference` with `llm_conversation` and a separate folder `llm_conversation`? I find myself confused by this.

packages/frontend/src/llm_conversation/document.ts
```diff
@@ -0,0 +1,203 @@
...
+export async function getLiveLLMConversation(
+    refId: Uuid,
+    api: Api,
+    models: ModelLibrary<Uuid>,
+): Promise<LiveLLMConversationDoc> {
+    const { liveDoc, docRef } = await api.getLiveDoc<LLMConversationDocument>(
+        refId,
+        "llmconversation",
+    );
+    const liveModel = await models.getLiveModel(liveDoc.doc.llmConversationOf._id);
```
epatters, line 63: Do we need the `LiveModelDoc` or just the `LiveDoc` for the model? The main difference is that the former provides an accessor to the elaborated model. So the question is whether we're going to pass a serialization of the complete elaborated model to the LLM, or just the (unelaborated) top-level model document. (I don't see either happening here.)

An advantage to *not* passing the elaborated model is that we could avoid coupling to the `ModelLibrary`. A disadvantage is that if the model (schema) is built compositionally, the LLM wouldn't see what is being instantiated.

epatters (commented):
Nice! Just a few questions below.

packages/frontend/src/inference/llm_conversation_adapter.ts
tslil-topos: thanks, i probably could have chose a better naming scheme:
* `src/llm_conversation/` is trying to follow the same model as other document types (model, for example)
* `src/inference/` is stuff that is related to inference and provider/library specific

so the intent of `src/inference/llm_conversation.ts` was "adapt llm_conversation for inference purposes". I'll rename it to `llm_conversation_adapter.ts` for now, happy to reconsider this structuring.

packages/frontend/src/inference/llm_conversation_adapter.ts
tslil-topos: thanks, i think instead of keeping them identical in this case i've amended the one to have slightly more information for the llm.

packages/frontend/src/llm_conversation/document.ts
tslil-topos: This is not something i know the answer to, i'm happy to pick either one. What we show to the llm, and how, i think is intended to be covered by the upcoming action API. Whether we want more, or what the correct surface for that API is not something i know. 

How would you design this part?

kasbah: My understanding for when we use the action API for this is to pass the LLM a `Notebook` (using the RFC terminology). The LLM then has the option to call `Notebook.validate()` to get the elaborated model.

I don't have opinions on how to handle it right now before the action API is done.

packages/frontend/src/inference/llm_conversation_adapter.ts
epatters: Thanks, that structure makes sense, and the rename to `llm_conversation_adapter` already helps clarify.

packages/frontend/src/llm_conversation/document.ts
epatters: Thanks Kaspar, in that case, I'll just regard what's currently here as a stub to be revisited when the action API is ready.

epatters (approved):

PR #1352, opened 2026-07-22
OBEE-18 Action API: Implementation for basic model notebooks 

kasbah:
Based on #1349 

This initial implementation assumes a simplified binder and we manipulate the document objects directly. It is mainly laying down the path of the types plumbing and the `defineObject`, `defineMorphism` and `defineShape` helpers.


packages/documents/src/notebook.ts
```diff
@@ -64,47 +64,30 @@ export function notebookFromDocument<S extends Shape>(
...
-                        name: object.label ?? "",
-                        obType: type.obType,
-                    },
-                });
-                return objectHandle(document, cellId, type as ObjectTypes<S>) as AddedCell<S, T>;
+                const judgment = Model.newObjectDecl(type.obType);
+                judgment.name = object.label ?? "";
+                const cell = Nb.newFormalCell<ModelJudgment>(judgment);
+                Nb.appendCell(document.notebook, cell);
+                return objectHandle(document, cell.id, type as ObjectTypes<S>) as AddedCell<S, T>;
```
kasbah: I am wondering about whether using `document-methods` for this is an improvement.

packages/documents/src/util/assert_exhaustive.ts
```diff
@@ -0,0 +1,9 @@
+export function assertExhaustive(value: never): never {
```
tslil-topos, line 1: would be good to eventually remove the copy from the front-end (which will have to depend on `documents` for the api) and replace it with this?

kasbah: we probably don't want to expose a util like this from `catcolab-documents`. we will probably need to make a `catcolab-utils` package eventually.

tslil-topos (approved):
this has some fairly sophisticated type wizardry that i have not been able to understand, but having built on-top of this a few times now i feel that the rest is very reasonable and a good start.

kasbah:
I would be interested to talk through the types with you to test if I actually understand them and make sure we are keeping them as simple as possible while still implementing the spec.

tslil-topos:
> I would be interested to talk through the types with you to test if I actually understand them and make sure we are keeping them as simple as possible while still implementing the spec.

Sure thing, let's make time tomorrow. I suppose my principal interest in this API is how it appears from JavaScript. We could imagine users eventually writing code in TypeScript for it, but that doesn't offer the dynamic pluggability of JS, so i'm perhaps less motivated to have the type system do clever things to help with TypeScript compile time.

PR #1353, opened 2026-07-23
OBEE-22 new LLMConversation document type

tslil-topos:
## Design decisions

For the complete design please see [outcome on notion](https://app.notion.com/p/toposinstitute/New-LLMConversation-document-type-39169ce79864800d8958d737832d5c46).

## Brief overview

* Add `LLMConversation` as a v2 document type. Each conversation belongs to one model (in the CatColab sense) through a required `conversationOf` link and fixes one OpenRouter model for the conversation.

* Store interactions as one ordered sequence with stable UUIDs. The supported interactions are user messages, LLM messages, `contextExec` executions, and user feedback requests (for future transaction semantics).

* Store inline file content as bytes and use a closed `FileType` enum, currently containing only `CSV`. Follow-up frontend work will match on this type and decode CSV content as UTF-8 when constructing the user message.

* Don't commit to transaction proposal semantics, it's currently opaque JSON values.

## Implementation details

Add the `llmconversation-of` link type and the `llmconversation` document discriminator. The Rust schema is in the `v2/llm_conversation.rs` module (with autogen ts bindings).

Add an `LLMConversationDocument` alias and minimal constructor to `document-methods`. Existing exhaustive document handling is updated only where required by the expanded union. This includes teaching some things that "not analysis" is no longer the correct way to determine acceptable document types.

UI/UX changes are not in-scope.

## Tests

I didn't have too much imagination here, so there's only a test which ensures that a conversation is indexed as depending on its parent model.

### DIFF
--- a/packages/document-methods/src/llm_conversation.ts
+++ b/packages/document-methods/src/llm_conversation.ts
@@ -14,15 +14,15 @@ import { currentVersion } from "catcolab-document-types";
 /** A document containing an LLM conversation attached to a model. */
 export type LLMConversationDocument = Document & { type: "llmconversation" };
 
-/** Create an empty LLM conversation for a model. */
+/** Create an empty LLM conversation for a document. */
 export const newLLMConversationDocument = (
-    modelRef: StableRef,
+    documentRef: StableRef,
     llmModel: string,
 ): LLMConversationDocument => ({
     name: "",
     type: "llmconversation",
     llmConversationOf: {
-        ...modelRef,
+        ...documentRef,
         type: "llmconversation-of",
     },
     llmModel,
--- a/packages/document-methods/src/llm_conversation.ts
+++ b/packages/document-methods/src/llm_conversation.ts
@@ -1,0 +1,98 @@
+import { v7 } from "uuid";
+
+import type {
+    EvalResult,
+    FeedbackResolution,
+    InlineFile,
+    Document,
+    LLMInteraction,
+    StableRef,
+    Uuid,
+} from "catcolab-document-types";
+import { currentVersion } from "catcolab-document-types";
+
+/** A document containing an LLM conversation attached to a model. */
+export type LLMConversationDocument = Document & { type: "llmconversation" };
+
+/** Create an empty LLM conversation for a model. */
+export const newLLMConversationDocument = (
+    modelRef: StableRef,
+    llmModel: string,
+): LLMConversationDocument => ({
+    name: "",
+    type: "llmconversation",
+    llmConversationOf: {
+        ...modelRef,
+        type: "llmconversation-of",
+    },
+    llmModel,
+    interactions: [],
+    version: currentVersion(),
+});
+
+/** Create an interaction containing a user message. */
+export function newUserMessage(
+    content: string,
+    files: InlineFile[],
+): Extract<LLMInteraction, { tag: "user-message" }> {
+    return {
+        tag: "user-message",
+        id: v7(),
+        timestamp: new Date().toISOString(),
+        content,
+        files,
+    };
+}
+
+/** Create an interaction containing a completed LLM message. */
+export function newLLMMessage(content: string): Extract<LLMInteraction, { tag: "llm-message" }> {
+    return {
+        tag: "llm-message",
+        id: v7(),
+        timestamp: new Date().toISOString(),
+        content,
+    };
+}
+
+/** Create an interaction containing one completed `contextExec` call. */
+export function newLLMCodeExecution(
+    toolCallId: string,
+    code: string,
+    result: EvalResult,
+    transaction?: unknown,
+): Extract<LLMInteraction, { tag: "llm-code-execution" }> {
+    return {
+        tag: "llm-code-execution",
+        id: v7(),
+        timestamp: new Date().toISOString(),
+        toolCallId,
+        code,
+        result,
+        ...(transaction === undefined ? {} : { transaction }),
+    };
+}
+
+/** Append an interaction to a conversation's ordered history. */
+export function appendLLMInteraction(
+    conversation: LLMConversationDocument,
+    interaction: LLMInteraction,
+) {
+    conversation.interactions.push(interaction);
+}
+
+/** Resolve a pending feedback request by its stable interaction ID. */
+export function resolveUserFeedbackRequest(
+    conversation: LLMConversationDocument,
+    requestId: Uuid,
+    resolution: Exclude<FeedbackResolution, "unresolved">,
+) {
+    const request = conversation.interactions.find(
+        (interaction): interaction is Extract<LLMInteraction, { tag: "user-feedback-request" }> =>
+            interaction.tag === "user-feedback-request" && interaction.id === requestId,
+    );
+    if (!request || request.resolution !== "unresolved") {
+        return false;
+    }
+    request.resolution = resolution;
+    return true;
+}
--- a/packages/document-types/src/v2/document.rs
+++ b/packages/document-types/src/v2/document.rs
@@ -4,6 +4,7 @@ pub use crate::v1::DocumentType;
 
 use super::analysis::Analysis;
 use super::api::Link;
+use super::llm_conversation::LLMConversationDocumentContent;
 use super::notebook::Notebook;
 
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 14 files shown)


packages/document-types/src/v2/llm_conversation.rs
```diff
@@ -0,0 +1,115 @@
...
+    UserFeedbackRequest(UserFeedbackRequest),
+}
+
+/// A sequential conversation attached to a CatColab model.
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct LLMConversationDocumentContent {
+    pub name: String,
+    #[serde(rename = "conversationOf")]
+    pub conversation_of: Link,
```
kasbah, line 103: I am wondering about how we can relate (multiple) instances of a schema to the LLM interaction. In my mockups it seemed useful to have the LLM edit the schema and the instance at the same time.

packages/document-types/src/v2/document.rs
```diff
@@ -4,6 +4,7 @@ pub use crate::v1::DocumentType;
 
 use super::analysis::Analysis;
 use super::api::Link;
+use super::llm_conversation::LLMConversationDocumentContent;
```
kasbah, line 7: The humanist in me hates that a "conversation" in the CatColab codebase is becoming an LLM interaction. I realise that it's not a technical point at all but I would be happier with "LLM interaction" naming throughout.

packages/document-types/src/v2/llm_conversation.rs
tslil-topos: the intention (in the design doc) is to have this be possible by traversing the links: we know the parent model, so we can walk its children too.

packages/document-types/src/v2/document.rs
tslil-topos: I don't see why a conversation partner needs to be human, i think that's unreasonably restrictive. I have conversations with my dog, i am perfectly happy with this naming scheme.

kasbah: True conversations are an exchange of ideas with living things (like your dog) but my objection isn't that. I don't think "conversation" is technically the wrong name: we do simulate conversations with LLMs.

What saddens me about this is that when we now say "conversation" in the CatColab codebase we will mean "LLM conversation". So we are adding robot conversations before we ever got to add real conversations in some way.

tslil-topos: For what its worth nothing references "conversation", everything calls this "llmconversation". That is not suggestive of exclusivity or priority.

> So we are adding robot conversations before we ever got to add real conversations in some way.

this is true, the MVP calls for this and we will deliver on that first. Perhaps the next scoped deliverable could involve a live chat between simultaneous users; i think it would be great to learn your ideas on this topic.

kasbah: Let's discuss offline. Just on this specific point:

I should have probably highlighted the `conversation-of` field but I had a different (already resolved) point about that. Since it seems the document is specific to LLMs maybe we can at least rename that field in the hopes we could ever eventually have a real  `conversation-of` a model between people.

packages/document-methods/src/llm_conversation.ts
```diff
@@ -0,0 +1,98 @@
...
+    Uuid,
+} from "catcolab-document-types";
+import { currentVersion } from "catcolab-document-types";
+
+/** A document containing an LLM conversation attached to a model. */
+export type LLMConversationDocument = Document & { type: "llmconversation" };
+
+/** Create an empty LLM conversation for a model. */
+export const newLLMConversationDocument = (
+    modelRef: StableRef,
```
epatters: While your design writeup explains the reasonable choice to assume for now that LLMs conversations will always be attached to a model (and any instances thereof), I do wonder what is gained by writing `modelRef` here and `model` in the docstring, when nothing in the data type requires that the target of `llmConversationOf` be a model. Perhaps just call this `docRef` instead? As a side benefit, there will be less risk of confusing CatColab models and LLM models.

packages/document-types/src/v2/llm_conversation.rs
```diff
@@ -0,0 +1,115 @@
...
+use serde_json::Value;
+use tsify::Tsify;
+use uuid::Uuid;
+
+/// A supported inline file type.
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub enum FileType {
+    CSV,
+}
```
epatters: Any reason to restrict the file type at the type level? I feel like the more common approach is to specify file extensions and/or MIME types at the data level. See e.g. packages for dealing with MIME types in Python or Rust.

packages/document-types/src/v2/llm_conversation.rs
```diff
@@ -0,0 +1,115 @@
...
+
+/// A file stored inline with a user message.
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct InlineFile {
+    pub filename: String,
+    #[serde(rename = "fileType")]
+    pub file_type: FileType,
+    pub content: Vec<u8>,
```
epatters, line 15: Should the file content get base64 encoded? I believe serde can be configured to do that automatically.

packages/frontend/src/model/model_library.ts
```diff
@@ -311,6 +311,9 @@ function elaborateAndValidateModel(
 
 /** Does the patch to the model document affect its formal content? */
 function isPatchToFormalContent(doc: Document, patch: Patch): boolean {
+    if (doc.type !== "model") {
+        return false;
+    }
```
epatters, line 316: Sorry you had to encounter this rather horrible code :) I haven't figured out how to make it better.

epatters (commented):
> ## Design decisions
> 
> For the complete design please see [outcome on notion](https://app.notion.com/p/toposinstitute/New-LLMConversation-document-type-39169ce79864800d8958d737832d5c46).

Thanks for this exceptionally clear and complete writeup!

As a meta comment, it seems a bit of shame to lock it away in Notion, where it will then be lost forever, but OTOH perhaps it is intended to ephemeral.

Code looks great. Just a few small questions/comments.

packages/document-methods/src/llm_conversation.ts
tslil-topos: thanks, for spotting this, that's a good point, i confused intent with implementation! Only confusion lies this way, the current usage patterns enforce this relationship, no need for poor names.

packages/document-types/src/v2/llm_conversation.rs
tslil-topos: That's true, originally i had in mind that we would dispatch different decoders for these when it came to making the files available in the contexExec tool for the llm. I think my training betrayed me here and i reached for a switch on type data.

I don't mind having this be runtime information, i'll make that refactor.

packages/document-types/src/v2/llm_conversation.rs
tslil-topos: What would be the motivation for this? Unfortunately i think i omitted this in the design document, [but here](https://github.com/ToposInstitute/CatColab/pull/1351/changes#diff-12a011d5d89d4d2777081a8fd0d467debcc2b3464f1145be273e719221470c8cR103) you can see the intended usage of this: we do some transformation to put this into a variable in context of the execution environment. 

One could imagine replacing this simple byte decode with turning a csv into a dataframe-like object (or some arrays, or whatever the case is), and for more complicated files "doing something reasonable". From that point-of-view, adding b64 to the on-disk format doesn't seem to be gaining much, but it's quite possible i've missed something important here.

epatters: It's likely not very important, but my thinking was that it is much more space efficient to store byte data as base64 than array-of-numbers when serializing JSON data, and I believe base64 is the standard convention. OTOH, it's possible that this isn't relevant to us because under the hood Automerge is doing something clever to compress the data. I don't know; maybe @kasbah does.

Moreover, our backend already caps the size of a single document (IIRC, at ~5 MB), so there's no risk of blowing up the DB with massive files. It's just a question of how much data the user gets to pack into file attachments.

Anyway, I'm happy to approve as is and defer to your judgment on how to handle this.

epatters (approved):

PR #1354, opened 2026-07-24
Motif finding for CLDs with indeterminates

### LINKED ISSUE #237: Loop finding for causal loop diagrams with indeterminates
Sequel to #206. Now there are *three* kinds of loops:

- reinforcing (positive)
- balancing (negative)
- indeterminate (no definite polarity)

This should be a straightforward job starting from the loop finding already implemented for ordinary causal loop diagrams.

tim-at-topos:
Closes #237. Nothing fancy happening here!


epatters (approved):
Looks great, thanks for knocking this out Tim!

PR #1359, opened 2026-07-29
OBEE-26 FE: add user setting for opt-in LLM capabilities

tslil-topos:
## Caveat emptor

I don't really understand Solid/JSX yet. I tried to follow existing codebase conventions, but the mapping of my intent onto code may be imperfect here. In particular, the UI is ugly. I leave that to someone more capable.

## What this achieves

image
image

Unless you tick this box, you can't:
* run inference
* see LLMConversation documents in any documents view
* make LLMConversation documents (in forthcoming UI)

OBEE-26 FE: add user setting, opt-in LLM capabilities

## Design decisions

* Add user settings to the backend, available for inspection or mutation over RPC. This is backed by the `users` table following the naming convention `settings_*` for the columns, as well as a struct `UserSettings` intended to hold all settings.

* Add a UserSettings provider to the front-end, with basic getter/setter "patch" functionality. It's intended that all user settings are tracked through this mechanism going forward.

* Introduce LLM capabilities setting. Make it so that inference keys are requested only after the authenticated user's settings have loaded and LLM capabilities are enabled. No policy decision in the backend about inference key endpoint interacting with this setting, this is purely FE work presently.

* Co-opt the "user profile" page into a settings page.

* Keep LLM Conversation documents in raw user state, but hide them from document lists and picker completions while the setting is disabled.

## Tests

Add a frontend behavior test verifying that disabled LLM capabilities prevent inference-key provisioning and hide LLM Conversation documents. Add small backend tests verifying that the RPC works.


packages/frontend/src/user/user_state_provider.tsx
```diff
@@ -34,33 +38,61 @@ export function UserStateProvider(props: { children: JSX.Element }) {
...
+    const updateSettings = (patch: Partial<UserSettings>) => {
+        if (currentDocHandle === null) {
+            return;
+        }
+
+        currentDocHandle.change((doc) => {
+            if (doc.settings === undefined) {
+                doc.settings = {};
+            }
+            Object.assign(doc.settings, patch);
```
tslil-topos: looking around this appears to be the standard idiom for applying shallow mutations. I admit to not really understanding what's going on here, if there's a better way i'd be happy to adopt.

packages/frontend/src/user/profile.tsx
```diff
@@ -7,28 +7,63 @@ import { Button, FormGroup, TextInputField } from "catcolab-ui-components";
...
 }
 
+/** Toggle the user's access to LLM-powered features. */
+function LLMCapabilitiesSetting() {
+    const { settings, isReady, updateSettings } = useUserSettings();
+
+    return (
+        <p>
+            <label>
+                <input
```
epatters: In `ui-components/src/form`, we have a set of reusable components for forms and form fields, including `CheckboxField`, that come with a uniform style and look pretty good.

packages/frontend/src/user/profile.tsx
```diff
@@ -7,28 +7,63 @@ import { Button, FormGroup, TextInputField } from "catcolab-ui-components";
...
+                    type="checkbox"
+                    checked={settings().llmCapabilitiesEnabled}
+                    disabled={!isReady()}
+                    onChange={(evt) =>
+                        updateSettings({
+                            llmCapabilitiesEnabled: evt.currentTarget.checked,
+                        })
+                    }
+                />{" "}
+                <span style={{ display: "inline-block", "vertical-align": "middle" }}>
```
epatters: We generally try to avoid inline CSS in JSX. In this case, it will hopefully suffice to using the already-styled form components (see above).

packages/frontend/src/user/user_state_provider.tsx
```diff
@@ -1,22 +1,26 @@
...
+import { type UserSettings, UserSettingsContext } from "./user_settings_context";
+import { type AppUserState, INITIAL_USER_STATE, UserStateContext } from "./user_state_context";
 
 export function UserStateProvider(props: { children: JSX.Element }) {
     const api = useApi();
     const firebaseApp = useFirebaseApp();
-    const [userState, setUserState] = createStore<UserState>(INITIAL_USER_STATE);
+    const [userState, setUserState] = createStore<AppUserState>(INITIAL_USER_STATE);
+    const [isReady, setIsReady] = createSignal(false);
+    const settings = createMemo<UserSettings>(() => ({
```
epatters: A Solidjs store has fine-grained reactivity, and the `settings` data is just a field in that store, so it shouldn't be necessary, and doesn't feel very idiomatic, to construct this memo derived from the user state. Consumers of the settings can just pull out the fields they need from the user state on demand.

packages/frontend/src/user/user_state_provider.tsx
```diff
@@ -34,33 +38,61 @@ export function UserStateProvider(props: { children: JSX.Element }) {
...
         unsubscribeAuth();
         teardownDocHandle();
     });
 
+    const settingsContext = { settings, isReady, updateSettings };
+
     return (
-        <UserStateContext.Provider value={userState}>{props.children}</UserStateContext.Provider>
+        <UserStateContext.Provider value={userState}>
+            <UserSettingsContext.Provider value={settingsContext}>
```
epatters: Building on the previous comment, on my understanding, you shouldn't need to create a separate context and provider for the settings, since it's already available through the context and provider for the user state.

packages/frontend/src/user/user_state_provider.tsx
```diff
@@ -1,22 +1,26 @@
...
-import { INITIAL_USER_STATE, UserStateContext } from "./user_state_context";
+import { type UserSettings, UserSettingsContext } from "./user_settings_context";
+import { type AppUserState, INITIAL_USER_STATE, UserStateContext } from "./user_state_context";
 
 export function UserStateProvider(props: { children: JSX.Element }) {
     const api = useApi();
     const firebaseApp = useFirebaseApp();
-    const [userState, setUserState] = createStore<UserState>(INITIAL_USER_STATE);
+    const [userState, setUserState] = createStore<AppUserState>(INITIAL_USER_STATE);
+    const [isReady, setIsReady] = createSignal(false);
```
epatters: Is this `isReady` signal necessary? Outside of tests we don't usually wait around for `docHandle.whenReady`.

packages/frontend/src/user/user_state_provider.tsx
epatters: Usually we expose the document handle and call `docHandle.change` with a specific change, i.e., setting some specific field, rather than wrapping the `docHandle.change` with a patch.

epatters (dismissed):
Thanks for making a start on this! Solidjs definitely takes some getting used to because it has lots of idioms (sometimes actual requirements) that aren't enforced by the type system or API but that the programmer is supposed to understand.

I left a few comments below.

A question I have for @kasbah (who should also review this PR :) is whether it's safe to add new fields to the user state solely on the frontend side, without ever persisting them to the DB. I know that the backend regularly updates the user state but I don't recall how destructive that is. My recollection of the design is that the user state is intended to be ephemeral.

packages/frontend/src/user/profile.tsx
tslil-topos: i looked at this, but CheckboxField places the label above the checkbox so it looks very silly for things with long labels. In a "compact" formgroup it's passable, i don't think this is much better than before but i'll leave this in anyway

image

packages/frontend/src/user/user_state_provider.tsx
tslil-topos: thanks, my approach was indeed convoluted here, i'm working on making this and the other simplifications as i come to understand the solid way :tm:

tslil-topos:
thanks for the review Evan! I hope to have this resolved by (your) Monday SOD.

packages/frontend/src/user/user_state_context.tsx
```diff
@@ -14,13 +15,27 @@ export const INITIAL_USER_STATE: AppUserState = {
     documents: {},
 };
 
-export const UserStateContext = createContext<UserState>(INITIAL_USER_STATE);
+type UserStateContextValue = {
+    userState: AppUserState;
+    docHandle: Accessor<DocHandle<AppUserState> | null>;
```
tslil-topos: this is the substantive change here, assuming i understood the feedback we now pass around the docHandle directly for writes to the automerge doc

packages/frontend/src/user/user_state_provider.tsx
```diff
@@ -1,33 +1,31 @@
...
+    );
 
-    let currentDocHandle: DocHandle<AppUserState> | null = null;
     let currentChangeHandler: ((arg: { doc: AppUserState }) => void) | null = null;
     let currentUserId: string | null = null;
 
     const teardownDocHandle = () => {
-        if (currentDocHandle && currentChangeHandler) {
-            currentDocHandle.off("change", currentChangeHandler);
+        const docHandle = currentDocHandle();
```
tslil-topos: turns out that there's some complicated interaction between firebase, react, and oxlint. I see that we have `packages/frontend/oxlint.config.ts`: `customReactiveFunctions:` which is maybe for stuff like this, but i really don't know. I will try to close over what's needed and pass that here instead of using react context.

tslil-topos:
After speaking with @kasbah we agreed that user state was the wrong place for settings like these (and moreover should probably be renamed). 

This change now introduces a backend RPC endpoint for accessing/mutating user settings, and these are backed by db columns in the user's table. We may want the schema to change, but for now it's a single new column.

kasbah (approved):
LGTM 

I opened #1380 and #1381 to improve on this later.

epatters (approved):
Thanks both, glad that we've converged to a better design.

@tslil-topos, there's a merge conflict to resolve, then this should be ready to go.

PR #1364, opened 2026-08-01
Refactor ODE analyses through `ODESemantics` in `catlog`, `catlog-wasm`, and `frontend`

### LINKED ISSUE #1098: RFC-0001: tracking issue

### LINKED ISSUE #1171: Stop using `fmt::Display` for LaTeX output

### LINKED ISSUE #1177: Refactor Lotka–Volterra and Linear ODE analyses (for LaTeX equations)

### LINKED ISSUE #1215: Polynomial ODE equations and simulation both show unassigned variables in compositional models

tim-at-topos:
*For some arcane GitHub reason, #1251 was automatically closed and I cannot reopen it. Then #1363 was a clone with #1319 merged in, but the automatic rebase that GitHub did with its new stacked PR feature messed everything up and broke all the tests. This is a third-time-lucky PR. I'll repeat most of the original PR descriptions below for the sake of the reader.*

_✨ Brand new features in this mega-merged PR, not previously appearing in any other PR ✨_

- So much more documentation and helpful comments
- Tests for things both old and new
- A stacked PR (#1439) with migration for existing ODE analysis documents to the new `v1` so that everybody's documents don't break

---

### PR 1: ODE semantics via contributions

This PR includes a massive redesign of the way that we handle ODE semantics, which has the bonus of introducing some new capabilities (#1177) as well as (hopefully) making adding new ODE semantics in the future considerably simpler.

p.s. I know this PR adds quite a lot of code for something called a "refactor", but (a) I added a lot of documentation and some new tests, and (b) subsequent PRs focussing on the front-end parts should now be possible, allowing for much more code removal.

---

### PR 2: LaTeX things and front-end refactors

Now that more ODE analyses have LaTeX equations, we should be stricter about the use of `fmt::Display` only for debugging.

This PR introduces some new types and traits (`Latex`, `ToLatex`, `ToLatexWithMap`) for dealing with LaTeX expressions and rendering object/morphism names nicely in the frontend. It also includes a bunch of tests for ODE semantics in the wasm layer. Finally, as a bonus, it removes a lot of repetitive code from `catlog-wasm::src::analyses` by defining `ode_semantics_simulation` and `ode_semantics_equations`, enabled by #1251, as well as removing some unnecessary types introduced there (such as `LotkaVolterraEquationsData`, which is just `null`).

Again, this refactor adds a lot of lines, but that's almost entirely due to how many nice new tests it adds 👼 

As a bonus, this PR fixes the following issues:
- #1171
- #1215

---

Although further refactoring is possible (e.g. `lotka_volterra.tsx` and `linear_ode.tsx` share a lot of code, and maybe we should do mass-action for stock-flow diagrams by simply migrating them to Petri nets), I think that the "correct" solution to this would be to start looking at generating that code from the definitions of the analyses rather than manually refactoring it. But this is now a much larger, rather separate project,  so I think it's a bit out of scope.

With that in mind, **I would suggest that this PR finally completes the last item in #1177 , and thus closes the RFC-0001 parent issue #1098.**


packages/catlog/src/stdlib/analyses/ode/ode_semantics.rs
```diff
@@ -0,0 +1,294 @@
...
+// FEEDBACK | (that is, those that come *before* the struct) here and have things all work? I'm still
+// _________/ a bit intimidated by all these `crg_attr(feature = "serde")` bits.
+//
+// #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+// #[cfg_attr(feature = "serde-wasm", derive(Tsify))]
+// #[cfg_attr(
+//     feature = "serde-wasm",
+//     tsify(into_wasm_abi, from_wasm_abi, hashmap_as_object)
+// )]
+pub trait ODESemanticsProblemData<P: ODEParameterType> {
```
tslil-topos, line 253: let's talk about this, but on the surface of things it seems to me that you should be able to do

```rust
pub trait ODESemanticsProblemData<P: ODEParameterType> {
    type EquationsData: ODESemanticsEquationsData;

    const EQUATIONS_DATA: Self::EquationsData;
    const INITIAL_VALUES: &'static [(QualifiedName, f32)];
    const DURATION: f32;

...
```

it's not typically possible to put a hasmap into a const, but it's easy enough to construct a hashmap from slices like this using `collect`.

tslil-topos: It's possible that i'll be able to answer these questions myself with more reading, but taking notes as they jump out at me.

maybe a follow-up question is: must this be a trait?

one could imagine a struct pattern like this

```rust
pub struct ODESemanticsProblemData<E: ODESemanticsEquationsDatat, P: ODEParameterType> {
  equations_data: E,
  initial_values: HashMap<QualifiedName, f32>,
  duration: f32,
  _param_type: PhantomData<P>
}

impl<E: ODESemanticsEquationsDatat, P: ODEParameterType>  ODESemanticsProblemData<E,P> {
  pub fn build_analysis(....) { ... }
}
```

and then we can figure out what `extend_scalars` is supposed to do?

tim-at-topos: i'm so silly, this should definitely be a struct instead of a trait i think

tim-at-topos: OK, I think I see the idea here. This would involve moving more things into the `EquationsData` than there currently are, but maybe that's not a bad idea... however, I think this would break all existing analysis documents, so we would have to finally do some sort of migration on this.

If what I've written isn't too unbearable, then it might actually be a good idea to stick with it.

tim-at-topos: scrub that, after much offline discussion I'm gonna try this

tim-at-topos:
@epatters Apart from actually implementing the migrations (`catlog::src::stdlib::analyses::ode::v1::migrate`) into (I assume, but suggestions and advice also welcome) `frontend/src/analysis/document.ts :: migrateAnalysis`, this is ready for review. Let me know if you would like anything from me that would make this easier to read!

I haven't actually marked it as ready for review because it's failing one check for `solid(reactivity)` but I'm going to chat with Kaspar about this, because the warning suggests that something that does indeed work when I test it should not be working, so it might be a case of an over-eager linter. But the point is that any changes needed to fix any problems should all be far away from the main content of this PR.

PR #1371, opened 2026-08-04
Fix deployment concurrency

kasbah:
I noticed there is a race when deleting "previous" deployments if another job is still running. The fix is to limit the concurrency (to one) and cancel any previous jobs.


epatters (approved):
Thanks!

PR #1372, opened 2026-08-05
Instance editor UI

kasbah:
This is our UI for tabular/data instances using the component from #1452. You should be able to create instances of schemas and ologs to try it out.


tslil-topos (commented):
played around with it, i think we should proceed but noting:
* we possibly have some concurrent edit races. `EditState = { rowId, headerId, text }` is about ids but ` CellPosition = { row, col } ` is about array indices. Looks like these worlds meet in `isEditing` (maybe elsewhere too) but basically at render time i think. So commitEdit calls setField which doesn't look at rowId and could miss a mutation that shifts indices. I think this would be difficult to exercise with LLMConversation, but you could probably do this to yourself editing the schema with in-progress work on the instance.
* we have to figure out what our error story should be for the various cases, eg addRow failing is currently not observable?

kasbah:
Thanks!

>* we possibly have some concurrent edit races. `EditState = { rowId, headerId, text }` is about ids but `CellPosition = { row, col }` is about array indices. Looks like these worlds meet in `isEditing` (maybe elsewhere too) but basically at render time i think. So commitEdit calls setField which doesn't look at rowId and could miss a mutation that shifts indices. I think this would be difficult to exercise with LLMConversation, but you could probably do this to yourself editing the schema with in-progress work on the instance.

I am addressing this on #1452 . See https://github.com/ToposInstitute/CatColab/pull/1452/changes/d670fba9b509f9fea34c09bf726a6c8cbd149a40 


>* we have to figure out what our error story should be for the various cases, eg addRow failing is currently not observable?

True, I would like to defer that, though not for too long.

tslil-topos (approved):
thanks, excited to develop this further

PR #1374, opened 2026-08-06
Fix inconsistency between `cellOrder` and `cellContents` in notebook editor

kasbah:
I noticed this come up while developing.


epatters (approved):
Thanks! Seems like we should be editing `cellOrder` and `cellContents` as part of a single transaction, but perhaps this is a Solid reactivity issue.

kasbah:
Yes, I believe this is caused by Solid rather than Automerge.

PR #1376, opened 2026-08-06
Support `QualifiedName` across instantations in instance document-type

kasbah:
The current instance document type can not reference formal content in instantiated models. Since we don't want to depend on catlog for `QualifiedName` this is the best solution I came up with.

### DIFF
--- a/packages/document-types/src/v2/instance.rs
+++ b/packages/document-types/src/v2/instance.rs
@@ -7,6 +7,7 @@ use uuid::Uuid;
 /// attribute morphism then we provide the value of the type; if the column corresponds to a
 /// mapping morphism then we provide the uuid of the row.
 #[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
 pub enum FieldValue {
     /// Base type: the empty type.
     Null,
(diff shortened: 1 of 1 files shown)


epatters (dismissed):
Thanks for noticing this problem, an oversight in the original PR that I missed in my review.

The `QualifiedName` type was designed to serialize to a bare string in TypeScript, so that the frontend can used qualified names as keys in objects or `Map`s without having to do any conversions (or even know about the underlying assumptions for naming in elaborated models).

It's annoying that with the current factoring we can't reuse the `QualifedName` type in `document-types`. As a workaround, we've been using `String` instead of `Uuid` (or `UuidPath`) and then manually deserializing the QualifiedName in `catlog`. For an example of this, see `document_types::model::Ob::Basic` and the `ob_syn` method in `notebook_elab`.

packages/document-types/src/v2/instance.rs
```diff
@@ -24,20 +26,75 @@ pub enum FieldValue {
...
 }
 
 /// A single table, corresponding to a single entity.
 #[derive(PartialEq, Debug, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi, hashmap_as_object)]
 pub struct Table {
-    /// The uuid of the entity to which this table corresponds.
-    pub id: Uuid,
+    /// The uuid path of the entity to which this table corresponds.
+    pub id: UuidPath,
```
epatters: So, this should be `String` instead of `UuidPath`.

kasbah: Wouldn't it be clearer if we kept this `UuidPath` type and used it in any other places too? It still serializes to a string in Typescript but at least keeps some information on the Rust side.

epatters: Hmm, I guess in that case I'd rather just use `QualifiedName` because it's designed to work for both textual and notebook formats and in the future we want to expose the plain text format to the user as well. The instance editor will get access to an elaborated model and it shouldn't have to think about whether the model was elaborated from plain text or from a notebook.

So the options I'd endorse are:

1. Use a `String` here as proposed above, or:
2. Use a `QualifiedName`, either by refactoring the code so that both `document_types` and `catlog` can depend on `QualifiedName` or by making copy of the data types `document_types`.

Honestly I still prefer (1) because it means that the frontend doesn't have to know *anything* about how naming works in the elaborated data structures, but I'd still prefer to (2) to `UuidPath` because `UuidPath` seems an awkward compromise between (1) and (2).

kasbah: Thanks, I've gone with 1 for now.

tslil-topos (approved):
given that @epatters said that he'd endorse this option, and we're looking to get this in

kasbah:
Thanks, it seems it still needs a re-review from @epatters to meet merge requirements.

tslil-topos:
> Thanks, it seems it still needs a re-review from @epatters to meet merge requirements.

I believe you can dismiss it as stale if you so desire (or i could)

epatters:
Sorry, I meant to follow up sooner. Thanks both!

PR #1377, opened 2026-08-07
Update oxlint to 1.76

kasbah:


epatters (approved):
Nice, good to see these explicit type coercions removed.

PR #1384, opened 2026-08-11
OBEE-28: implement in-memory binder, route all views/changes through that interface

tslil-topos:
The only salient point here is that there does not appear to be a satisfactory way to have `createBinder<Handle>(store?: ...)` overload `createBinder()`, so instead we have `createBinder()` (no args) and `createBinderWithStore<Handle>(store: ...)` both exported.

### DIFF
--- a/packages/documents/package.json
+++ b/packages/documents/package.json
@@ -19,7 +19,8 @@
     "dependencies": {
         "catcolab-document-methods": "link:../document-methods",
         "catcolab-document-types": "link:../document-types/pkg",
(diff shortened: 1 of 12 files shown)


kasbah:
Did you look at enabling some additional tests to cover the added functionality?

tslil-topos:
> Did you look at enabling some additional tests to cover the added functionality?

I was hoping to talk to you about it, i'm not sure which tests those would be or how to go about doing so.

kasbah (approved):

PR #1386, opened 2026-08-12
OBEE-28 Document binder tests

kasbah:
(GH helpfully closed my other PR for me)

@tslil-topos In reply to https://github.com/ToposInstitute/CatColab/pull/1384#issuecomment-5266542986, this is very much LLM assisted test generation so that's how I would do it.

This enables some of the tests and I generated some additional ones for the in-memory store specifically.


tslil-topos:
@kasbah lgtm, thanks. we should add this to the stack i suppose?

tslil-topos (approved):

kasbah:
It seems because this branch was already on the stack with another PR GH is confused about it. I'll just merge this into your branch.

PR #1390, opened 2026-08-13
Update GAIOS deploy

paulsonnentag:
We'd like to add CatColab as a default tool in [gaios.sgai.uk](https://gaios.sgai.uk/). At the moment we check a bundled copy of CatColab into the [gaios.sgai.uk repo](https://github.com/inkandswitch/gaios.sgai.uk/tree/sept-26-tools/public/packages/catcolab), which means it goes stale as soon as you ship a new version. It would be much nicer to hook into your existing deploy workflow so that each CatColab release also uploads a bundle to Netlify that gaios can load directly.

This PR contains:

- A workflow that [builds the gaios bundle and publishes it to Netlify on deploy]
- An update to the CatColab tool so that creating a model also creates an analysis by default, giving you both a model and an analysis in gaios. Previously only models could be created.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -319,6 +319,90 @@ jobs:
           name: rfc
           path: rfc/_site
 
+  build_gaios:
+    name: Build gaios Patchwork plugin
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+        with:
+          # Need full history for `git log` for cache key
+          fetch-depth: 0
+
+      - name: Compute cache key
+        id: cache-key
+        run: |
+          # Get git hash of the most recent commit affecting gaios or any of its dependencies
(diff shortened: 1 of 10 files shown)


paulsonnentag:
Is there a way to get a build preview? It would be nice if we could test that the deployed version is loadable in gaios before we merge this

kasbah:
Probably best if we add you as a contributor but I'll leave that up to @epatters . I copied your branch for now: #1393 

The build passed: https://github.com/ToposInstitute/CatColab/actions/runs/31703563860/artifacts/9182366436

epatters:
Thanks @paulsonnentag! I've invited you to join the repo; once you join, you should get build previews. (Confusingly, the one linked by GitHub above is from Kaspar's copy of the PR, which I've just closed.)

paulsonnentag:
@epatters could you have a look at this PR and let me know if this looks good to you

epatters (approved):
Looks great, thanks for these improvements, Paul!

PR #1391, opened 2026-08-13
Refactor ID fields in tabular instance document-type

kasbah:
Two changes to the instance document type:

1. Remove `id` field from `TableRow`. Since we already store `TableRow` in a hashmap with id keys.
2. ~~Rename the `Table.id` to `Table.entity_id` because, when e.g. working on the TS API, having a `table.id` that is not the table id is confusing.~~ Remove `id` field from `Table` for the same reason.

### DIFF
--- a/packages/document-types/src/v2/document.rs
+++ b/packages/document-types/src/v2/document.rs
@@ -10,7 +10,6 @@ use super::notebook::Notebook;
 
 use serde::{Deserialize, Serialize};
 use tsify::Tsify;
-use uuid::Uuid;
 
 /// This is the content of a model document. For legacy reasons, we reserve
 /// the name "ModelDocument" for `Document & { type: "model" }`.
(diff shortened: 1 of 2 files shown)


tim-at-topos (approved):

kasbah:
An alternative would be to also remove `entity_id` and make the hash map of tables use the schema's entity id as the key. The result would be that tables no longer have their own UUID but only that of the schema's entity but I am struggling to come up with a case where that would be a problem.

tslil-topos:
> An alternative would be to also remove `entity_id` and make the hash map of tables use the schema's entity id as the key. The result would be that tables no longer have their own UUID but only that of the schema's entity but I am struggling to come up with a case where that would be a problem.

This feels like the logical conclusion of the principle "don't add redundant degrees of freedom" when choosing encoding of these structures, and strikes as me as the best outcome. I don't know what it would mean for there to not be a bijection between entities and tables, so i think relying on this makes sense and simplifies the code.

kasbah:
Thanks, I've removed the field and changed the hashmap to be keyed by a string that will be the string encoded `QualifiedName` of the schema entity id.

tslil-topos (approved):
lgtm, thanks!

epatters (approved):
Thanks, I agree that these redundant fields should be removed.

PR #1394, opened 2026-08-13
OBEE-29 add CellKind, iterator/filter to API

tslil-topos:
The current type wizardry is too clever for me to follow or replicate, consequently there are no informative type constraints on these methods. For now our priority is using this from JavaScript, so we can revisit the type system later on to make narrowing and other TypeScript-only, advanced features possible

### DIFF
--- a/packages/documents/src/model/notebook.ts
+++ b/packages/documents/src/model/notebook.ts
@@ -4,6 +4,7 @@ import type { DocumentStore } from "../document-store";
 import type { NotebookDocument } from "../notebook-document";
 import { getRichTextCell, type RichTextCell } from "../rich-text";
 import type {
+    AnyCellType,
     CellTypeOf,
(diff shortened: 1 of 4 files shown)


packages/documents/src/model/notebook.ts
```diff
@@ -51,13 +53,121 @@ type AddedCellOf<S extends Shape, T extends CellTypeOf<S>> = T extends RichTextT
...
+            for (const candidate of shape.morphisms) {
+                if (morphismTypesEqual(candidate.morType, type.morType)) {
+                    return true;
+                }
+            }
+            return false;
+    }
+}
+
+function shapeSupportsShape(shape: Shape, required: Shape): boolean {
```
kasbah, line 87: Thought about this one for a while but I can't come up with a better name either.

kasbah (approved):

PR #1395, opened 2026-08-13
OBEE-31: add simple-schemas to logics

tslil-topos:
It's simple (schema)


kasbah:
I don't think we want an automerge store implementation as part of `catcolab-documents`, automerge should remain a dev dependency for tests. We will put an implementation of an automerge+solid+catcolab-backend store in the frontend.

tslil-topos:
> I don't think we want an automerge store implementation as part of `catcolab-documents`, automerge should remain a dev dependency for tests. We will put an implementation of an automerge+solid+catcolab-backend store in the frontend.

this change now only contains simple-schema in logics

kasbah (approved):

PR #1396, opened 2026-08-14
OBEE-28 Replace `createBinderWithStore` with overload

kasbah:
Referring to the inability to overload nicely in #1384, I randomly realized it should be possible after all.


tslil-topos (approved):
silly typescript

PR #1399, opened 2026-08-17
OBEE-35: getCoreTheory, supportsInstances, validate model notebooks, deletion, pipe elab errors

tslil-topos:
Prior to this catlog-wasm was swallowing elaboration errors, this change surfaces them in a minimal way: through the same error stream as all other validation errors. This does mean that callers in typescript will see a bunch of generic Issue that they'll have to inspect to understand where in the pipeline the model validation failed, but it's more information than we had before. If/when it becomes important to distinguish validation from elaboration we can revisit this.

The same approach was taken for supportsInstances as my previous changes on the API: i don't know how to make the typescript type system do very clever things here, so we rely on runtime errors for the most part. Again my reasoning is that this is our immediate use-case, and likely the most important one for the next while.

I'm aware that there is `elaborateAndValidateModel` but (beside being private) this is operating at the wrong layer with a different intention to our `validateModelDocument` --- in particular we need to generate error paths conforming to some interface. It's possible that later work could move/extract the generic `elaborateModel` + `model.validate` operation from `frontend/src/model/model_library.ts` into catcolab-documents in some way, but for now we have two methods with slightly differing semantics.

Validation subscription is implemented by using the change subscription method of the underlying store, so whenever there are validation listeners we incur one validate call per update regardless of the nature of the update. Future logic could improve this by figuring out how to filter the updates to "semantic altering" ones only, but for now this is probably fine.

### DIFF
--- a/packages/documents/src/model/cell.ts
+++ b/packages/documents/src/model/cell.ts
@@ -1,6 +1,7 @@
 import { Nb } from "catcolab-document-methods";
 import type { Ob } from "catcolab-document-types";
 import type { DocumentStore } from "../document-store";
+import { deleteNotebookCell } from "../notebook-document";
 import { getRichTextCell, type RichTextCell } from "../rich-text";
 import { findMorphismType, findObjectType } from "../shape";
 import type {
@@ -12,7 +13,7 @@ import type {
     ObjectTypesOf,
     Shape,
 } from "../shape";
-import { getModelJudgment, type ModelDocument } from "./document";
+import { tryGetModelJudgment, type ModelDocument } from "./document";
 
 /** Runtime names for the discriminants on notebook cell handles. */
 export const CellKind = {
@@ -25,18 +26,19 @@ export interface ObjectCell<O extends ObjectType> {
     readonly kind: "object";
     readonly id: string;
     readonly type: O;
-    readonly label: string;
+    readonly label: string | undefined;
 
     update(patch: Partial<{ label: string | null }>): void;
+    delete(): void;
 }
 
 export interface MorphismCell<S extends Shape, M extends MorphismType> {
     readonly kind: "morphism";
     readonly id: string;
     readonly type: M;
-    readonly label: string;
-    readonly from: ObjectCell<DomainObjectTypesOf<S, M>> | null;
-    readonly to: ObjectCell<CodomainObjectTypesOf<S, M>> | null;
+    readonly label: string | undefined;
+    readonly from: ObjectCell<DomainObjectTypesOf<S, M>> | null | undefined;
+    readonly to: ObjectCell<CodomainObjectTypesOf<S, M>> | null | undefined;
 
     update(
         patch: Partial<{
@@ -45,6 +47,7 @@ export interface MorphismCell<S extends Shape, M extends MorphismType> {
             to: ObjectCell<CodomainObjectTypesOf<S, M>> | null;
         }>,
     ): void;
+    delete(): void;
 }
 
 export type CellOf<S extends Shape> =
@@ -57,6 +60,7 @@ export function getObjectCell<Handle, O extends ObjectType>(
     handle: Handle,
     cellId: string,
     type: O,
+    onFormalChange: () => void,
 ): ObjectCell<O> {
     return {
         kind: "object",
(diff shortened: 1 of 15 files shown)


packages/documents/src/model/cell.ts
```diff
@@ -25,18 +26,19 @@ export interface ObjectCell<O extends ObjectType> {
...
 }
 
 export interface MorphismCell<S extends Shape, M extends MorphismType> {
     readonly kind: "morphism";
     readonly id: string;
     readonly type: M;
-    readonly label: string;
-    readonly from: ObjectCell<DomainObjectTypesOf<S, M>> | null;
-    readonly to: ObjectCell<CodomainObjectTypesOf<S, M>> | null;
+    readonly label: string | undefined;
```
tslil-topos, line 39: this is my understanding of the RFC's specification that deleting a morphism (say) should render its data undefined

> After deletion, reading fields off the stale handle returns undefined.

there wasn't enough test coverage to figure out exactly what the intended boundary is, but this seems reasonable to me.

packages/documents/src/model/notebook.ts
```diff
@@ -171,19 +174,93 @@ export interface Notebook<S extends Shape, D extends NotebookDocument = Notebook
...
+    function validationResultsEqual(left: Result<DblModel>, right: Result<DblModel>): boolean {
+        if (left.tag !== right.tag) {
+            return false;
+        }
+        if (left.tag === "Ok") {
+            return true;
+        }
+        return (
+            right.tag === "Err" && JSON.stringify(left.content) === JSON.stringify(right.content)
+        );
```
kasbah: - Why are they always equal if the left is "Ok"? I guess this function is handling more logic than the name implies.
- `JSON.stringify` for equality is always a code smell and relies on keys in JS objects staying the same order. There should be a better way.

kasbah:
> A real divergence from the RFC is a simplified onValidate subscription: currently we don't subscribe to the actual document, we sneak triggers into the api method calls. This will need to be assessed in the future, but it suffices for LLM interactions/table editing which must be entirely through the API at present.

I don't think this is acceptable because everything is a collaborative document so changes can come from anywhere. Even in local interactions you will personally be able to edit your schema, not using the API at present.

tslil-topos:
> > A real divergence from the RFC is a simplified onValidate subscription: currently we don't subscribe to the actual document, we sneak triggers into the api method calls. This will need to be assessed in the future, but it suffices for LLM interactions/table editing which must be entirely through the API at present.
> 
> I don't think this is acceptable because everything is a collaborative document so changes can come from anywhere. Even in local interactions you will personally be able to edit your schema, not using the API at present.

acceptable for who or for what? 

In any event, if github wasn't broken i'd edit the description. This was fixed in https://github.com/ToposInstitute/CatColab/pull/1399/commits/d90fdae005d5f4fdf543f5c47d0d1cbfa176b4a3

kasbah:
Acceptable to me, or I hope us, as CatColab devs. I guess that is always the implied voice for my GH comments on this repo.

packages/documents/src/model/validation.ts
```diff
@@ -0,0 +1,119 @@
...
+        case "UnsupportedFeature":
+            return {
+                message: `The model uses an unsupported feature: ${error.content.tag}`,
+            };
+        case "InvalidLink":
+            return {
+                message: `Instantiation \`${generatorName(notebook, error.content)}\` is invalid`,
+                path: generatorPath(notebook, error.content),
+            };
+    }
```
kasbah, line 88: Very nice to see these enumerated!

kasbah (changes requested):
Looks mostly good to me. Separating instance support and validation into two PRs would have made it a bit easier to review.

packages/documents/src/model/notebook.ts
kasbah: We depend on `deep-equal` in other places. On the initial PR I managed to avoid it by implementing model equality [thus](https://github.com/ToposInstitute/CatColab/blob/main/packages/documents/src/model/equality.ts). I think it would be fine to depend on `deep-equal` for convenience though.

tslil-topos: * if left.tag == right.tag and left.tag == Ok then right.tag == Ok and this isn't an "edge transition"
* happy for recommendations?

tslil-topos: I have removed all of this logic following our conversation

kasbah:
Reminder that before you go further with instances it probably makes sense to rebase on #1391

tslil-topos:
> Reminder that before you go further with instances it probably makes sense to rebase on #1391

thanks, is there anything you're looking for on this after i removed the validation "edge change" filtering logic?

kasbah:
rest looks good to me

kasbah:
Maybe you could weigh in on https://github.com/ToposInstitute/CatColab/pull/1391#issuecomment-5328172870 and we can decide on the hashmap table id structure first. I am leaning towards removing the `entity_id`.

kasbah (approved):

PR #1400, opened 2026-08-17
Use `exitForward` when exiting completions

kasbah:
Instead of exiting right. In the list editor we still go to the next list child which is auto-added if there is none.


epatters (approved):
Agreed, thanks!

PR #1401, opened 2026-08-18
OBEE-34: automerge binder in FE

tslil-topos:
Implement an automerge store in the FE operating on refId. This doesn't make the connection to actual backend documents yet, and given that the functionality here is implemented in terms of `frontend/src/api/types.ts: Api` the test follows the convention of mocking this api.

I was not sure what to do with versions or servers, this seemed reasonable to me but it's possible that some missing knowledge would give a better codepath here.
```typescript
32             if (ref.version !== null) {
33                 return {
34                     tag: "Err",
35                     content: [{ message: "Versioned document references are not supported." }],
36                 };
37             }
38             if (ref.server && ref.server !== api.serverHost) {
39                 return {
40                     tag: "Err",
41                     content: [
42                         { message: `Cannot resolve a document from server "${ref.server}".` },
43                     ],
44                 };
45             }
```

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -50,6 +50,7 @@
         "catcolab-api": "link:../backend/pkg",
         "catcolab-document-methods": "link:../document-methods",
         "catcolab-document-types": "link:../document-types/pkg",
+        "catcolab-documents": "link:../documents",
         "catcolab-ui-components": "link:../ui-components",
         "catlog-wasm": "link:../catlog-wasm/dist/pkg-browser",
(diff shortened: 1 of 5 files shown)


kasbah:
We are starting to overlap work now with my #1372 

I had linked you this in another discussion before:
https://github.com/ToposInstitute/CatColab/pull/1372/changes#diff-197d2d2d6e8b325763aa0dd5f0f2a69374573abe32ee58d2304dfacf337bb6d2R40-R118

tslil-topos:
kaspar has already done this!

PR #1403, opened 2026-08-18
OBEE-38: Shape for Petri nets

tim-at-topos:


packages/logics/src/petri-net.ts
```diff
@@ -0,0 +1,22 @@
+import { defineMorphism, defineObject, defineShape, RichText } from "catcolab-documents";
+
+export const Place = defineObject({ tag: "Basic", content: "Object" });
+
+export const Transition = defineMorphism(
+    { tag: "Hom", content: Place.obType },
+    {
+        domain: { apply: { tag: "Basic", content: "tensor" }, modality: "SymmetricList" },
+        codomain: { apply: { tag: "Basic", content: "tensor" }, modality: "SymmetricList" },
+    },
```
kasbah, line 10: Since we confirmed that a "Hom"  is an endomorphism we could refine `defineMorphism` to only require a single argument `object:` when the tag is "Hom". Not required for this PR, just making a note.

kasbah: https://github.com/ToposInstitute/CatColab/issues/1427

kasbah (approved):

PR #1404, opened 2026-08-18
Consistent heading levels across menu pages (and some CSS modules)

tim-at-topos:


epatters (changes requested):
Thanks for paying attention to consistency!

As it stands, these header styles are *global* across the application and they conflict with the existing header styles in `packages/ui-components/src/global.ss`. I bet this is why you were having those issues with the fonts changing that you showed me earlier today.

tim-at-topos:
ok, I've split out a separate css for menu pages (i.e. pages accessed directly through the hamburger menu), but it seems like we should one day sit down and go through and package everything up into more specific css modules, like you were saying the other day

packages/frontend/src/user/menu_pages.css
```diff
@@ -0,0 +1,18 @@
...
+        font-size: var(--h3-font-size);
+        margin: 1.25em 0 0.5em 0;
+        border-bottom: 1px solid black;
+    }
+
+    h3 {
+        font-size: var(--h4-font-size);
+        margin: 1em 0 0.5em 0;
+    }
+}
```
kasbah: Can this be a menu_pages.module.css?

tim-at-topos: I just mentioned this in a comment above. Happy to do this if ya like?

kasbah: If there are no blockers then yes please.

epatters:
> it seems like we should one day sit down and go through and package everything up into more specific css modules, like you were saying the other day

We should perhaps do this systematically one day, but, consistent with Kaspar's comment, I also think it's fine to start module-izing bits of the CSS piecemeal, whenever it's convenient and especially if it help prevent actual bugs.

packages/frontend/src/user/profile.tsx
```diff
@@ -10,22 +10,22 @@ import { LoginGate } from "./login";
...
+
 /** Page to configure user settings. */
 export default function UserProfilePage() {
     const appTitle = import.meta.env.VITE_APP_TITLE;
 
     return (
         <>
             <Title>User Settings - {appTitle}</Title>
-            <div class="growable-container">
+            <div class={`user-settings-page ${styles.menuPage}`}>
```
kasbah: What happened to the `growable-container` styles, do we still need them? Where are the `user-settings-page` styles?

kasbah (approved):

epatters (approved):
Thanks, looks great.

PR #1408, opened 2026-08-20
Show unnamed as "Unnamed" more consistently

kasbah:
In completions and morphism and object inputs. Previously there was an almost imperceptible space you could select in completions.

image


epatters (approved):
Thanks as always for your attention to these details!

PR #1411, opened 2026-08-20
Modal editing for identity morphisms in path equations

epatters:
Followup to #1247 changing how identity morphisms work. Rather than wrapping and parsing object labels from `id(...)`, identity morphisms are entered by giving the object name directly. When input field is no longer active, the identity is displayed as `id_{...}` with a real subscript.

### DIFF
--- a/packages/ui-components/src/inline_list_editor.tsx
+++ b/packages/ui-components/src/inline_list_editor.tsx
@@ -237,12 +237,20 @@ export function InlineListEditor<T>(originalProps: InlineListEditorProps<T>) {
                 }
             }}
             onFocusOut={(evt) => {
-                // Lose focus only when it moves outside the list entirely.
+                // Lose focus only when it moves outside the list entirely. The
+                // check is deferred because the browser resets the focus to the
+                // document body when a focused input is unmounted, as happens
+                // when an item is deleted. The focus is then restored within the
+                // list, synchronously, before the deferred check runs.
                 const next = evt.relatedTarget as Element | null;
                 if (next && listRef.contains(next)) {
                     return;
                 }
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/inline_list_editor.tsx
```diff
@@ -237,12 +237,20 @@ export function InlineListEditor<T>(originalProps: InlineListEditorProps<T>) {
                 }
             }}
             onFocusOut={(evt) => {
-                // Lose focus only when it moves outside the list entirely.
+                // Lose focus only when it moves outside the list entirely. The
+                // check is deferred because the browser resets the focus to the
+                // document body when a focused input is unmounted, as happens
+                // when an item is deleted. The focus is then restored within the
+                // list, synchronously, before the deferred check runs.
```
epatters: Warning: This was found by Claude and seemed to fix a focus issue I was having, though even with the explanation I don't really understand it. This kind of delicate focus management makes me nervous, and now it's even more delicate.

kasbah:
I pushed a commit to reduce the font size of the composition symbol.

image

packages/ui-components/src/inline_list_editor.tsx
kasbah: Can you explain how to replicate the focus issue please?

epatters: And... of course now I can't replicate it. Perhaps this "fix" was working around some other problem with an earlier version of this code that has since been fixed. I've reverted this change.

epatters:
Thanks for the CSS improvement. Any other comments on this one, @kasbah?

kasbah (approved):
LGTM

PR #1412, opened 2026-08-21
OBEE-36: add instance types and validation

tslil-topos:
*tests will be in the last pr on this stack*

Define instance, table, and error types as well as add runtime support for instance-capable shapes. Add temporary implementations of atomic column typing and instance validation until we decide how we want this to work mathematically. We do not add capability to actually generate anything of these types in this change.

Adjust the pnpm config to build generated document types before checking (we now depend on catlog-wasm) and increase the validation test timeout for loading catlog-wasm.

This change includes the revised interface for instances and usage pattern that Kaspar and i had discussed. See `src/instance/instance.ts` for details.

In order to obtain `tables` for example, the expected pattern is
```typescript
// subscribe to validation
const unsubscribe = instance.onValidate(async (validation) => {
  // ignore invalid schemas or instances
  if (validation.tag !== "Ok") return;
  // both were valid at the callback time, try to read the table content
  const result = await instance.tables();
  // something invalidated either the schema or instance
  if (result.tag !== "Ok") return;
  // use tables...
  const tables = result.content;
});

// finally
unsubscribe();
```

### DIFF
--- a/packages/documents/src/instance/instance.ts
+++ b/packages/documents/src/instance/instance.ts
@@ -32,29 +32,29 @@ export interface Instance<H, S extends Shape> {
     updateRow(
         row: TableRow,
         values: Record<string, LiteralValue | TableRow>,
-    ): Promise<Result<undefined, ReadonlyArray<Issue | TableFieldIssue>>>;
+    ): Promise<Result<void, ReadonlyArray<Issue | TableFieldIssue>>>;
     updateRows(
         updates: ReadonlyArray<{
             row: TableRow;
             values: ReadonlyArray<Record<string, LiteralValue | TableRow>>;
         }>,
-    ): Promise<Result<undefined, ReadonlyArray<Issue | TableFieldIssue>>>;
+    ): Promise<Result<void, ReadonlyArray<Issue | TableFieldIssue>>>;
     set(
         row: TableRow,
         morphism: { id: string },
         value: LiteralValue | TableRow,
-    ): Promise<Result<undefined, ReadonlyArray<Issue | TableFieldIssue>>>;
+    ): Promise<Result<void, ReadonlyArray<Issue | TableFieldIssue>>>;
 
     /** Delete stored rows without requiring a valid schema. */
     deleteRow(tableId: string, rowId: string): void;
     deleteRows(rows: ReadonlyArray<{ tableId: string; rowId: string }>): void;
 
     /* Validate both the schema and then the instance */
-    validate(): Promise<Result<undefined, ReadonlyArray<Issue | TableFieldIssue>>>;
+    validate(): Promise<Result<void, ReadonlyArray<Issue | TableFieldIssue>>>;
     /** Subscribe to changes to either the instance document or its schema. */
     onChange(callback: () => void): () => void;
     /** Revalidate initially and whenever either the instance or its schema changes. */
     onValidate(
-        callback: (result: Result<undefined, ReadonlyArray<Issue | TableFieldIssue>>) => void,
+        callback: (result: Result<void, ReadonlyArray<Issue | TableFieldIssue>>) => void,
     ): () => void;
 }
--- a/packages/documents/src/instance/instance.ts
+++ b/packages/documents/src/instance/instance.ts
@@ -1,0 +1,60 @@
+import type { InstanceDocument } from "catcolab-document-methods";
+import type { Issue, Result } from "../result";
+import type { Shape } from "../shape";
+import type { TableFieldIssue } from "./errors";
+import type { FieldValue, InstancePath, InstanceTable, LiteralValue, TableRow } from "./tables";
+
+export type { InstanceDocument } from "catcolab-document-methods";
+
+/** API for an instance document and its schema-derived tables. */
+export interface Instance<H, S extends Shape> {
+    readonly handle: H;
+    readonly shape: S;
+    readonly document: Readonly<InstanceDocument>;
+    readonly title: string;
+
+    tables(): Promise<Result<ReadonlyArray<InstanceTable>>>;
+    get(path: InstancePath): Promise<Result<InstanceTable | TableRow | FieldValue>>;
+
+    update(patch: Partial<{ title: string }>): void;
+    dump(): InstanceDocument;
+
+    addRow(
+        table: InstanceTable,
+        values?: Record<string, LiteralValue | TableRow>,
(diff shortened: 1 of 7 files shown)


packages/documents/src/instance/instance.ts
```diff
@@ -0,0 +1,66 @@
+import type { InstanceDocument } from "catcolab-document-methods";
+import type { Issue, Result } from "../result";
+import type { Shape } from "../shape";
+import type { TableFieldIssue } from "./errors";
+import type { FieldValue, InstancePath, InstanceTable, LiteralValue, TableRow } from "./tables";
+
+export type { InstanceDocument } from "catcolab-document-methods";
+
+/** Store-backed API for an instance document and its schema-derived tables. */
```
kasbah: "Store-backed" as opposed to... ?

packages/documents/src/instance/instance.ts
```diff
@@ -0,0 +1,66 @@
...
+    get(path: InstancePath): Promise<Result<InstanceTable | TableRow | FieldValue>>;
+
+    update(patch: Partial<{ title: string }>): void;
+    dump(): InstanceDocument;
+
+    addRow(
+        table: InstanceTable,
+        args?: Record<string, LiteralValue | TableRow>,
+    ): Promise<Result<TableRow>>;
+    addRowMany(
```
kasbah: I would prefer just adding `s` for the bulk methods: `addRows`

packages/documents/src/instance/instance.ts
```diff
@@ -0,0 +1,66 @@
...
+        morphism: { id: string },
+        value: LiteralValue | TableRow,
+    ): Promise<Result<undefined, ReadonlyArray<Issue | TableFieldIssue>>>;
+    setMany(
+        updates: ReadonlyArray<{
+            row: TableRow;
+            values: ReadonlyArray<[{ id: string }, LiteralValue | TableRow]>;
+        }>,
+    ): Promise<Result<undefined, ReadonlyArray<Issue | TableFieldIssue>>>;
+
```
kasbah: Are you sure this one is useful? Setting multiple different values across different tables and rows in bulk?

kasbah (changes requested):
Am I reading this right that there is no implementation here yet?

packages/documents/src/instance/instance.ts
tslil-topos: happy to remove.

packages/documents/src/instance/instance.ts
tslil-topos: `set`?

packages/documents/src/instance/instance.ts
tslil-topos: we're sure of nothing, we can later subtract but it's basically a simple loop over the primitive.

packages/documents/src/instance/instance.ts
tslil-topos: `delete`?

kasbah: `deleteRows` I believe and I don't think we should have `setMany` at all.

packages/documents/src/instance/instance.ts
kasbah: My worry with `setMany` is that LLMs will use it and it will result in less readable code vs multiple `set` calls.

tslil-topos: I don't think i see the difference between five consecutive set calls and a single set call containing the array of arguments between those five?

kasbah: I find this much more readable:

```ts
await instance.set(alice, name, "Alice");
await instance.set(alice, age, 32);
await instance.set(alice, manager, bob);
await instance.set(bob, name, "Bob");
```

vs

```ts
await instance.setMany([
    {
        row: alice,
        values: [
            [name, "Alice"],
            [age, 32],
            [manager, bob],
        ],
    },
    {
        row: bob,
        values: [[name, "Bob"]],
    },
]);
```

Given the functions are async I can see why `setMany` was added, however even then the bulk operation is more readable with `updateRow` or `updateRows`, plus this becomes less of an issue when we add transactions?

```ts
await instance.updateRow(alice, {
    name: "Alice",
    age: 32,
    manager: bob,
});

await instance.updateRow(bob, {
    name: "Bob",
    age: 41,
});
```

```ts
await instance.updateRows([
    {
        row: alice,
        args: {
            name: "Alice",
            age: 32,
            manager: bob,
        },
    },
    {
        row: bob,
        args: {
            name: "Bob",
            age: 41,
        },
    },
]);
```


`set` was meant as an escape hatch when we can't use the `label` as the key in `args` which is why I don't think we need the bulk version.

> it's basically a simple loop over the primitive 

But it's an added method on our API surface. In API design less can be more and I think we should cut `setMany` now.

packages/documents/src/instance/instance.ts
```diff
@@ -0,0 +1,66 @@
...
+        }>,
+    ): Promise<Result<ReadonlyArray<TableRow>>>;
+    updateRow(
+        row: TableRow,
+        args: Record<string, LiteralValue | TableRow>,
+    ): Promise<Result<undefined, ReadonlyArray<Issue | TableFieldIssue>>>;
+    updateRowMany(
+        updates: ReadonlyArray<{
+            row: TableRow;
+            args: ReadonlyArray<Record<string, LiteralValue | TableRow>>;
```
kasbah: I think `args` should be called `values`.

tslil-topos: is a tableRow a value?

packages/documents/src/instance/instance.ts
tslil-topos: Fine with me

packages/documents/src/instance/instance.ts
kasbah: These are the values of the table row. "args" of the table row makes less sense to me i.e. this reads better:

```ts
await instance.updateRows([
    {
        row: alice,
        values: {
            name: "Alice",
            age: 32,
            manager: bob,
        },
    },
    {
        row: bob,
        values: {
            name: "Bob",
            age: 41,
        },
    },
]);
```

kasbah:
One minor suggestion for you: https://github.com/ToposInstitute/CatColab/pull/1417/

kasbah (approved):

PR #1413, opened 2026-08-21
OBEE-39: implement table runtime and backing for createInstance factory

tslil-topos:
*tests will be included in the last change stacked on this*

This is an absolute grind of a change. There are simply twelve quadrillion things that have to happen to make things like changing rows or adjusting tables work correctly.

The basic pattern is visible in instance-runtime.ts :

```typescript
async function withValidatedSchema<
    Handle,
    S extends Shape,
    T,
    E extends ReadonlyArray<Issue> = ReadonlyArray<Issue>,
>(
    schema: Notebook<S, ModelDocument>,
    operation: (schemaModel: DblModel) => Result<T, E>,
): Promise<Result<T, E>>
```

is used to perform the steps of obtaining a dblmodel (or erroring if the schema doesn't validate), learning about the table structure, and finally running the operation (which might itself fail in a way at least as exciting as ReadonlyArray<Issue>). So this is what is contained in the factory functions that will (in a later change) back the Instance factory: `createTablesMethod`, `createGetMethod`, `createAddRowManyMethod`, etc.

Then there's the elephant: table-methods.ts. The basic premise for all of these is to permissively apply everything that makes sense, and report everything that failed. Thus an `Ok` is only possible if went well. The only kinds of failures that are important here are addressing failures: mentioning a table that doesn't exist, mentioning a row that doesn't exist. Type issues, missing entries, bad links, and so on are all semantic checks that will be caught by validation passes.

### DIFF
--- a/packages/documents/src/instance/table-methods.ts
+++ b/packages/documents/src/instance/table-methods.ts
@@ -1,0 +1,419 @@
+import { v7 as uuid } from "uuid";
+
+import type { InstanceDocument } from "catcolab-document-methods";
+import type * as DocumentTypes from "catcolab-document-types";
+import type { DblModel, ObGenerator } from "catlog-wasm";
+import type { DocumentStore } from "../document-store";
+import { objectTypesEqual } from "../model/equality";
+import type { Issue, Result } from "../result";
+import type { InstanceCapableShape } from "../shape";
+import type { FieldPath } from "./errors";
+import type {
+    FieldValue,
+    InstancePath,
+    InstanceTable,
+    LiteralValue,
+    TableHeader,
+    TableRow,
+} from "./tables";
+import { atomicTypeOfAttributeType } from "./validation";
+
+/** Read one table, row, or field using a validated schema model.
+
+Addressing failures are reported as issues. */
+export function readInstancePathFromStore<Handle>(
+    shape: InstanceCapableShape,
+    store: DocumentStore<Handle>,
+    handle: Handle,
+    schemaModel: DblModel,
+    path: InstancePath,
+): Result<InstanceTable | TableRow | FieldValue> {
+    const schemaTables = instanceTablesFromModel(shape, store, handle, schemaModel);
+    const schemaTableById = new Map(
+        schemaTables.map((schemaTable) => [schemaTable.id, schemaTable]),
+    );
+    const document = store.getDocumentView(handle) as Readonly<InstanceDocument>;
+    const [tableId, rowsSegment, rowId, fieldsSegment, fieldId, ...rest] = path;
+    const schemaTable = schemaTableById.get(tableId);
+    if (schemaTable === undefined) {
+        return pathError(`Table \`${tableId}\` does not exist`);
+    }
+    if (path.length === 1) {
+        return { tag: "Ok", content: schemaTable };
+    }
+    if (rowsSegment !== "rows" || rowId === undefined) {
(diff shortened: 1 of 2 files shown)


packages/documents/src/instance/table-methods.ts
```diff
@@ -0,0 +1,419 @@
...
+            ordered.push(id);
+        }
+    }
+    for (const id of Object.keys(table.rows)) {
+        if (!seen.has(id)) {
+            ordered.push(id);
+        }
+    }
+    return ordered;
+}
```
kasbah, line 415: This is being defensive in the wrong place and also papers over any structural issues if they are there. It speaks to the overall flaw I see in this implementation: we should parse our doc into a structure where we know:

- `table.rowOrder` is not undefined
- `table.rowOrder` has all the keys of `table.rows` and no more
- any other invariants we want to hold for the JSON structure

We should then be able to use `table.rowOrder` directly instead of this function. In general it should allow us to split up the process into something like "parse-doc.ts" and then "table-methods.ts" that operate on the parsed doc.

kasbah (changes requested):

packages/documents/src/instance/table-methods.ts
tslil-topos: I think flaw is a strong word. The comments and PR text explain the approach taken: all actions are maximally permissive. If the instance is in a bad state we should still be able to take all actions that we are interested in. It's up to users of this api to ensure that valid passes, and all of these functions share the guarantee: if the schema and instance are valid then all operations are faithful and exhibit all data.

kasbah: I don't know what word to use but I disagree with the approach exhibited by this defensive error swallowing code and with things like:

> Every update that fails to address is skipped and the rest are still applied.

Where "skipped" means "silently skipped"

kasbah: Maybe you could weigh in on this one @epatters since tslil and I seem to have very different views on this.

tslil-topos: > I don't know what word to use but I disagree with the approach exhibited by this defensive error swallowing code and with things like:
> 
> > Every update that fails to address is skipped and the rest are still applied.
> 
> Where "skipped" means "silently skipped"

it does not, did you see the preceding sentence or the code paths `Only addressing failures are reported as issues.`?

tslil-topos:
@kasbah please approve this change so that we can proceed to the difficult and time-consuming integration work that lives on-top of this stack. If we want to revisit this approach, assuming that this doesn't break fundamental UI operations (would be happy to hear a case that the split of error reporting between addressing issues on operations, all issues on validate is difficult for consumers), then we can do so after the deadline and we could open an issue now to remind us to do so once there is more time.

packages/documents/src/instance/table-methods.ts
kasbah: Ok, looks good on re-reading. You pushed some updates to comments?

I still think that the `function orderedRowIds` is not the right approach. I'll open a PR for how I think it can be improved.

kasbah: https://github.com/ToposInstitute/CatColab/issues/1426

kasbah (approved):

packages/documents/src/instance/table-methods.ts
tslil-topos: thank you!

PR #1415, opened 2026-08-21
OBEE-32: integrate instances

tslil-topos:
*tests will come at the end of this commit stack*

This change exposes store handles on notebooks so that binders can validate schemas and form instance-of links before creating instances. This allows us to make creating instances actually possible for the first time in this stack.

Propagate the notebook handle type through existing consumers.

### DIFF
--- a/packages/documents/src/binder.ts
+++ b/packages/documents/src/binder.ts
@@ -1,9 +1,11 @@
-import { Model } from "catcolab-document-methods";
+import { Instance as InstanceMethods, Model } from "catcolab-document-methods";
 import type { Document } from "catcolab-document-types";
 import type { DocumentStore } from "./document-store";
-import { createInMemoryStore } from "./document-store/in-memory";
+import { createInMemoryStore } from "./document-store";
+import { instanceFromStore, type Instance } from "./instance/instance";
 import type { ModelDocument } from "./model/document";
 import { modelNotebookFromStore, type Notebook } from "./model/notebook";
+import type { Result } from "./result";
 import type { Shape } from "./shape";
 
 export interface Binder<Handle> {
@@ -13,7 +15,12 @@ export interface Binder<Handle> {
     createNotebook<S extends Shape & { readonly theory: string }>(
         shape: S,
         options: { title: string },
-    ): Promise<Notebook<S, ModelDocument>>;
+    ): Promise<Notebook<S, ModelDocument, Handle>>;
+
+    createInstance<S extends Shape>(
+        schema: Notebook<S, ModelDocument, Handle>,
+        options: { title: string },
+    ): Promise<Result<Instance<Handle, S>>>;
 }
 
 /* Overloads rather than a single signature `createBinder<Handle>(store?:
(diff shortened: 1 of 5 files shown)


packages/documents/src/binder.ts
```diff
@@ -48,5 +55,45 @@ function binderFromStore<Handle>(store: DocumentStore<Handle>): Binder<Handle> {
...
+                return schemaResult;
+            }
+
+            const schemaModel = schemaResult.content;
+            try {
+                const schemaRef = store.getDocumentRef(schema.handle);
+                const document = InstanceMethods.newInstanceDocument({
+                    _id: schemaRef.id,
+                    _version: schemaRef.version,
+                    _server: schemaRef.server ?? "",
```
tslil-topos, line 85: not sure what to do here

packages/documents/src/model/notebook.ts
```diff
@@ -162,10 +162,11 @@ function cellMatchesFilter<S extends Shape>(cell: CellOf<S>, filter: AnyCellType
     }
 }
 
-export interface Notebook<S extends Shape, D extends NotebookDocument = NotebookDocument> {
+export interface Notebook<H, S extends Shape, D extends NotebookDocument = NotebookDocument> {
```
kasbah: Offline I suggested this:

```suggestion
export interface Notebook<S extends Shape, D extends NotebookDocument = NotebookDocument, H = unknown> {
```

packages/documents/src/binder.ts
kasbah: what's the problem?

tslil-topos: `schemaRef.server ?? ""`, it's unclear to me what the default should be

packages/documents/src/model/notebook.ts
tslil-topos: this is a minor aesthetic change, we can discuss its merits separately to this stacked pr, open an issue, or otherwise revisit it when there are less pressing things to attend to. If there are no other concerns please approve this change.

kasbah: This will be annoying to refactor in the future so we should make the change now. Without this the type is un-ergonomic for notebooks of generic shapes.

packages/documents/src/binder.ts
kasbah: I think we can leave it as is for now and make an issue to improve it. It's non-optional in the document-types but optional in the TS to make the example of loading a notebook from a ref less confusing: the server field would be required but currently can't actually be anything other than the local "server". 

```ts
const loaded = await binder.loadNotebookFromRef(PetriNet, {
    id: "some-document-id",
    version: null,
    server: "why-do-i-have-to-add-this.com"
});
```

We might add a `store.server` field, and mint a UUID one at construction if empty, to solve this properly in the future.

packages/documents/src/model/notebook.ts
tslil-topos: alright, i will make this change now.

kasbah (approved):

PR #1416, opened 2026-08-21
OBEE-40: test instances and tables

tslil-topos:
Cover schema-validated creation and mutation, async table derivation, cross-table bulk operations, schema-independent deletion, path behavior, combined subscriptions, atomic typing, and semantic field errors.

### DIFF
--- a/packages/documents/test/instances.test.ts
+++ b/packages/documents/test/instances.test.ts
@@ -1,69 +1,491 @@
 import { Attr, AttrType, Entity, Mapping, SimpleSchema } from "catcolab-logics/simple-schema";
 import { describe, expect, test } from "vitest";
 
-// RFC-0006 "Tabular instances": an instance API that mirrors the notebook API,
-// with `add` keyed by schema entities and a `rowsOf` that mirrors `cellsOf`.
-// Unlike cells, rows keep a separate `values` field to keep them in their own
-// namespace.
-import { createBinder } from "catcolab-documents";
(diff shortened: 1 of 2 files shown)


packages/documents/test/instances.test.ts
```diff
@@ -1,69 +1,491 @@
...
+            );
+
+            expect(predatorPrey.title).toBe("Predator-prey");
+            expect(predatorPrey.shape).toBe(SimpleSchema);
+            expect(predatorPrey.document.tables).toEqual({});
+            expect(predatorPrey.document.instanceOf._id).toBe(
+                refOf(binder.store, causalLoop.handle).id,
+            );
+        },
+    );
```
kasbah, line 124: I understand that this was likely in the tests because of the example I put into the RFC but we should change the example to be a less confusing schema.

tslil-topos: The agreed-upon model in https://github.com/ToposInstitute/CatColab/pull/1346#pullrequestreview-4905688030 allowed for machine generated tests that nobody has inspected carefully. I don't see why we should start now with this change.

If you'd like to change this please open an issue and when we have less pressing concerns we can revisit this and other tests for the api.

kasbah: Yes yes, my intention wasn't to block the PR on this, just making a note to improve this.

kasbah (approved):

PR #1418, opened 2026-08-21
TableEditor component

kasbah:
<a href="https://branch-kb-table-component--catcolab.netlify.app/dev/ui-components/?path=/docs/forms-inputs-tableeditor--docs"> 

image

 </a>

Includes some hacks to get the 8 year old MIT licensed Hansontable to work like we want.

[try it out](https://branch-kb-table-component--catcolab.netlify.app/dev/ui-components/?path=/docs/forms-inputs-tableeditor--docs)

### DIFF
--- a/packages/ui-components/src/table_editor/table_editor.tsx
+++ b/packages/ui-components/src/table_editor/table_editor.tsx
@@ -5,7 +5,7 @@ import {
     type HandsontableInstance,
     type HandsontableSettings,
 } from "./handsontable";
-import Handsontable from "./handsontable_core.js";
+import type Handsontable from "./handsontable_core.js";
 
 import styles from "./table_editor.module.css";
 
--- a/packages/ui-components/src/table_editor/table_editor.tsx
+++ b/packages/ui-components/src/table_editor/table_editor.tsx
@@ -53,7 +53,7 @@ export function TableEditor(props: TableEditorProps) {
     const deleteRowRenderer = (instance: Handsontable, td: HTMLElement, row: number) => {
         const button = document.createElement("button");
         button.type = "button";
-        button.className = styles.deleteRow!;
+        button.className = styles.deleteRow;
         button.title = "Delete row";
         button.setAttribute("aria-label", "Delete row");
         button.textContent = "×";
@@ -63,7 +63,7 @@ export function TableEditor(props: TableEditorProps) {
             event.stopPropagation();
             instance.alter("remove_row", row);
         });
-        td.classList.add(styles.deleteCell!);
+        td.classList.add(styles.deleteCell);
         td.replaceChildren(button);
         return td;
     };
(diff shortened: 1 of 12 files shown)


packages/ui-components/src/table_editor/table_editor.tsx
```diff
@@ -0,0 +1,137 @@
+import { type JSX, Show } from "solid-js";
+
+import {
+    Handsontable as HandsontableGrid,
+    type HandsontableInstance,
+    type HandsontableSettings,
+} from "./handsontable";
+import Handsontable from "./handsontable_core.js";
```
epatters: I'm guessing that the JS blob is pretty big. If so, we should lazily import Handsontable to enable code splitting in the bundle, as we do for other big dependencies like Graphviz and Elk. The `TableEditor` could be a natural place to put this logic, but we could also have this import be eager and then lazy load the table editor itself downstream. What do you think?

packages/ui-components/src/table_editor/table_editor.tsx
```diff
@@ -0,0 +1,137 @@
...
+    /** Called after the Handsontable instance is created. */
+    onReady?: (instance: TableEditorInstance) => void;
+};
+
+/** A spreadsheet-esque table editor with controls for adding and deleting rows. */
+export function TableEditor(props: TableEditorProps) {
+    let table: TableEditorInstance | undefined;
+
+    const deleteRowRenderer = (instance: Handsontable, td: HTMLElement, row: number) => {
+        const button = document.createElement("button");
```
epatters, line 61: Is it not workable to have this renderer be a Solid component?

packages/ui-components/src/table_editor/table_editor.tsx
```diff
@@ -0,0 +1,137 @@
...
+                      : settings.colWidths;
+                  const columnWidth =
+                      typeof column.width === "number"
+                          ? column.width
+                          : typeof configuredWidth === "number"
+                            ? configuredWidth
+                            : DEFAULT_COLUMN_WIDTH;
+                  return total + columnWidth;
+              }, 0)
+            : undefined;
```
epatters: The body of this function is pretty hard to read. Is there a nicer way to write this?

epatters (commented):
Thanks, this is cool! I'm pleasantly surprised at how little monkey-patching is required given that this version of the dependency has been unmaintained for 8 years. I agree that with this level of patching it's cleaner to avoid vendoring the code.

A few comments below.

packages/ui-components/src/table_editor/table_editor.tsx
kasbah: Part of this was that Handsontable has a lot of things it accepts in settings. I tightened our type to restrict what we accept and simplified this logic.

packages/ui-components/src/table_editor/table_editor.tsx
kasbah: I made the component lazy load it.

packages/ui-components/src/table_editor/table_editor.tsx
kasbah: That seems complicated and while possible, not worth it. Handsontable owns the life-cycle of the cell content and the expected interface is a `HTMLElement`. We would have to re-create SolidJS reactive root on-demand.

epatters (approved):

PR #1419, opened 2026-08-24
Make frontend test users unique

kasbah:
This solves some intermittent failures with frontend tests like [this run](https://github.com/ToposInstitute/CatColab/actions/runs/32509855266/job/96858171510?pr=1411). The LLM analysis of the situation:

> - This job started Vitest at 13:50:23.
> - Another CI run, 32509819873 (https://github.com/ToposInstitute/CatColab/actions/runs/32509819873), started the same integration tests at 13:50:10.
> - Both runs used hard-coded accounts such as test-doc-editing@catcolab.org and deleted them during afterAll.
> - The earlier run deleted accounts while this run was still using them, invalidating its sessions.
>
> That produced the characteristic cascade:
> - auth/user-token-expired
> - auth/invalid-credential
> - auth/too-many-requests after repeated login attempts
> - Not authorized to access ref


epatters (approved):
Thanks! Seems like the right approach in general.

PR #1422, opened 2026-08-24
OBEE-43 Add LLMConversation document to API

tslil-topos:
This small change adds LLMConversation documents and some tests to the documents API surface. This is a step towards integrating the API with LLM inference in that the conversation scope in the front-end will be changed to deal strictly in binders/stores and it is therefore necessary to port from using document-methods directly in the front-end to the documents API surface.

### DIFF
--- a/packages/documents/package.json
+++ b/packages/documents/package.json
@@ -44,6 +44,6 @@
         "vite": "^7.2.2",
         "vite-plugin-solid": "^2.11.10",
         "vite-plugin-wasm": "^3.5.0",
-        "vitest": "4.1.10"
+        "vitest": "^4.0.8"
     }
 }
--- a/packages/documents/package.json
+++ b/packages/documents/package.json
@@ -44,6 +44,6 @@
         "vite": "^7.2.2",
         "vite-plugin-solid": "^2.11.10",
         "vite-plugin-wasm": "^3.5.0",
-        "vitest": "^4.0.8"
+        "vitest": "4.1.10"
     }
 }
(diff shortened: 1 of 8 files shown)


packages/documents/package.json
```diff
@@ -44,6 +44,6 @@
         "vite": "^7.2.2",
         "vite-plugin-solid": "^2.11.10",
         "vite-plugin-wasm": "^3.5.0",
-        "vitest": "4.1.10"
+        "vitest": "^4.0.8"
```
kasbah:
```suggestion
        "vitest": "4.1.10"
```

tslil-topos: yeah unfortunately the lockfile is also wrong, i will regenerate with this change

packages/documents/test/llm-conversation.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+import { LLMConversation as LLMConversationMethods } from "catcolab-document-methods";
+import { createBinder, llmConversationFromStore } from "catcolab-documents";
+
+async function createConversation(binder: ReturnType<typeof createBinder>, title: string) {
+    const model = await binder.createNotebook(SimpleOlog, { title: "Attached model" });
+    return binder.createLLMConversation(model, "test-model", { title });
+}
+
+describe("LLM conversation documents", () => {
```
kasbah, line 12: These should be enabled in CI 

https://github.com/ToposInstitute/CatColab/blob/b17bab150c1104468108c8307a1edebb14f0609f/.github/workflows/ci.yml#L127

kasbah (approved):

PR #1424, opened 2026-08-25
OBEE-42 integrate catcolab-documents with LLM, add validation enforcement, enhance retry logic

tslil-topos:
(this change has a large +/- because it's about porting to a new interface which did not exist when the FE LLM pathways were originally written. But hey, +752-380 is not a large number overall :) )

This change moves LLM turn execution onto the new catcolab-documents APIs. The new `runLLMConversationTurn` and `retryLastLLMConversationResponse` handle all scoping, prompting, and validation concerns so that callers can be agnostic with respect to the underlying details.

```typescript
   const result = await runLLMConversationTurn(
       conversation,
       store,
       inferenceKey,
       {
           content: userMessage,
           files,
       },
       streamingCallback,
   );
```

The result tells the caller how to proceed:

```typescript
   type LLMConversationTurnResult =
       | { tag: "Completed"; content: string }
       | { tag: "Incomplete"; reason: string }
       | { tag: "Failed"; error: string }
       | { tag: "Retryable"; error: string; attempts: readonly LLMInteraction[]; };
```

and the streaming operates on

```typescript
export type ChatTurnEvent =
    | { tag: "Streaming"; delta: string; snapshot: string }
    | { tag: "RunTool"; code: string }
    | { tag: "ToolResult"; code: string; result: EvalResult };
```

Under the hood, document edits are performed on isolated in-memory copies with pseudo-transaction semantics. If all documents in the scope pass validity and the LLM is done, then we update the contents of the real documents that were passed in to reflect the changes.

The current logic is roughly as follows:
* runLLMConversationTurn: persists one user message and invokes runChatTurn exactly once.
* runChatTurn: invokes the inference provider via the SDK's runTools, limited MAX_PROVIDER_REQUESTS_PER_ATTEMPT
* `contextExec` runs JS locally against in-memory copies of the relevant documents, if execution succeeds a validation hook runs and reports any errors back for the llm to fix.
* when runChatTurn finishes (because the LLM is done or because it exhausted its attempts) we inspect the situation by running validate and looking at the reason that runChatTurn finished
    
Persisting the LLM responses to the conversation and committing the changes is governed by the following policy

 | Outcome                                      | Persist model/tool messages | Commit document edits | Result                              |
|----------------------------------------------|-----------------------------|-----------------------|-------------------------------------|
| Final response, valid scope                  | Yes                         | Yes                   | `Completed`                         |
| Provider limit/incomplete response, valid scope | Yes                      | Yes                   | `Incomplete`                         |
| Validation failure                           | No                          | No                    | `Retryable`  |
| Network/protocol failure                     | No                          | No                    | `Retryable`                         |

Because we're reporting errors via a hook in the tool call, these will be persisted in the actual LLMConversation document for later inspection whenever the interactions are persisted.

We are making the conscious trade-off for simplicity here: if the documents are in an invalid state and the LLM does something unrelated it will be told again that they are in an invalid state. Future changes might enable damage tracking or other sophisticated mechanisms which would allow us to turn down the "noise" on this process.

### DIFF
--- a/packages/documents/test/instances.test.ts
+++ b/packages/documents/test/instances.test.ts
@@ -202,7 +202,7 @@ describe("instance schema validation", () => {
 });
 
 describe("tabular instances", () => {
-    test("tables and headers are derived and rows can be edited", { timeout: 10_000 }, async () => {
+    test("tables and headers are derived and rows can be edited", { timeout: 20_000 }, async () => {
         const binder = createBinder();
         const schema = await binder.createNotebook(SimpleSchema, {
             title: "Company schema",
@@ -333,7 +333,7 @@ describe("tabular instances", () => {
 
     test(
         "mistyped row references and literals report their field paths",
-        { timeout: 10_000 },
+        { timeout: 20_000 },
         async () => {
             const binder = createBinder();
             const schema = await binder.createNotebook(SimpleSchema, {
@@ -430,7 +430,7 @@ describe("tabular instances", () => {
 
     test(
         "deleting a referenced row reports a dangling field path",
-        { timeout: 10_000 },
+        { timeout: 20_000 },
         async () => {
             const binder = createBinder();
             const schema = await binder.createNotebook(SimpleSchema, {
--- a/packages/frontend/src/inference/chat.test.ts
+++ b/packages/frontend/src/inference/chat.test.ts
@@ -5,9 +5,9 @@ import { afterAll, assert, beforeAll, describe, test } from "vitest";
 
 import { createFetchWithAuth, createRpcClient } from "../api/rpc.ts";
 import { createTestUserAuth } from "../util/test_util.ts";
-import { createInferenceClient, runOpenAIChatTurn, type OpenAITranscript } from "./chat.ts";
+import { createInferenceClient, runChatTurn, type ChatTranscript } from "./chat.ts";
 
-// When inference is configured, these tests exercise a live `runOpenAIChatTurn`
+// When inference is configured, these tests exercise a live `runChatTurn`
 // against OpenRouter, using a free model. A backend without
 // `OPENROUTER_PROVISIONING_KEY` is expected to report inference as unavailable,
 // in which case the OpenRouter assertions have nothing to exercise.
@@ -57,10 +57,10 @@ describe("chat turn over OpenRouter", () => {
                 return;
             }
 
-            const transcript: OpenAITranscript = [
+            const transcript: ChatTranscript = [
                 { role: "user", content: "Reply with exactly the word: pong" },
             ];
-            const result = await runOpenAIChatTurn(client, transcript, {}, { model: testModel });
+            const result = await runChatTurn(client, transcript, {}, { model: testModel });
 
             assert.ok(result.content.length > 0, "final content should be a non-empty string");
             assert.strictEqual(transcript[0]?.role, "user");
@@ -69,7 +69,7 @@ describe("chat turn over OpenRouter", () => {
     );
 
     test(
-        "records the contextExec call and result as OpenAI messages",
+        "records the contextExec call and result as chat messages",
         { timeout: 120000 },
         async ({ skip }) => {
             if (!client) {
@@ -77,13 +77,13 @@ describe("chat turn over OpenRouter", () => {
                 return;
             }
 
-            const transcript: OpenAITranscript = [
+            const transcript: ChatTranscript = [
                 {
                     role: "user",
                     content: "Use contextExec to evaluate 6 * 7, then tell me the result.",
                 },
             ];
-            const result = await runOpenAIChatTurn(client, transcript, {}, { model: testModel });
+            const result = await runChatTurn(client, transcript, {}, { model: testModel });
 
             const assistant = result.generatedMessageDelta.find(
                 (message) =>
--- a/packages/frontend/src/inference/chat.test.ts
+++ b/packages/frontend/src/inference/chat.test.ts
@@ -60,7 +60,7 @@ describe("chat turn over OpenRouter", () => {
             const transcript: OpenAITranscript = [
                 { role: "user", content: "Reply with exactly the word: pong" },
             ];
-            const result = await runOpenAIChatTurn(client, transcript, {}, undefined, testModel);
+            const result = await runOpenAIChatTurn(client, transcript, {}, { model: testModel });
 
             assert.ok(result.content.length > 0, "final content should be a non-empty string");
             assert.strictEqual(transcript[0]?.role, "user");
@@ -83,7 +83,7 @@ describe("chat turn over OpenRouter", () => {
                     content: "Use contextExec to evaluate 6 * 7, then tell me the result.",
                 },
             ];
-            const result = await runOpenAIChatTurn(client, transcript, {}, undefined, testModel);
+            const result = await runOpenAIChatTurn(client, transcript, {}, { model: testModel });
 
             const assistant = result.generatedMessageDelta.find(
                 (message) =>
--- a/packages/frontend/src/inference/chat.test.ts
+++ b/packages/frontend/src/inference/chat.test.ts
@@ -60,7 +60,7 @@ describe("chat turn over OpenRouter", () => {
             const transcript: ChatTranscript = [
                 { role: "user", content: "Reply with exactly the word: pong" },
             ];
-            const result = await runChatTurn(client, transcript, {}, { model: testModel });
+            const result = await runChatTurn(client, transcript, {}, undefined, testModel);
 
             assert.ok(result.content.length > 0, "final content should be a non-empty string");
             assert.strictEqual(transcript[0]?.role, "user");
@@ -83,7 +83,7 @@ describe("chat turn over OpenRouter", () => {
                     content: "Use contextExec to evaluate 6 * 7, then tell me the result.",
                 },
             ];
-            const result = await runChatTurn(client, transcript, {}, { model: testModel });
+            const result = await runChatTurn(client, transcript, {}, undefined, testModel);
 
             const assistant = result.generatedMessageDelta.find(
                 (message) =>
--- a/packages/frontend/src/inference/chat.test.ts
+++ b/packages/frontend/src/inference/chat.test.ts
@@ -63,6 +63,7 @@ describe("chat turn over OpenRouter", () => {
             const result = await runChatTurn(client, transcript, {}, undefined, testModel);
 
             assert.ok(result.content.length > 0, "final content should be a non-empty string");
+            assert.deepStrictEqual(result.termination, { tag: "FinalResponse" });
             assert.strictEqual(transcript[0]?.role, "user");
             assert.strictEqual(result.generatedMessageDelta.at(-1)?.role, "assistant");
         },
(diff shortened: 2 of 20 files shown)


packages/logics/package.json
```diff
@@ -28,6 +28,6 @@
         "typescript": "^6.0.3",
         "vite": "^7.2.2",
         "vite-plugin-wasm": "^3.5.0",
-        "vitest": "4.1.10"
+        "vitest": "^4.0.8"
```
kasbah:
```suggestion
        "vitest": "4.1.10"
```

tslil-topos: yep, thanks, same story as the ancestor

packages/frontend/src/llm_conversation/execution_scope.ts
```diff
@@ -0,0 +1,207 @@
...
+    type LLMConversationAttachment,
+    modelNotebookFromStore,
+    type ModelDocument,
+    type Notebook,
+    type DocumentStore,
+    type Shape,
+} from "catcolab-documents";
+import type { ContextExecScope } from "../inference/context_exec";
+
+const API_PROMPT = `The document bindings expose the CatColab document API. A schema binding has \`title\`, \`cells()\`, \`cellsOf(type)\`, \`add(type, values)\`, \`update(patch)\`, and \`validate()\`. An instance binding has \`title\`, \`tables()\`, \`get(path)\`, row editing methods, \`update(patch)\`, and \`validate()\`. These APIs mutate in-memory working copies; changes are applied to the user's documents only after every document validates.`;
```
tslil-topos: this is a placeholder, it should grow to some static description of RFC0006 and other tidbits

packages/frontend/src/llm_conversation/execution_scope.ts
```diff
@@ -0,0 +1,207 @@
...
+            throw new Error("The instance schema is not a model document.");
+        }
+        sourceSchema = modelNotebookFromStore(sourceInstance.shape, store, schemaHandle);
+    }
+
+    if (!sourceSchema.shape.theory) {
+        throw new Error("The document shape does not identify its theory.");
+    }
+    const shape = sourceSchema.shape as Shape & { readonly theory: string };
+    const schemaHandle = sourceSchema.handle;
```
tslil-topos: here is where we make the in-memory copy

packages/frontend/src/llm_conversation/execution_scope.ts
```diff
@@ -0,0 +1,207 @@
...
+            if (result.tag === "Err") {
+                return formatIssues(result.content).map((issue) => `${schemaBinding}: ${issue}`);
+            }
+            result.content.free();
+            return [];
+        },
+    ];
+    const commits: Array<() => void> = [
+        () => {
+            replaceDocument(store, schemaHandle, copySchema.dump());
```
tslil-topos: here is where we flush these back to the other store

packages/frontend/src/llm_conversation/execution_scope.ts
kasbah: i don't see where the actual cells are copied

tslil-topos: it's below in `replaceDocument` which is quite gnarly. I'm refactoring this change to make the transaction stuff simpler

packages/documents/test/instances.test.ts
```diff
@@ -202,7 +202,7 @@ describe("instance schema validation", () => {
 });
 
 describe("tabular instances", () => {
-    test("tables and headers are derived and rows can be edited", { timeout: 10_000 }, async () => {
```
tslil-topos, line 205: wasm loading timeouts in CI

packages/frontend/src/inference/chat.test.ts
```diff
@@ -5,9 +5,9 @@ import { afterAll, assert, beforeAll, describe, test } from "vitest";
 
 import { createFetchWithAuth, createRpcClient } from "../api/rpc.ts";
 import { createTestUserAuth } from "../util/test_util.ts";
-import { createInferenceClient, runOpenAIChatTurn, type OpenAITranscript } from "./chat.ts";
+import { createInferenceClient, runChatTurn, type ChatTranscript } from "./chat.ts";
```
tslil-topos, line 8: the fact that we're using the OpenAI api package doesn't mean that we need to name all of our concepts and our types after them, i regret this misstep

packages/frontend/src/llm_conversation/document.test.ts
tslil-topos, line 1: this file was too in-the-weeds for what it wanted to achieve, there are now three integration style tests that cover the external caller surface area

packages/frontend/src/llm_conversation/document.ts
```diff
@@ -104,119 +112,143 @@ export function resolveLLMConversationFeedback(
...
- * Streaming assistant text is reported only through `onContent` and is never persisted.
- */
-export async function runLLMConversationTurn(
-    conversation: LiveLLMConversationDoc,
+/** Persist a user message, run one model turn, and apply its valid document edits. */
+export async function runLLMConversationTurn<
+    Handle,
+    Attachment extends LLMConversationAttachment<Shape, Handle>,
+>(
+    conversation: LLMConversationAPI<Attachment, Handle>,
```
tslil-topos, line 71: mostly the changes here are a port from LiveDoc to the documents API, as well as the introduction of automatic scope and prompt preparation

packages/frontend/src/llm_conversation/execution_scope.ts
```diff
@@ -0,0 +1,158 @@
...
+    type Shape,
+} from "catcolab-documents";
+import type { ContextExecScope } from "../inference/context_exec";
+import {
+    createScopedDocument,
+    type ScopedDocument,
+    type ScopedDocumentRole,
+} from "./scoped_document";
+
+const API_PROMPT = `The document bindings expose the CatColab document API. A notebook binding has \`title\`, \`cells()\`, \`cellsOf(type)\`, \`add(type, values)\`, \`update(patch)\`, and \`validate()\`. A tabular document binding has \`title\`, \`tables()\`, \`get(path)\`, row editing methods, \`update(patch)\`, and \`validate()\`. These APIs mutate in-memory working copies; changes are applied to the user's documents only after every document validates.`;
```
tslil-topos: this is expected to be replaced in the future by some condensed version of RFC0006 and whatever has been done since then

packages/frontend/src/llm_conversation/document_transaction.ts
```diff
@@ -0,0 +1,82 @@
+import type { Document } from "catcolab-document-types";
+import type { DocumentStore } from "catcolab-documents";
+
+/**
+ * Transaction semantics for staging edits to a document behind a working copy,
+ * and later committing that copy back over the original. This file is expected
+ * to be replaced with a more principled implementation that is kinder to
+ * automerge in the future.
+ */
```
kasbah, line 9: I'm still wondering why this is easier than an actual implementation. Did you see the "Possible implementation" details in the RFC? I feel like it offers a simpler to implement solution that is also kinder  to automerge.

> To implement this, we can extend the `DocumentStore` to have a `Version` parameter and some added methods.
> 
> ```ts
> interface Commit<Version> {
>     before: Version;
>     after: Version;
> }
> 
> interface DocumentStore<Handle, Version> {
>     /* ... */
> 
>     createDraft(handle: Handle): Handle;
> 
>     commitDraft(handle: Handle, draft: Handle): Commit<Version>;
> 
>     revertCommit(handle: Handle, commit: Commit<Version>): void;
> }
> ```
> 
> For usage with Automerge our implementation could be similar to this:
> 
> ```ts
> import { type Heads, diff, applyPatches, getHeads } from "@automerge/automerge";
> import { type DocHandle, Repo } from "@automerge/automerge-repo";
> 
> // No need to pollute our synced repo with drafts.
> const draftRepo = new Repo();
> 
> const automergeStore: DocumentStore<DocHandle, Heads> = {
>     /* ... */
> 
>     createDraft: (handle) => {
>         return draftRepo.clone(handle);
>     },
> 
>     commitDraft: (handle, draft) => {
>         const before = getHeads(handle.doc());
>         handle.merge(draft);
>         const after = getHeads(handle.doc());
>         return { before, after };
>     },
> 
>     revertCommit: (handle, commit) => {
>         const inverse = diff(handle.doc(), commit.after, commit.before);
>         handle.change((doc) => {
>             applyPatches(doc, inverse);
>         });
>     },
> };
> ```

tslil-topos: thanks, i did not see that sketch. I also think it makes sense to have this as something that exists on a store as opposed to between stores. I'll think about what this means for the in-memory version and have a go.

kasbah: I had a PoC where the in-memory version of this just made a copy as the "draft" and reset to copies on revert. `Version` was also just a full copy of the document JS object.

tslil-topos:
turns out in testing that we need to:
* return more information in the retryable branch
* generalise the callback or the ui will miss streaming tool calls
* be less strict about mixing prose and narration in a given inference cycle

i will additionally be implementing the transaction stuff in catcolab-documents for this change, as per the discussion above

packages/frontend/src/llm_conversation/document_transaction.ts
kasbah: > ```ts
>     commitDraft: (handle, draft) => {
>         const before = getHeads(handle.doc());
>         handle.merge(draft);
>         const after = getHeads(handle.doc());
>         return { before, after };
>     },
> ```

I think this is not technically correct in the case of a remote edit while we are merging. Probably more correct to make another draft first and get the `after` from that.

kasbah: Although, that may result in `Heads` that are not in the actual doc :thinking:

tslil-topos:
looking at the complexity of this change and the stack, and seeing the uncertainty in the surface area of transactions around atomicity/concurrency i think the best thing to do for now is to leave the current very naive implementation in this change, and once we have answers to questions we can put a new change on-top that follows a careful design we believe in.

i also think that this change is already doing several things at once, so let's separate out that revision for another PR

packages/logics/src/simple-schema.ts
```diff
@@ -22,3 +22,11 @@ export const SimpleSchema = defineShape({
...
     },
 });
+
+export const simpleSchemaCellTypes = {
+    schema_entity: Entity,
+    schema_attr_type: AttrType,
+    schema_mapping: Mapping,
+    schema_attr: Attr,
+    schema_rich_text: RichText,
+};
```
kasbah, line 32: We discussed offline that this should use consistent names with what we will use in documentation etc.

```ts
export const simpleSchemaCellTypes = { Entity, AttrType, Mapping, Attr, RichText };
```

In the future we should change all the `define*` helpers to take a name string as the first argument and auto collate these from that. If the shapes have names too it will help for error messages. We currently have some unhelpful `Shape "unnamed"` in error messages.

kasbah (changes requested):
One minor change and I'm ready to accept. I haven't looked too closely at the pseudo-transactions since work to replace that is already in progress.

packages/logics/src/simple-schema.ts
tslil-topos: i agree, we should force people to name things meaningfully and then we can use those in a variety of ways. for now i have dropped the explicit keys from these mappings.

kasbah: i saw these were pushed to a follow-on stacked PR. i am assuming that was intentional.

kasbah (approved):

packages/logics/src/simple-schema.ts
tslil-topos: haha, it was not intentional. Oh well, it's in the stack. Thanks

PR #1425, opened 2026-08-25
Update vite and vitest dependencies

kasbah:
Mostly to align our dependencies across our packages.


tslil-topos (approved):
fantastic, thank you

PR #1430, opened 2026-08-25
Document binder for frontend

kasbah:
This adds an Automerge+our-backend binder for our frontend and a context for it. This does not yet have a fine-grained reactive view of the document. 

This also wires the instance document type through to everywhere, including some compatibility with the existing `liveDoc` way of doing things.

There was one place that revealed that doing things a normal JS way doesn't necessarily work with the Automerge proxy.


tslil-topos (approved):
i guess in the distant future we'll port everything over to the API, but in the mean time the LiveInstanceDoc definitely seems like the geodesic path. I'll follow this example for the LLMConversation stuff, thanks

PR #1433, opened 2026-08-26
Add `ElaboratedModel` type that mirrors notebook API

kasbah:
This gives the elaborated model an interface with `judgments()` and `judgmentsOf()` that mirror the notebook cell methods.

The reason I am adding this now is that it seems to be the first step in a path to get a non-hacky reactive view of our instance tables.


tslil-topos (approved):
We discussed this IRL, my sense is that some part of this should move to catlog @epatters to avoid duplicating ideas about models and judgements in the FE, but i don't currently have the bandwidth to make a concrete proposal

PR #1434, opened 2026-08-27
Reactive view for frontend binder

kasbah:
Adds a fine-grained reactive view to documents via `@automerge/automerge-repo-solid-primitives` `makeDocumentProjection`. The projection already strips the Automerge proxies so we only need to strip the Solid proxies for `copyValue`. I was unsure of this so I added a bunch of tests around copying values.


tslil-topos (approved):
thanks, this was definitely one of the missing pieces

PR #1435, opened 2026-08-27
Add a reactive validation view to notebooks and instances

kasbah:
Extends the document store with:

```ts

export interface ReactiveView<T extends object> {
    readonly current: Readonly<T>;
    replace(next: T): void;
}

export interface DocumentStore<Handle> {
    /* ... */
    // Create a reactive view for values derived from documents. Stores may
    // use this hook to integrate their host application's reactive primitives.
    createReactiveView?<T extends object>(initial: T): ReactiveView<T>
    
}    
```

and the notebook interface with:

```ts
/** A live view of a notebook's validation state.

While the view is active, the notebook revalidates whenever its document
changes; `model` and `issues` reflect the latest outcome. The caller must
dispose the view when it is no longer needed.

Elaboration and validation are separate steps, so `model` is always available:
it reflects the latest elaboration even when validation fails. It is empty
only before the first elaboration completes or when elaboration itself fails. 

In the case of an empty model the `model.judgments` and `model.judgementsOf` methods 
return empty arrays. */
export interface ValidationView<out S extends Shape> {
    readonly model: ElaboratedModel<S>;
    /** Validation issues; empty when the notebook is valid. */
    readonly issues: ReadonlyArray<Issue>;
    dispose(): void;
}

export interface Notebook<S, /* ... */> {
    /* ... */
    /** Create a live, reactive view of the notebook's validation state. The
     * caller must dispose the view when it is no longer needed. */
    createValidationView(): ValidationView<S>;
}

```

The primary use case of this is for use with SolidJS components, e.g. our completions, and for implementing a reactive view of instance tables that depends on the reactive view of the validated schema (next PR).


.github/workflows/ci.yml
```diff
@@ -124,13 +124,7 @@ jobs:
...
       - name: Typescript documents package tests
         run: |
-          pnpm --filter catcolab-documents exec vitest run \
-            test/creating-and-editing.test.ts \
-            test/definitions.test.ts \
-            test/binder.test-d.ts \
-            test/llm-conversation.test.ts \
-            test/instances.test.ts \
-            test/serialization.test.ts \
+          pnpm --filter catcolab-documents test
```
kasbah, line 127: Turns out vitest doesn't care if you import non-existent things.

kasbah:
Updated the title to include instances since the whole stack was merged into here.

tslil-topos (approved):
after sifting through the disaster that i created here by merging kaspar's entire stack into this one commit, my stance on this reduces to: does this satisfy @epatters needs for the reactive view?

the rest of the stuff about instance/schema and the useful way to pair these with invalid state kaspar and i discussed IRL and i think this PR is a good start.

I expect that we'll want to think carefully about this design pattern because we'll have to deal with dependent types in several more places in the api in forthcoming expansions (eg analyses).

Still there is something special about tabular data where it's more useful to say "here's everything" than "here are the valid things" + "here are some unlabelled columns that we don't know how to handle now".

epatters (approved):
There's a lot here that I don't really understand and at some point I'll ask you or Kaspar to walk me through this stuff. But now is not that time. I'm going to approve so that you can keep making progress.

PR #1436, opened 2026-08-28
UI for LLM conversation

epatters:
Features carried over from tslil's [prototype UI](https://github.com/ToposInstitute/CatColab/tree/tslil/llmconversation_ui_proto):

- [x] Conversation controller handling the event loop for a turn
- [x] Sending messages and displaying responses in a transcript
- [x] Markdown rendering of messages received from LLM
- [x] Displaying code executions
- [x] Displaying status and notes/errors from LLM harness
- [x] Attaching files to user messages [done in #1456]
- [ ] Handling feedback requests [implemented but not yet tested or styled]
- [ ] Retrying failed turns
- [ ] Read-only conversations


packages/frontend/src/llm_conversation/conversation_editor.tsx
```diff
@@ -0,0 +1,88 @@
...
+}) {
+    const controller = createLLMConversationController(
+        () => props.conversation,
+        () => props.documentStore,
+        useInferenceKey(),
+    );
+
+    const conversationDoc = () => {
+        const doc = props.conversation.handle.docView;
+        invariant(doc.type === "llmconversation");
```
epatters: @kasbah, re: #1434, this doesn't feel right: I shouldn't have to make this type assertion when I already know that the doc view belongs to an `LLMConversation`.

packages/frontend/src/llm_conversation/conversation_controller.ts
```diff
@@ -0,0 +1,123 @@
...
+    state: LLMTurnState;
+
+    /** Run a turn of the LLM conversation. */
+    runTurn: (userInput: LLMConversationUserInput) => Promise<LLMConversationTurnResult>;
+};
+
+export function createLLMConversationController<
+    Handle,
+    Attachment extends LLMConversationAttachment<Shape, Handle>,
+>(
```
epatters: Having to know and explicitly state these type parameters just for the sake of functions far down the call stack is a little painful. Is the generality enabled by these type parameters doing any work for us?

packages/frontend/src/llm_conversation/conversation_editor.tsx
kasbah: `.handle` was never meant to be part of the public API. Since we seem to need it on the doc we should rename it to `._handle` to indicate it's supposed to be private. You may want `.document` or probably even just `.interactions()`.

packages/frontend/src/llm_conversation/conversation_editor.tsx
```diff
@@ -0,0 +1,88 @@
...
+import type { ApiLLMConversation } from "./live_doc_compatibility";
+
+/** Form data for a message to send to the LLM. */
+type LLMMessageForm = {
+    message: string;
+};
+
+export function LLMConversationEditor(props: {
+    conversation: ApiLLMConversation;
+    documentStore: ApiDocumentStore;
```
kasbah: What I was envisioning is that the thing to be passed around is the binder not the store. The binder should be available via its context i.e. `useBinder`.

packages/frontend/src/llm_conversation/conversation_controller.ts
kasbah: I can't fully tell why this needs to be here. @tslil-topos might be better equipped to answer this.

packages/frontend/src/llm_conversation/conversation_editor.tsx
epatters: Are `.document` and `.interactions()` reactive for use in Solid components?

kasbah: They should be, yes.

packages/frontend/src/llm_conversation/conversation_controller.ts
kasbah: It's possible that this is resolved by switching to the binder, i.e. https://github.com/ToposInstitute/CatColab/pull/1436#discussion_r3880368657. Then at least we shouldn't need to care about `Handle`.

packages/frontend/src/llm_conversation/conversation_editor.tsx
epatters: Thanks for clarifying. Fixed in 795d553ca1dff864b8eba05f261f9e807b91d9ab

packages/frontend/src/llm_conversation/conversation_controller.ts
epatters: That did indeed resolve this nit too!

kasbah (approved):

PR #1437, opened 2026-08-28
Make elaborated model always available from validation

kasbah:
Based on #1435 but didn't want to sit on top of @epatters WIP #1436 on the stack.

This adds `ModelValidation` which is the validation type from #1435 minus the reactivity. It always has the elaborated model and an array of issues. We no longer return a `Result` from validation, instead we always return a `ModelValidation`. 

```diff
-    validate(): Promise<Result<ElaboratedModel<S>>>;
-    onValidate(callback: (result: Result<ElaboratedModel<S>>) => void): () => void;
+    validate(): Promise<ModelValidation<S>>;
+    onValidate(callback: (result: ModelValidation<S>) => void): () => void;
```

```ts
export interface ModelValidation<out S extends Shape> {
    readonly model: ElaboratedModel<S>;
    /** Validation issues; empty when the notebook is valid. */
    readonly issues: ReadonlyArray<Issue>;
}
```

### DIFF
--- a/packages/frontend/src/llm_conversation/scoped_document.ts
+++ b/packages/frontend/src/llm_conversation/scoped_document.ts
@@ -6,7 +6,7 @@ export type DocumentBinding<Handle = unknown, E extends Issue = Issue> = {
     readonly document: Readonly<Document>;
     readonly handle: Handle;
     readonly title: string;
-    validate(): Promise<Result<unknown, ReadonlyArray<E>>>;
+    validate(): Promise<Result<unknown, ReadonlyArray<E>> | { issues: ReadonlyArray<E> }>;
 };
 
 export type ScopedDocumentLink = {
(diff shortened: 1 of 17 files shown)


packages/frontend/src/llm_conversation/scoped_document.ts
```diff
@@ -6,7 +6,7 @@ export type DocumentBinding<Handle = unknown, E extends Issue = Issue> = {
     readonly document: Readonly<Document>;
     readonly handle: Handle;
     readonly title: string;
-    validate(): Promise<Result<unknown, ReadonlyArray<E>>>;
+    validate(): Promise<Result<unknown, ReadonlyArray<E>> | { issues: ReadonlyArray<E> }>;
```
kasbah, line 9: In the next PR I am going to thread this through to the instance validation at which point I'll be able to clean this and some other code up.

epatters:
> Based on #1435 but didn't want to sit on top of @epatters WIP #1436 on the stack.

Can the stack be reordered? Should I not have put my draft PR onto the stack? Haven't used this feature before and I'm not sure what the protocol is.

PR #1443, opened 2026-09-01
OBEE-44: transactions and plumbing this into the LLM scope

tslil-topos:
# What is this

A """"real"""" transaction system implemented at the store and binder level. Stores are required to know how to create drafts, deal with `DocumentChange`s, and partition their resolution space. The binder implementation lives atop this and its only contribution on top of tracking this data is to ensure that depedent supported documents point to their draft parents if those were also present in the transaction scope.

We also take the opportunity to factor out an enum of everything that the document api supports, for now it reads.

```typescript
export type SupportedDocument<S extends Shape, H, V> =
    | Notebook<S, ModelDocument, H, V>
    | Instance<H, S, V>;
```

# Why the divergance from the original sketch

Separation of concerns, in our use case the LLM scope is an advesary and giving it the ability to commit its changes back to the live document would break our "enforce validity of models" rule.

# How to use it

Stage a set of documents for atomic editing:

```typescript
const { tx, draftDocs } = await binder.beginTransaction({ schema, instance, etc });
```

`draftDocs` has the same shape as the input record, but each document is a draft: edits through it are invisible through the originals until the transaction ends through one of two ways

```typescript
tx.commit();  // merges every draft back into its source; returns a Commit
tx.abort();   // discards every draft, leaving the sources untouched
```

The returned `Commit` records the changes made to each source document and can be undone per document: `schema.revert(commit)`.

Documents staged together see each other's edits: an instance staged with its schema is bound to the schema's draft, so schema edits are visible through the draft instance's tables and validation, while the original instance continues to see the committed schema until commit. Documents outside the transaction resolve as usual, so an instance staged alone is bound to the real schema, live-updating.

Drafts are ordinary documents as far as the store is concerned: a draft's own ref resolves to the draft through `getHandle`, while its source's ref keeps resolving to the committed document, and the draft's ref stops resolving once the transaction ends.

Out of scope for this change is document creation or deletion during a transaction, a possible design for this would be to have a draftBinder returned from `binder.beginTransaction(..)` that would be safe to use.

# Notes

This had to sprinkle another generic parameter through the codebase, `V` for `Version` so the change touches many files but most in a relatively minor way.

Some tests are rewritten, others have been added.


packages/frontend/src/api/document_store.ts
```diff
@@ -110,5 +131,35 @@ export function createApiDocumentStore(api: Api): ApiDocumentStore {
...
+        },
+        discardDraft: (draft) => {
+            handles.delete(draft.automergeHandle.documentId);
+            draftRepo.delete(draft.automergeHandle.documentId);
+        },
+        revertCommit: (handle, change: DocumentChange<Heads>) => {
+            // Trust automerge to figure this out.
+            const inverse = diff(handle.automergeHandle.doc(), change.after, change.before);
+            handle.automergeHandle.change((doc) => {
+                applyPatches(doc, inverse);
```
kasbah, line 170: Just making a note that this will likely be broken on deletions (or I guess creation since it's the inverse!) till https://github.com/automerge/automerge/pull/1308 is merged and released. ~~We can pnpm patch it in automerge too but my patch on #1448 only fixes it for `automerge-solid-primitives` since that bundles its own copy of `applyPatches`.~~

On second thought, I'll add a patch to automerge on #1448

packages/documents/test/transactions.test.ts
```diff
@@ -1,32 +1,157 @@
 import { Entity, SimpleSchema } from "catcolab-logics/simple-schema";
 import { describe, expect, test } from "vitest";
 
-// RFC-0006 "Transactions": a set of changes staged on a draft, invisible on the
-// source notebook until committed, and revertible as a whole via the returned
+// RFC-0006 "Transactions": a set of changes staged on drafts, invisible on the
```
kasbah:
```suggestion
// Transactions: a set of changes staged on drafts, invisible on the
```

packages/frontend/src/api/document_store.ts
```diff
@@ -21,10 +23,13 @@ export type ApiDocumentHandle = {
     ref: DocumentRef;
 };
 
-export type ApiDocumentStore = DocumentStore<ApiDocumentHandle>;
+/** The version type of the API document store: Automerge heads. */
+export type ApiDocumentVersion = Heads;
```
kasbah, line 27: What do we need this alias for?

kasbah (changes requested):
Two minor points while reading through.

The more important potential issues, around atomicity and with computing the inverse as we do, were discussed off-thread and we'll continue to look at that after merge.

packages/frontend/src/api/document_store.ts
tslil-topos: this is part of the newtype design paradigm that i inherited from (actually strongly, statically) typed languages: even if you are dealing with a plain type `T` that already exists but you are using it in a way that you don't want to confuse with other usages of `T` then you should alias it.

in this case it's true that we're using automerge heads, but we're storing pairs of them and adding our own layer on-top of them so that the "attitude" of the usage is different --- it's incidental that you could put any old head into our model, and not intended.

kasbah: ok

kasbah (approved):

PR #1444, opened 2026-09-01
Allow instances to be created from partial schemas

kasbah:
Minor change but a breaking to the API so thought I would ask @tslil-topos to sign-off.


tslil-topos (approved):
lgtm

Issue #1426, opened 2026-08-25
Validate underlying JSON before manipulating documents


kasbah:
From https://github.com/ToposInstitute/CatColab/pull/1413#discussion_r3831017617

> we should parse our doc into a structure where we know:
> 
> - `table.rowOrder` is not undefined
> - `table.rowOrder` has all the keys of `table.rows` and no more
> - any other invariants we want to hold for the JSON structure

This is applicable to notebooks as well.

PR #1445, opened 2026-09-02
OBEE-47 equations in the API, not participating in validation

tslil-topos:
The main contribution is `equation.ts` which adds `EquationSide`, `EquationCell`. We can now do

```typescript
const equation = notebook.add(PathEquation, {
    label: "employment",
    lhs: [worksIn, partOf],
    rhs: [employer],
});

let foo = notebook.cellsOf(PathEquation);

if (notebook.supports(PathEquation)) { .. };
```
etc.

Tests are added, PathEquation is injected into the LLM scope.

While doing this i noticed some things that we should probably work towards fixing in the future:
* there are four conventions for failure modes in documents/, and it's not obvious to me that this was planned or is demarcating a useful distinction for callers.
  * `undefined` (eg `tryGetStoredRichTextCell`)
  * `null` (eg deleted morphism endpoints)
  * `throw` (eg `tryGetModelJudgement`)
  * `Result` (eg `loadNotebookFromRef`)
* documents owns a lot of fiddling with the underlying Rust types in a way that probably should be abstracted by document-methods/document-types.
* Rust has some way to go on the equation front:
  * `InvalidDblModel::Eqn` and `Feature::PartialEquation` don't carry the equation id so we can't say anything useful in the API
  * equations in modal & tabulator models
  * ModelPresentation **doesn't** actually include equations yet

This last part means that for now `judgmentsOf(PathEquation)` returns []. The next change will tackle surfacing these equations and validation of instances subject to equations.


kasbah:
> there are four conventions for failure modes in documents/, and it's not obvious to me that this was planned or is demarcating a useful distinction for callers.
> 
> * `undefined` (eg `tryGetStoredRichTextCell`)
> 
>* `null` (eg deleted morphism endpoints)
> 
>* `throw` (eg `tryGetModelJudgement`)
> 
>* `Result` (eg `loadNotebookFromRef`)

That first one really shouldn't exist, which is part of what I was getting at with https://github.com/ToposInstitute/CatColab/issues/1426 

The second one seems unavoidable because I don't know any better way to handle deletion.
   
The 3rd and 4th are deliberate and the delineation from the RFC that we decided on was:

> At runtime invalid inputs:
> * Should throw an exception, ideally carrying a Standard Schema compatible issues array, if it is something TypeScript would have caught
> * Should return a Result type otherwise

kasbah (approved):

PR #1447, opened 2026-09-02
Allow editing instance when schema is partially valid

kasbah:
Another minor one but breaking so need input from @tslil-topos


tslil-topos (approved):
yeah, this seems like a natural consequence of the previous change

PR #1450, opened 2026-09-02
OBEE-49 Rust names equations in errors, tie equations into TS Judgement

tslil-topos:
~This is a bit of a judgement call given that it's a new feature in catlog, but the place that seemed most reasonable to insert names for equations was directly into PathEq. I'm prepared to defend this mathematically (after all, equations are data and the only part of f.p. category data that are not presently named, not to mention the HoTT perspective ...) but i'd also like to draw attention to the pragmatic perspective: tracking what certain generators are called in a layer above the data of the generators themselves is bound to cause issuse.~

~It's worth being clear that these names are **not** in practice the *labels* that the notebook supplies, they're the QualifiedName (so some UUID) of the cell in the notebook. It's up to the typescript to figure out whether the cell has a label.~

~The rest of this change tries to be surgical in having catlog report names of equations that are causing errors at the typescript interface, where we are sure that we received UUIDs to begin with.~

See #1464 

On the typescript side it's more of the same that we've seen it the mirror universe, tying this into the judgement structure so that future code can figure use validated models as a source of truth for equations.

### DIFF
--- a/packages/catlog/src/tt/notebook_elab.rs
+++ b/packages/catlog/src/tt/notebook_elab.rs
@@ -181,7 +181,11 @@ impl<'a> Elaborator<'a> {
                 nb::path::Path::Id(ob) => {
                     let (stx, val, ob_type) = self.ob_syn(ob)?;
                     let mor_type = self.theory().hom_type(ob_type)?;
-                    Some((stx, val.clone(), TyV::morphism(mor_type, val.clone(), val.clone())))
+                    Some((
+                        stx,
+                        TmV::id(val.clone()),
+                        TyV::morphism(mor_type, val.clone(), val.clone()),
+                    ))
                 }
                 nb::path::Path::Seq(ms) => match ms.as_slice() {
                     [] => None,
--- a/packages/documents/src/model/elaborated-model.ts
+++ b/packages/documents/src/model/elaborated-model.ts
@@ -1,13 +1,16 @@
+import type { Mor } from "catcolab-document-types";
 import type { ModelPresentation, MorGenerator, QualifiedLabel } from "catlog-wasm";
 import type { Issue } from "../result";
 import {
     findMorphismType,
     findObjectType,
     type AnyCellType,
+    type EquationType,
     type MorphismType,
     type MorphismTypesOf,
     type ObjectType,
     type ObjectTypesOf,
+    type RichTextType,
     type Shape,
 } from "../shape";
 import { morphismTypesEqual, objectTypesEqual } from "./equality";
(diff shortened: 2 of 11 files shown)


packages/catlog/src/tt/notebook_elab.rs
```diff
@@ -181,7 +181,11 @@ impl<'a> Elaborator<'a> {
                 nb::path::Path::Id(ob) => {
                     let (stx, val, ob_type) = self.ob_syn(ob)?;
                     let mor_type = self.theory().hom_type(ob_type)?;
-                    Some((stx, val.clone(), TyV::morphism(mor_type, val.clone(), val.clone())))
```
tslil-topos, line 184: i believe that this was a pre-existing bug, though nothing could have exercised it afaict?

packages/catlog/src/dbl/discrete/model.rs
```diff
@@ -53,13 +53,23 @@ impl DiscreteDblModel {
         self.category.add_equation(eq);
     }
 
+    /// Iterates over equations between morphisms, with their names.
+    pub fn named_equations(
```
tslil-topos: maybe the base trait should flip to this or this should be relocated into the base trait

tslil-topos:
recording that @epatters will take a look at the catlog core stuff and potentially move things around (we had discussed f.p. category owning the labels for equation rather than PathEq)

tslil-topos:
thanks to @epatters in #1464 this change now reaches no deeper than catlog-wasm to add equations to the representation

packages/documents/src/model/elaborated-model.ts
```diff
@@ -35,17 +38,46 @@ export interface MorphismJudgment<
...
 The API mirrors the notebook's `cells`/`cellsOf`: judgments are returned in
-presentation order (objects first, then morphisms). */
+presentation order (objects first, then morphisms, then equations). */
 export interface ElaboratedModel<out S extends Shape> {
     judgments(): ReadonlyArray<JudgmentOf<S>>;
+
+    /** Judgments of the given cell type or shape, in presentation order.
+
+    Filtering by a specific cell type narrows the result accordingly. */
+    judgmentsOf(type: RichTextType): ReadonlyArray<never>;
```
kasbah, line 77: Does this need to be defined for some reason?

kasbah (approved):

packages/documents/src/model/elaborated-model.ts
tslil-topos: it does not, but it's a convenient static analysis guide for the code: if you iterate on this result tsc will narrow to never and you know you probably should not be doing this

PR #1452, opened 2026-09-03
Our own table editor component

kasbah:
Replaces the handsontable-mit table editor with one more specifically designed to work with our instances of schemas and with SolidJS.

https://branch-kb-table-component-nih--catcolab.netlify.app/dev/ui-components/?path=/docs/forms-inputs-tableeditor--docs


tslil-topos:
i've played around with this, some minor notes:
* deleteRow (table_editor.tsx:499-505) records the deleted row's index regardless of where selection was, so we can end up jumping to the next row when the focus was elsewhere --- presumably not intentional
* when changing a reference (to another row) the drop-down highlights the first entry (alphabetically?) rather than the current selection so "click -> enter" changes the value rather than does nothing.

tslil-topos (changes requested):
with those two minor ui points i think this is g2g

kasbah:
I believe I've addressed both points and the UUID vs index point from #1372.

tslil-topos (approved):
thanks for the changes

PR #1458, opened 2026-09-04
OBEE-46 system prompt containing explanation of api surface, glm 5.3

tslil-topos:
We will need several iterations of this once the UI and substack are merged so that we can experiment with the demos we actually want


epatters (commented):
Thanks, I look forward to playing around with an improved system prompt.

Assuming we'll continue to have a system prompt that includes a big static string, I'd suggest moving it into a standalone Markdown file, which will improve human readability and editability. Though I haven't tried it, Vite appears to natively support loading external files as strings: https://vite.dev/guide/assets#importing-asset-as-string

tslil-topos:
> Thanks, I look forward to playing around with an improved system prompt.
> 
> Assuming we'll continue to have a system prompt that includes a big static string, I'd suggest moving it into a standalone Markdown file, which will improve human readability and editability. Though I haven't tried it, Vite appears to natively support loading external files as strings: https://vite.dev/guide/assets#importing-asset-as-string

turns out this was straightforward with vite, thanks

packages/frontend/src/inference/system_prompt.md
```diff
@@ -0,0 +1,114 @@
...
+- `MistypedRowRef` --- a row reference points to a row of the wrong table.
+- `OrphanedField` --- a stored field has no matching column in the schema.
+- `OrphanedTable` --- a stored table has no entity in the schema.
+- `EquationViolation` --- a row violates a path equation of the schema; `equationId` identifies the equation. At most 10 counterexamples are reported per equation, followed by a summary issue on the table.
+
+Column typing: a column's type comes from the schema morphism's codomain. A codomain object labeled `"Bool"`, `"Int"`, `"Float"`, or `"String"` gives a column of that literal type; any other label gives a `String` column. A codomain that is another table entity gives a row-reference column.
+
+Example --- populate tables:
+```js
+const { tables } = await inst.validate();
```
kasbah: May be a good idea to always show getting the issues and logging them whenever calling validate.

packages/frontend/.prettierignore
```diff
@@ -0,0 +1,2 @@
+# exempt system prompt from code formatting
+src/inference/system_prompt.md
```
kasbah: I think it's fine to have oxfmt format the markdown.

We are not using prettier so this should go in some ox* config if it is needed.

kasbah (changes requested):

packages/frontend/.prettierignore
tslil-topos: i'm not sure what you mean? this file toggled the CI behaviour correctly

kasbah: I think that's oxfmt going the extra mile. Like I said, I think it's better to auto-format the file anyway.

packages/frontend/src/inference/system_prompt.md
tslil-topos: done, also added an example of the batch api usage

packages/frontend/.prettierignore
tslil-topos: i disagree, this file can have arbitrary content including but not limited to mathematics and data (eg csv). I don't think a code formatter has any place touching mixed prose and code in potentially many languages.

tslil-topos: According to https://oxc.rs/docs/guide/usage/formatter/language-support.html oxfmt delegates to prettier for markdown (for now at least). So at the very least this is consistent with their usage guide.

I did some searching, there _were_ issues like https://github.com/prettier/prettier/issues/16664 but this appears fixed. So as long as we're on a version past June/July this should be a non-issue. In that case i'm ok to opt this into formatting until we run into something. I'll remove this exemption.

kasbah: I wasn't aware of any of this so good to know, thanks.

kasbah (approved):

PR #1459, opened 2026-09-04
OBEE-51 implement listUsedBy, listDependsOn to store, significantly clean up LLM coversation scope code

tslil-topos:
Thanks to @kasbah for the push on this, the store now has

```typescript
export type HandlesByLinkType<Handle> = Record<LinkType, ReadonlyArray<Handle>>;
...
export interface DocumentStore<Handle, Version = unknown> {
...
    // List the documents that depend on the document at `handle`, indexed by
    // the type of the link through which each depends on it.
    listUsedBy(handle: Handle): Promise<HandlesByLinkType<Handle>>;
    // List the documents that the document at `handle` depends on, indexed by
    // the type of each of its links.
    listDependsOn(handle: Handle): Promise<HandlesByLinkType<Handle>>;
...
}
```

and supporting machinery, as well as implementations for the two existing stores. This move required shuffling around the context containing the fe's binder, and introducing a binder context provider.

This change makes scope building for the LLM use these two functions and pull in the "star" of dependencies around the attachment point: all those things that the store thinks are used by the attachment point and upon which the attachment point depends.

The uniform approach opened my eyes to a bunch of pointless over-specialisation in the code paths for the llmconversation. So this change:
* removed LLMConversationAttachment type, it's now SupportedDocument (which includes LLMConversation) -- in code it's possible to attach a conversation to a conversation now (though by UI it's not possible)
* add a recursive helper in the FE whose only job is handle -> supported document so that we can cross from store to binder
* significantly reduced execution_scope to a uniform system in the presence of this new store API: pull in the "star" about the attachment point (filtering out the current conversation itself), throw all of this into a transaction, make some strings describing the relationships

As a final simplification, instanceFromStore took a shape parameter, even though it also took a schema parameter and that schema is authoritative on the shape.

### DIFF
--- a/packages/documents/src/document-store/document-store.ts
+++ b/packages/documents/src/document-store/document-store.ts
@@ -39,6 +39,11 @@ export interface DocumentStore<Handle, Version = unknown> {
     getDocumentRef(handle: Handle): DocumentRef;
     // Get a document view from a handle
     getDocumentView(handle: Handle): Readonly<Document>;
+    // List the handles of all instance documents that link back to the
+    // document at `handle`, that is, all instances of it. Matching is by
+    // document id, so instances of any version of the document are included.
+    // The order of the results is unspecified.
+    listInstancesOf(handle: Handle): Promise<ReadonlyArray<Handle>>;
     // Create a working copy of the document at `handle`. Changes made to the
     // draft are not visible through `handle` until it is committed.
     createDraft(handle: Handle): Handle;
--- a/packages/documents/src/document-store/document-store.ts
+++ b/packages/documents/src/document-store/document-store.ts
@@ -39,11 +39,11 @@ export interface DocumentStore<Handle, Version = unknown> {
     getDocumentRef(handle: Handle): DocumentRef;
     // Get a document view from a handle
     getDocumentView(handle: Handle): Readonly<Document>;
-    // List the handles of all instance documents that link back to the
-    // document at `handle`, that is, all instances of it. Matching is by
-    // document id, so instances of any version of the document are included.
-    // The order of the results is unspecified.
-    listInstancesOf(handle: Handle): Promise<ReadonlyArray<Handle>>;
+    // List the handles of all child documents that link back to the document
+    // at `handle`, that is, all instances of it. Matching is by document id,
+    // so children of any version of the document are included. The order of
+    // the results is unspecified.
+    listChildren(handle: Handle): Promise<ReadonlyArray<Handle>>;
     // Create a working copy of the document at `handle`. Changes made to the
     // draft are not visible through `handle` until it is committed.
     createDraft(handle: Handle): Handle;
--- a/packages/documents/src/document-store/document-store.ts
+++ b/packages/documents/src/document-store/document-store.ts
@@ -1,10 +1,44 @@
-import type { Document } from "catcolab-document-types";
+import type { Document, Link, LinkType } from "catcolab-document-types";
 import type { Result } from "../result";
 
 export interface DocumentRef {
     id: string;
     version: string | null;
     server?: string;
+}
+
+/** Handles of documents related to a document, indexed by the type of the
+ * link relating them. Every link type is present; a type with no matches
+ * maps to an empty array. */
+export type HandlesByLinkType<Handle> = Record<LinkType, ReadonlyArray<Handle>>;
+
+/** The links a document has to other documents. */
+export function documentLinks(document: Document): Link[] {
+    switch (document.type) {
+        case "model":
+            return [];
+        case "diagram":
+            return [document.diagramIn];
+        case "instance":
+            return [document.instanceOf];
+        case "analysis":
+            return [document.analysisOf];
+        case "llmconversation":
+            return [document.llmConversationOf];
+    }
+}
+
+/** An index of handles by link type with no handles in it. The arrays are
+ * fresh, so store implementations can fill them in before answering a link
+ * query. */
+export function emptyHandlesByLinkType<Handle>(): Record<LinkType, Handle[]> {
+    return {
+        "analysis-of": [],
+        "diagram-in": [],
+        "instance-of": [],
+        "llmconversation-of": [],
+        instantiation: [],
+    };
 }
 
 /**
@@ -39,11 +73,12 @@ export interface DocumentStore<Handle, Version = unknown> {
     getDocumentRef(handle: Handle): DocumentRef;
     // Get a document view from a handle
     getDocumentView(handle: Handle): Readonly<Document>;
-    // List the handles of all documents that depend on the document at
-    // `handle`, that is, all instances of it. Matching is by document id, so
-    // documents depending on any version of the document are included. The
-    // order of the results is unspecified.
-    listUsedBy(handle: Handle): Promise<ReadonlyArray<Handle>>;
+    // List the documents that depend on the document at `handle`, indexed by
+    // the type of the link through which each depends on it.
+    listUsedBy(handle: Handle): Promise<HandlesByLinkType<Handle>>;
+    // List the documents that the document at `handle` depends on, indexed by
+    // the type of each of its links.
+    listDependsOn(handle: Handle): Promise<HandlesByLinkType<Handle>>;
     // Create a working copy of the document at `handle`. Changes made to the
     // draft are not visible through `handle` until it is committed.
     createDraft(handle: Handle): Handle;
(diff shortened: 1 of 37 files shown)


packages/documents/src/document-store/document-store.ts
```diff
@@ -39,6 +39,11 @@ export interface DocumentStore<Handle, Version = unknown> {
     getDocumentRef(handle: Handle): DocumentRef;
     // Get a document view from a handle
     getDocumentView(handle: Handle): Readonly<Document>;
+    // List the handles of all instance documents that link back to the
+    // document at `handle`, that is, all instances of it. Matching is by
+    // document id, so instances of any version of the document are included.
+    // The order of the results is unspecified.
+    listInstancesOf(handle: Handle): Promise<ReadonlyArray<Handle>>;
```
kasbah: I would prefer that we add a more general `listChildren` and it lists sub-documents of all types.

I think even better would be get both directions i.e. add the ability to get what currently have in user state: `getUsedBy` and `getDependsOn`.

packages/frontend/src/api/document_store.ts
```diff
@@ -98,6 +112,33 @@ export function createApiDocumentStore(api: Api): ApiDocumentStore {
             };
         },
         getDocumentRef: (handle) => handle.ref,
+        async listInstancesOf(handle) {
+            // The user state document tracks backlinks between documents:
+            // every instance of this document appears in its `usedBy` relations
+            // with relation type "instance-of". Resolve those refs back into
+            // store handles like any other document ref.
+            const state = await getUserState(api);
```
kasbah: We should already have the user state in context and be able grab it from there. If there are issues with that then we probably just need to make sure the user state context is added before the binder context.

kasbah (changes requested):

packages/documents/src/document-store/document-store.ts
tslil-topos: we can *call* this function whatever we want, the implementation won't change for now: the only link we have is instances. I'm happy to rename it.

the other direction is easily reached now, i'd need to see a use-case for what amounts to obtaining the link and calling resolve on it

kasbah: that's fine but let's call it, and make the types reflect, something future-looking or we just need to change it later

the other direction is interesting for instantiations: this doc depends on all these docs for instantiations

packages/frontend/src/api/document_store.ts
tslil-topos: ok, the most obvious way to achieve this i think was to bring the binder creation from Root into user state, so that the context was visible. take a look at https://github.com/ToposInstitute/CatColab/pull/1459/commits/0b32dbe0afb8dcfbfe85a9ce81536cb4d6cc17c7

packages/documents/src/document-store/document-store.ts
tslil-topos: made the listChildrenOf change, i've added the other direction to the task for instantiations

kasbah: if we agree we will update this for instantiations then "children" is the wrong word unfortunately. something that instantiates another model is not a child of the model. the first model is "usedBy" the model that is instantiating it.

packages/frontend/src/api/document_store.ts
kasbah: i see, looks good. would be good to rename it to `MetaDocumentContext` or similar, to show it includes more than the user state.

packages/documents/src/document-store/document-store.ts
tslil-topos: makes sense, i'll think about some clear terminology

tslil-topos: Going with `listUsedBy` and `listDependsOn`

packages/frontend/src/user/meta_document_context.tsx
```diff
@@ -2,19 +2,19 @@ import type { PermissionInfo, UserInfo, UserState } from "catcolab-api/src/user_
 import { createContext, useContext } from "solid-js";
 import invariant from "tiny-invariant";
 
-export const INITIAL_USER_STATE: UserState = {
+export const INITIAL_META_DOCUMENT: UserState = {
```
kasbah: That one is still user state.

tslil-topos: thanks, that's what a for loop and sed get me. I'll double check

kasbah: Or in my view anyway the `MetaDocumentContext` provides `userState` and `binder`. Maybe later we will rename `userState` to `metaDocInfo` or something but that feels like outside this PR.

packages/documents/src/document-store/document-store.ts
kasbah: The title will need updating.

packages/frontend/src/user/meta_document_context.tsx
```diff
@@ -8,13 +8,13 @@ export const INITIAL_USER_STATE: UserState = {
...
-export function useUserState(): UserState {
-    const userState = useContext(UserStateContext);
-    invariant(userState, "User state should be provided as context");
-    return userState;
+/** Retrieve the meta document from application context. */
+export function useMetaDocument(): UserState {
+    const metaDocument = useContext(MetaDocumentContext);
+    invariant(metaDocument, "Meta document should be provided as context");
+    return metaDocument;
 }
```
kasbah: I proposed the rename to `MetaDocumentContext` when it seemed it needed to contain the binder as well. Since now nothing is changing about this context I don't think it should be renamed on this PR.

When we are ready to refactor this then we can rename the frontend and backend together and purge the name "user state" from the codebase. If we don't do that then we have confusing dual naming of the same thing.

packages/frontend/src/user/meta_document_provider.tsx
```diff
@@ -61,6 +67,10 @@ export function UserStateProvider(props: { children: JSX.Element }) {
     });
 
     return (
-        <UserStateContext.Provider value={userState}>{props.children}</UserStateContext.Provider>
+        <BinderContext.Provider value={binder}>
+            <MetaDocumentContext.Provider value={metaDocument}>
+                {props.children}
+            </MetaDocumentContext.Provider>
+        </BinderContext.Provider>
```
kasbah: It seems like these two contexts are now independent and should no longer be in the same file. I had thought this would need to be one context  where we have `<MetaDocumentContext value={{ userState, binder }} >`

I had proposed the rename to `MetaDocumentContext` when it seemed it needed to contain the binder as well. If nothing is actually changing about the context then it shouldn't be renamed on this PR.

When we are ready to refactor this then we can rename the frontend and backend together and purge the name "user state" from the codebase. If we don't do it all together then we have confusing dual naming of the same thing.

kasbah (changes requested):

packages/frontend/src/user/meta_document_provider.tsx
tslil-topos: i think that this is more about me not understanding solid than anything else, if you look [here](https://github.com/ToposInstitute/CatColab/pull/1459/changes#diff-506846a4bc0dfe38761c149592b074e51cb4e09f47646c111b73a747c6a03bc1R22) the binder is definitely intended to be a part of the same provider as the user state.

kasbah: yes, sorry, i think i spaced out on getting to the root of this when we talked about it before.

i think they should just be separate and we should defer the rename. the binder also depends on the api provided as context so this sort of "you'de better provide the right context when using this one" is unfortunately how we need to do it, otherwise we have to jumble them together like you did here.

so, like this:

```ts

export function BinderProvider(props: { children: JSX.Element }) {
    const binder = createApiBinder(useApi(), useUserState());

    return (
        <BinderContext.Provider value={binder}>
            {props.children}
        </BinderContext.Provider>
    );
}

```

tslil-topos: alrighty, there is now `packages/frontend/src/api/binder_provider.tsx` and it comes after the user state. About the only delta on this axis to main then is [here](https://github.com/ToposInstitute/CatColab/pull/1459/changes#diff-b8fd65e22fb61951fca8e70762da762f1f6b93f872da5751d7689b64a9a27766R11) where there's potentially a footgun if we leave this defaulted: the binder uses this particular piece of state and you'd never know that you accidentally made the binder context before the user state.

Unfortunately until we do a refactor correctness is a convention here, afaict, and it relies on the list ordering of multiprovider: binder comes after user state.

kasbah:
Looks like there is one minor merge conflict.

tslil-topos:
> Looks like there is one minor merge conflict.

given that the theme of this change is reworking with the benefit of hindsight, i'm glad we had this merge conflict. this part too can be simplified and unified. i wasn't going to do it originally, but there's no point in resolving the conflict for worse code.

kasbah (approved):

PR #1460, opened 2026-09-04
Press Enter to send in LLM conversation

kasbah:


tslil-topos (approved):

epatters (approved):
To close the loop on this: it seems that everyone except me prefers Enter to Shift+Enter, so we're going with it.

PR #1465, opened 2026-09-08
Allow creating LLM conversation from instance menu

kasbah:
Minor fixup that I'm likely to replace in a follow-on re-design of how LLM conversations fit in with documents.

### DIFF
--- a/packages/frontend/src/llm_conversation/live_doc_compatibility.ts
+++ b/packages/frontend/src/llm_conversation/live_doc_compatibility.ts
@@ -14,9 +14,11 @@ import type {
     ApiDocumentVersion,
     DocRef,
     LiveDoc,
+    LiveDocWithRef,
 } from "../api";
 import type { LiveModelDoc, ModelLibrary } from "../model";
 import { notebookShapes, shapeForTheory } from "../model/shapes";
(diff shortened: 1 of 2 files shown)


tslil-topos (approved):

packages/frontend/src/llm_conversation/live_doc_compatibility.ts
```diff
@@ -71,28 +73,58 @@ async function loadNotebook(api: Api, binder: ApiBinder, modelRefId: string) {
     return notebook.content;
 }
 
+async function loadInstance(
+    api: Api,
+    binder: ApiBinder,
+    instanceRefId: string,
+    modelRefId: string,
+) {
```
tslil-topos: given that this is temporary code it's not super important, but this would be nicer with a return type

kasbah: The code that is truly temporary is for the the menu items, some of this other code will stick around longer so please do review with that in mind. But yes, this is also for the compat layer that we do hope to eventually get rid of.

I added the type.

tslil-topos (approved):
hopefully no conflict with #1459

PR #1466, opened 2026-09-08
Show/hide tables and list them in sidebar

kasbah:
Lists the tables in the sidebar.

image

They can be minimized with a "-" button.

image

I haven't decided what should happen if you click a table again that is already being shown, so for now that does nothing at all.


tslil-topos (approved):
table2

PR #1467, opened 2026-09-08
LLM conversation pane

kasbah:
Adds a right-hand-side drawer where all the conversations are listed and made.

image


kasbah:
This is sort of incomplete in that you can still open the doc from "my documents" and it will not open in this pane but rather as its own doc.

tslil-topos (approved):
works for me, but we should make it possible to delete conversations somehow

PR #1472, opened 2026-09-09
Add buttons to delete orphaned data to table editor

kasbah:
Lets users delete the orphaned data. Also adds a new "outline" button modifier because the solidly filled one didn't look right.

https://branch-kb-delete-orphans-table-editor--catcolab.netlify.app/dev/ui-components/?path=/docs/forms-inputs-tableeditor--docs#orphaned-table


tslil-topos (approved):
works for me

PR #1475, opened 2026-09-09
Avoid incompatibility b/w Automerge's rich text and Solid integrations

### LINKED ISSUE #1474: Instances break when model notebook contains rich text
To reproduce:

1. Create an instance of a schema, where the schema notebook contains no rich text cells
2. Navigate to the schema-only page
3. Hard refresh
4. Add a rich text cell with a couple lines of text
5. Attempt to open the instance

Get the following error:

```
Error: target is not an array or object for patch
    Gm live_doc_compatibility.ts:82
error_boundary.tsx:9:27
```

epatters:
This PR works around #1474 by simply switching back to our old Automerge-Solid integration.

**Warning**: While I can confirm that this fixes #1474 for me, the diagnosis and test were LLM-generated, and I have not attempted to verify the precise origin of the problem, which seems to involve `automerge-repo-solid-primitives` being incompatible with rich text in Automerge but may run deeper than that. Of course it would be best to fix this upstream.

### DIFF
--- a/packages/frontend/src/api/document_store.test.ts
+++ b/packages/frontend/src/api/document_store.test.ts
@@ -1,7 +1,7 @@
 // @vitest-environment happy-dom
 // The happy-dom environment makes solid-js resolve to its reactive client
 // build, which the reactive projection tests depend on.
-import { getObjectId } from "@automerge/automerge";
+import { getObjectId, splice, splitBlock } from "@automerge/automerge";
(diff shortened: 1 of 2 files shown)


kasbah (approved):
Approving. I have some follow-ups starting with https://github.com/ToposInstitute/CatColab/pull/1478

PR #1476, opened 2026-09-10
Display row-level errors in instance editor

epatters:
The presence of errors is indicated by coloring the new row headers red and the errors themselves are displayed on hover in a tooltip.

image

### DIFF
--- a/packages/documents/src/instance/equation-validation.ts
+++ b/packages/documents/src/instance/equation-validation.ts
@@ -48,7 +48,9 @@ export function validatePathEquations<S extends Shape>(
             continue; // the source object has no table
         }
         const tableLabel = sourceTable.label ?? sourceId;
-        const equationName = judgment.label.join(".") || judgment.id;
+        const equationLabel = judgment.label.join(".");
+        const equationText =
+            equationLabel === "" ? "Unnamed equation" : `Equation \`${equationLabel}\``;
 
         let violations = 0;
         for (const row of sourceTable.rows) {
@@ -62,24 +64,26 @@ export function validatePathEquations<S extends Shape>(
             if (violations <= MAX_COUNTEREXAMPLES_PER_EQUATION) {
                 issues.push({
                     message:
-                        `Equation \`${equationName}\` is violated by a row of table \`${tableLabel}\`: ` +
+                        `${equationText} is violated by a row of table \`${tableLabel}\`: ` +
                         `left-hand side yields ${describeSideValue(left)}, ` +
                         `right-hand side yields ${describeSideValue(right)}`,
                     path: [sourceId, "rows", row.id],
                     issueType: "EquationViolation",
                     equationId: judgment.id,
+                    equationLabel: judgment.label,
                 });
             }
         }
         if (violations > MAX_COUNTEREXAMPLES_PER_EQUATION) {
             issues.push({
                 message:
-                    `Equation \`${equationName}\` is violated by ` +
+                    `${equationText} is violated by ` +
                     `${violations - MAX_COUNTEREXAMPLES_PER_EQUATION} more rows of table ` +
                     `\`${tableLabel}\``,
                 path: [sourceId],
                 issueType: "EquationViolation",
                 equationId: judgment.id,
+                equationLabel: judgment.label,
             });
         }
     }
--- a/packages/ui-components/src/table_editor/table_editor.stories.tsx
+++ b/packages/ui-components/src/table_editor/table_editor.stories.tsx
@@ -234,8 +234,8 @@ export const DoubleClickBooleanCell: Story = {
             name: "active",
         }) as HTMLTableCellElement;
         const activeColumn = activeHeader.cellIndex;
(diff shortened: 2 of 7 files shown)


packages/documents/src/instance/equation-validation.ts
```diff
@@ -62,24 +64,26 @@ export function validatePathEquations<S extends Shape>(
             if (violations <= MAX_COUNTEREXAMPLES_PER_EQUATION) {
                 issues.push({
                     message:
-                        `Equation \`${equationName}\` is violated by a row of table \`${tableLabel}\`: ` +
+                        `${equationText} is violated by a row of table \`${tableLabel}\`: ` +
```
epatters, line 67: @tslil-topos, this error message was not directly usable since it includes raw UUIDs for the left and right sides when the equation is between Mappings rather than Attributes. I didn't try to fix this because I understand the TS-level instance validation to be temporary pending the development of suitable Rust-level data structures.

packages/ui-components/src/table_editor/table_editor.tsx
```diff
@@ -692,10 +708,6 @@ export function TableEditor(props: TableEditorProps) {
                                                     }
                                                     aria-selected={isSelected(cell())}
                                                     aria-invalid={cellIsInvalid(cell())}
-                                                    title={
-                                                        cellIssueMessages(cell()).join("\n") ||
-                                                        undefined
-                                                    }
```
epatters: Maybe it was too aggressive to delete this right now. Anyway, I'd be happy to display the cell-level errors more nicely too in a follow-up PR.

packages/ui-components/src/table_editor/table_editor.tsx
```diff
@@ -797,6 +809,49 @@ export function TableEditor(props: TableEditorProps) {
...
+            {...triggerProps}
+            class={styles.rowHeader}
+            scope="row"
+            classList={{ [styles.invalid]: invalid() }}
+        />
+    );
+
+    return (
+        <Show when={invalid()} fallback={header()}>
+            <Tooltip hoverableContent={false} openOnFocus={false}>
```
epatters, line 831: @kasbah, note that this tooltip is only getting the right styling due to CSS leakage from other components. Likely worth adding a `Tooltip` component to `ui-components`.

packages/ui-components/src/table_editor/table_editor.stories.tsx
```diff
@@ -234,8 +234,8 @@ export const DoubleClickBooleanCell: Story = {
             name: "active",
         }) as HTMLTableCellElement;
         const activeColumn = activeHeader.cellIndex;
-        const firstRow = canvas.getAllByRole("row")[1]!;
-        const cell = within(firstRow).getAllByRole("gridcell")[activeColumn]!;
+        const firstRow = canvas.getAllByRole("row")[1] as HTMLTableRowElement;
+        const cell = firstRow.cells[activeColumn]!;
```
epatters, line 238: I don't want to complain about having tests for UI components, but these are rather fragile.

packages/ui-components/src/table_editor/table_editor.tsx
```diff
@@ -693,8 +709,9 @@ export function TableEditor(props: TableEditorProps) {
                                                     aria-selected={isSelected(cell())}
                                                     aria-invalid={cellIsInvalid(cell())}
                                                     title={
-                                                        cellIssueMessages(cell()).join("\n") ||
-                                                        undefined
+                                                        cellIssues(cell())
+                                                            .map(issueMessage)
+                                                            .join("\n") || undefined
```
epatters, line 714: It would be good to improve the styling of these too. I could do that in a follow-up PR.

packages/documents/src/instance/equation-validation.ts
tslil-topos: thanks for taking a look, apologies for the oversight. I agree broadly, it's not worth investing too much into fixing the *validation.ts files we have as long as the basic functionality is enough to share the vision it has probably served its purpose

packages/ui-components/src/table_editor/table_editor.stories.tsx
kasbah: That's fair, I'll prune these and look at making them less brittle.

packages/ui-components/src/table_editor/table_editor.tsx
kasbah: https://github.com/ToposInstitute/CatColab/issues/1482

kasbah (approved):
Just UI wise I'd prefer highlighting (background) the whole row in the same pink/red we are using for the other errors. I guess then the tooltip needs to appear on the whole row as well. I'd be ok with changing that on a follow-up PR though. Rest LGTM.

epatters:
Thanks for the review! I actually started by coloring the whole row red, but thought it looked too aggressive, as well as possibly being confusing with cell-level errors. Anyway, we can play around with it.

PR #1478, opened 2026-09-10
Fix rich-text issue with transactions

kasbah:
Continues with #1475 by patching the issue in automerge, since it is in `applyPatches` and affects transactions too. I reported upstream: https://github.com/automerge/automerge/issues/1548. The actual patch to automerge is LLM generated and I have not verified it beyond the included test.

I am also still opting for completely removing am-solid-primitives because there is another issue that would need to be patched there (already reported), so it's just easier to fix that on our own implementation for now. I'll follow up with a test and fix for that.


kasbah:
Somehow I can't change base to main so it might easiest to push this to main manually when it has been reviewed.

epatters:
> Somehow I can't change base to main so it might easiest to push this to main manually when it has been reviewed.

Just to note that I wasn't able to delete the branch associated with the merged #1475 until I rebased the next (and final) in the stack, #1476. More confusing and possibly broken aspects of the PR stack feature.

epatters (approved):
Thanks for reporting upstream. Happy to approve these temporary measures as is.

PR #1479, opened 2026-09-10
Fix `makeDocHandleReactive` output being affected when changing clone

kasbah:
This is a manifestation of a similar issue as am-solid-primitives' https://github.com/automerge/automerge-repo/issues/721 in our own `makeDocHandleReactive`.


epatters (approved):
It's unfortunate to make a full clone on every document change, but better to be correct than efficient while we await a proper upstream fix.

PR #1489, opened 2026-09-14
Fix test for orphaned columns story

kasbah:


tslil-topos (approved):
how was this passing?

kasbah:
This was caused by interleaved merges. 

- PR #1472 added the orphan-column test expecting 3 headers
- PR #1476 independently added the row-header <th> and was merged first.

PR #1497, opened 2026-09-14
Add a retry button to LLM conversation

kasbah:
image


packages/frontend/src/llm_conversation/conversation_controller.ts
```diff
@@ -114,23 +122,26 @@ export function createLLMConversationController(
...
-            const result = await runLLMConversationTurn(
-                conversation(),
-                binder.store,
-                key,
-                userInput,
-                handleTurnEvent,
-            );
+            const result = await run(key);
+            if (result.tag === "Retryable") {
+                const latestInteraction = conversation().interactions().at(-1);
```
tslil-topos: i don't think you should need to do this, `retryLastLLMConversation` already deals with this case 

```typescript
export async function retryLastLLMConversationResponse<...>(...): Promise<LLMConversationTurnResult> {
...
    if (conversation.interactions().at(-1)?.tag !== "user-message") {
        return { tag: "Failed", error: "The latest interaction is not a user message." };
    }
...
}
```

did you notice a deficiency?

tslil-topos (changes requested):
i don't know how to actually force this situation for testing, so it's difficult for me to gauge whether we're honouring the ephemeral display of `attempts` in the return value of `LLMConversationTurnResult`.

That aside, seems to me like we're duplicating a check?

PR #1505, opened 2026-09-16
add busy spinner to the llm conversation after the static "Running" text

tslil-topos:
We had no obvious indicator that the page was "still doing something"

image


packages/ui-components/src/llm_conversation/llm_conversation_editor.tsx
```diff
@@ -128,6 +130,9 @@ export function LLMConversationEditor(props: LLMConversationEditorProps) {
             </div>
             <div class={styles.composer}>
                 <div class={styles.status} aria-live="polite">
+                    <Show when={props.busy}>
+                        <Spinner class="small" aria-hidden="true" />
```
kasbah: My convention has been to use a prop e.g. `size="small"` and the set the style class inside the component.

kasbah (changes requested):

packages/ui-components/src/llm_conversation/llm_conversation_editor.tsx
tslil-topos: thanks, gave it another go

kasbah (approved):

PR #1510, opened 2026-09-17
Update Automerge to v3.5.0

kasbah:
The [release](https://github.com/automerge/automerge/releases/tag/js/automerge-3.5.0) fixes two of the issues we were patching for. 

> - Large list insertions no longer fail with "Maximum call stack size exceeded"
> when applying consolidated patches. The WASM bindings apply insertions in
> bounded chunks rather than passing every element to a single splice call.
> 
> - applyPatch and applyPatches support inserting and updating embedded
> rich-text blocks. When applied to plain JavaScript strings, blocks are
> represented by object replacement characters and updates to their contents
> are ignored. When applied inside an Automerge change callback, block data
> is preserved.


kasbah:
I've kicked the tires a little bit and not found any issues.

epatters (approved):
Glad to hear it, thanks for testing the new release out.

